PR #248, opened 2014-08-01
an upgrade fix and further pubIds porting

bozana:


bozana:
I will close this, because it is very old and overwritten with https://github.com/pkp/pkp-lib/issues/1457

PR #277, opened 2014-09-04
*8774* Grid redesign

beghelli:


beghelli:
It's already in master.

PR #369, opened 2015-01-05
pkp/pkp-lib#267 Fix cross-journal CrossRef URL generation

asmecher:
See pkp/pkp-lib#267.


jnugent:
Tested, and the Travis builds have now passed after restarting.  Merging.

PR #371, opened 2015-01-05
Check that the PHP zip extension is installed.

ubermichael:
Call the zipInstalled() method when checking if the zip extension is installed.

Fixes #266.


ubermichael:
PR sent against wrong branch.

PR #380, opened 2015-01-06
Allow Dataverse plugin to connect via API token

jwhitney:
pkp/pkp-lib#271 Allow API token to be submitted in place of username/password.


asmecher:
Not tested, but looks good. Merged, thanks.

PR #381, opened 2015-01-06
*8982* Make HTML galley pages validate by introducing a new function tha...

eirikhanssen:
...t grabs the body of the html document to be shown in the HTML galley, changes it to a div and returns that string.  ##eirikhanssen/ojs-dev-2_4-Bug8982##

### DIFF
--- a/classes/article/ArticleHTMLGalley.inc.php
+++ b/classes/article/ArticleHTMLGalley.inc.php
@@ -5,7 +5,6 @@
  *
  * Copyright (c) 2013-2014 Simon Fraser University Library
  * Copyright (c) 2003-2014 John Willinsky
- * Copyright (c) 2014 Eirik Hanssen
  * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
  *
  * @class ArticleHTMLGalley
@@ -34,7 +33,7 @@ class ArticleHTMLGalley extends ArticleGalley {
 	}
 
 	/**
-	* Eirik Hanssen, Oslo and Akershus University College of Applied Sciences
+	* Eirik Hanssen, Oslo and Akershus University College of Applied Sciences 2015
 	* Return string contains the <body> element of the HTML Galley file renamed to a <div> element.
 	* The returned string is a html fragment where empty tags will be self closed, as required by xhtml rulesybe.
 	* The purpose is to allow to allow OJS pages displaying the HTML Galley of an article to validate. Without filtering the HTML file string
@@ -47,8 +46,8 @@ class ArticleHTMLGalley extends ArticleGalley {
 	function getHTMLBodyContents($htmldoc){
 		// first check if DOMDocument class is available
 		if (class_exists('DOMDocument')) {
-			$originalHtmlDocument = new DOMDocument;
-			$htmlBodyContents = new DOMDocument;
+			$originalHtmlDocument = new DOMDocument();
+			$htmlBodyContents = new DOMDocument();
 			$originalHtmlDocument->loadHTML($htmldoc);
 
 			// Get the body element node from the full html document.
@@ -146,7 +145,7 @@ class ArticleHTMLGalley extends ArticleGalley {
 
 		// Eirik Hanssen 2014-05-15:
 		// to make HTML Galley webpages in OJS validate, $contents will be filtered through the new getHTMLBodyContents() function
-		// This wil grab the <body> element of the article (with all it's contents), rename it to a <div> tag  and return that instead as the HTML Galley html-string.
+		// This will grab the <body> element of the article (with all it's contents), rename it to a <div> tag  and return that instead as the HTML Galley html-string.
 		return $this->getHTMLBodyContents($contents);
 	}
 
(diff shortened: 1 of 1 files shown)


classes/article/ArticleHTMLGalley.inc.php
```diff
@@ -33,6 +34,51 @@ function isHTMLGalley() {
...
+	* This function requires the php DOMDocument module to operate. If DOMDocument module is not available, (often part of a php-xml package),
+	* a warning in form of a html-comment is appended to the html string instead of performing this filtering.
+	* @return string
+	*/
+	function getHTMLBodyContents($htmldoc){
+		// first check if DOMDocument class is available
+		if (class_exists('DOMDocument')) {
+			$originalHtmlDocument = new DOMDocument;
+			$htmlBodyContents = new DOMDocument;
+			$originalHtmlDocument->loadHTML($htmldoc);
```
eirikhanssen: I totally agree. I will introduce a check if the document can be parsed, and just pass the  $htmlDoc parameter unchanged if it can't. I'll also test with some rubbish html to see how it can handle that.

classes/article/ArticleHTMLGalley.inc.php
```diff
@@ -33,6 +34,51 @@ function isHTMLGalley() {
...
 	/**
+	* Eirik Hanssen, Oslo and Akershus University College of Applied Sciences
+	* Return string contains the <body> element of the HTML Galley file renamed to a <div> element.
+	* The returned string is a html fragment where empty tags will be self closed, as required by xhtml rulesybe.
+	* The purpose is to allow to allow OJS pages displaying the HTML Galley of an article to validate. Without filtering the HTML file string
+	* through the getHTMLBodyContents() function, a valid (full) HTML document will simply be inserted in a <div> element in the OJS page,
+	* and then that resulting webpage won't be valid HTML since it is illegal to have a full html-document inside a div tag.
+	* This function requires the php DOMDocument module to operate. If DOMDocument module is not available, (often part of a php-xml package),
+	* a warning in form of a html-comment is appended to the html string instead of performing this filtering.
+	* @return string
```
eirikhanssen: Would this be the right way to describe param?
@param $htmlDoc - The string representation of a html document. Intended to be used with the html-file uploaded to a html galley after getHTMLContents() has performed the filtering, but before getHTMLContents() returns.

classes/article/ArticleHTMLGalley.inc.php
```diff
@@ -98,7 +144,10 @@ function getHTMLContents() {
 			$contents = str_replace('{$' . $key . '}', $value, $contents);
 		}
 
-		return $contents;
+		// Eirik Hanssen 2014-05-15:
+		// to make HTML Galley webpages in OJS validate, $contents will be filtered through the new getHTMLBodyContents() function
+		// This wil grab the <body> element of the article (with all it's contents), rename it to a <div> tag  and return that instead as the HTML Galley html-string.
```
eirikhanssen: thanks for pointing out my typo

classes/article/ArticleHTMLGalley.inc.php
```diff
@@ -5,6 +5,7 @@
  *
  * Copyright (c) 2013-2014 Simon Fraser University Library
  * Copyright (c) 2003-2014 John Willinsky
+ * Copyright (c) 2014 Eirik Hanssen
```
eirikhanssen: Sure, I'll just remove it from the copyright section. My name is in the description before the function anyway.

classes/article/ArticleHTMLGalley.inc.php
```diff
@@ -33,6 +34,51 @@ function isHTMLGalley() {
...
+	* through the getHTMLBodyContents() function, a valid (full) HTML document will simply be inserted in a <div> element in the OJS page,
+	* and then that resulting webpage won't be valid HTML since it is illegal to have a full html-document inside a div tag.
+	* This function requires the php DOMDocument module to operate. If DOMDocument module is not available, (often part of a php-xml package),
+	* a warning in form of a html-comment is appended to the html string instead of performing this filtering.
+	* @return string
+	*/
+	function getHTMLBodyContents($htmldoc){
+		// first check if DOMDocument class is available
+		if (class_exists('DOMDocument')) {
+			$originalHtmlDocument = new DOMDocument;
```
eirikhanssen: it doesn't hurt to supply the brackets like in new DOMDocument(). Fixing.

classes/article/ArticleHTMLGalley.inc.php
eirikhanssen: After feeding it with several potentially problematic test html galleys, It seems that domDocument->loadHtml() is quite robust. I haven't been able to make it fail parsing, even if the structure is totally messed up. Of course, if the structure is all messed up, that will show in the output. However there might be some gotchas with character sets. I need to investigate a little more regarding to different charager sets. http://php.net/manual/en/domdocument.loadhtml.php

classes/article/ArticleHTMLGalley.inc.php
asmecher: You'd do...
   @param $htmlDoc string The string representation...
We use Doxygen to generate code documentation (e.g. http://pkp.sfu.ca/ojs/doxygen/stable/html/index.html). We're a lot better documented on the master branch, for what it's worth.

classes/article/ArticleHTMLGalley.inc.php
asmecher: That sounds positive. I assume you're testing end-to-end, so that suggests that saveXML is similarly robust -- that's a bit of a surprise, but also a positive one.

PR #382, opened 2015-01-09
*9022* include rights and license information in webfeed RSS

jnugent:


plugins/generic/webFeed/templates/rss.tpl
```diff
@@ -82,6 +83,15 @@
 				<dc:creator>{$author->getFullName()|strip|escape:"html"}</dc:creator>
 			{/foreach}
 
+			<dc:rights>
+				{translate|escape key="submission.copyrightStatement" copyrightYear=$article->getCopyrightYear() copyrightHolder=$article->getLocalizedCopyrightHolder()}
+			</dc:rights>
+			{if $article->getAccessStatus() == $smarty.const.ARTICLE_ACCESS_OPEN || ($article->getAccessStatus() == $smarty.const.ARTICLE_ACCESS_ISSUE_DEFAULT && $issue->getAccessStatus() == $smarty.const.ISSUE_ACCESS_OPEN) }
+				<cc:license rdf:resource="{$article->getLicenseURL()|escape}" />
```
asmecher: That's fine for CC licenses, but what about non-CC?

jnugent:
@asmecher this now does a simple test to see if it's a CC license or not. Also, the license URL is now included in the dc:rights element which seems to be an accepted practice:

http://dublincore.org/documents/usageguide/elements.shtml#rights

We don't store the radio button test on step 3 of the journal setup which may or may not have made this a bit more straight forward.

asmecher:
Ah, I missed that, thanks. Go ahead & merge.

PR #383, opened 2015-01-12
pkp/pkp-lib#275: Cleanup mismatched ol/ul in "How to ensure a blind peer review" help

ctgraham:
Can be picked to stable-2_4_5 as well.


asmecher:
Thanks, Clinton! I'll leave it out of ojs-stable-2_4_5 as it's unlikely to cause practical problems.

PR #385, opened 2015-01-12
*9016* PLN plugin now checks for null settings and applies defaults.

ubermichael:
*9016* PLN plugin fix update - Settings are not created for plugins during an upgrade installation. This work around supplies defaults for any setting that isn't found. 

(cherry picked from commit 5700ede93bf52baa79d554c14b2db966e7fad865)
(cherry picked from commit 9c59d866df64be8041e9b473f9cba73bb5b2a088)

Fixes bugzilla 9016.

### DIFF
--- a/plugins/generic/pln/PLNPlugin.inc.php
+++ b/plugins/generic/pln/PLNPlugin.inc.php
@@ -19,7 +19,7 @@ import('classes.issue.Issue');
 
 define('PLN_PLUGIN_NAME','plnplugin');
 
-define('PLN_PLUGIN_NETWORK', 'pkp-pln.lib.sfu.ca');
+define('PLN_PLUGIN_NETWORK', 'http://pkp-pln.lib.sfu.ca');
 
 define('PLN_PLUGIN_HTTP_STATUS_OK', 200);
 define('PLN_PLUGIN_HTTP_STATUS_CREATED', 201);
@@ -169,11 +169,10 @@ class PLNPlugin extends GenericPlugin {
 	
 	/**
 	 * @see PKPPlugin::getSetting()
-	 * @param $journalId string
+	 * @param $journalId int
 	 * @param $settingName string
 	 */
 	function getSetting($journalId,$settingName) {
-                        
 		// if there isn't a journal_uuid, make one
 		if ($settingName == 'journal_uuid') {
 			$uuid = parent::getSetting($journalId, $settingName);
--- a/plugins/generic/pln/PLNPlugin.inc.php
+++ b/plugins/generic/pln/PLNPlugin.inc.php
@@ -177,7 +177,7 @@ class PLNPlugin extends GenericPlugin {
 		// if there isn't a journal_uuid, make one
 		if ($settingName == 'journal_uuid') {
 			$uuid = parent::getSetting($journalId, $settingName);
-			if ($uuid) return $uuid;
+			if (!is_null($uuid) && $uuid != '') return $uuid;
 			$this->updateSetting($journalId, $settingName, $this->newUUID());
 		}
 		
--- a/plugins/generic/pln/PLNPlugin.inc.php
+++ b/plugins/generic/pln/PLNPlugin.inc.php
@@ -433,7 +433,7 @@ class PLNPlugin extends GenericPlugin {
 		
 		// retrieve the service document
 		$result = $this->_curlGet(
-			'http://' . PLN_PLUGIN_NETWORK . PLN_PLUGIN_SD_IRI,
+			PLN_PLUGIN_NETWORK . PLN_PLUGIN_SD_IRI,
 			array(
 				'On-Behalf-Of: '.$this->getSetting($journalId, 'journal_uuid'),
 				'Journal-URL: '.$journal->getUrl()
--- a/plugins/generic/pln/PLNPlugin.inc.php
+++ b/plugins/generic/pln/PLNPlugin.inc.php
@@ -19,9 +19,7 @@ import('classes.issue.Issue');
 
 define('PLN_PLUGIN_NAME','plnplugin');
 
-define('PLN_PLUGIN_NETWORKS', serialize(array(
-	'PKP' => 'pkp-pln.lib.sfu.ca'
-)));
+define('PLN_PLUGIN_NETWORK', 'pkp-pln.lib.sfu.ca');
 
 define('PLN_PLUGIN_HTTP_STATUS_OK', 200);
 define('PLN_PLUGIN_HTTP_STATUS_CREATED', 201);
@@ -448,13 +446,12 @@ class PLNPlugin extends GenericPlugin {
 	 */
 	function getServiceDocument($journalId) {
 			
-		$plnNetworks = unserialize(PLN_PLUGIN_NETWORKS);
 		$journalDao =& DAORegistry::getDAO('JournalDAO');
 		$journal =& $journalDao->getById($journalId);
 		
 		// retrieve the service document
 		$result = $this->_curlGet(
-			'http://' . $plnNetworks[$this->getSetting($journalId, 'pln_network')] . PLN_PLUGIN_SD_IRI,
+			'http://' . PLN_PLUGIN_NETWORK . PLN_PLUGIN_SD_IRI,
 			array(
 				'On-Behalf-Of: '.$this->getSetting($journalId, 'journal_uuid'),
 				'Journal-URL: '.$journal->getUrl()
--- a/plugins/generic/pln/PLNPlugin.inc.php
+++ b/plugins/generic/pln/PLNPlugin.inc.php
@@ -173,29 +173,12 @@ class PLNPlugin extends GenericPlugin {
 	 * @param $settingName string
 	 */
 	function getSetting($journalId,$settingName) {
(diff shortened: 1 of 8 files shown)


plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -175,12 +175,29 @@ function getStyleSheet() {
 	 * @param $settingName string
 	 */
 	function getSetting($journalId,$settingName) {
-
-		// if there isn't a journal_uuid, make one
-		if ($settingName == 'journal_uuid') {
-			$uuid = parent::getSetting($journalId, $settingName);
-			if ($uuid) return $uuid;
-			$this->updateSetting($journalId, $settingName, $this->newUUID());
+                $settingValue = parent::getSetting($journalId, $settingName);
```
asmecher: Watch out again for mixed indentation. Use tabs throughout.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -175,12 +175,29 @@ function getStyleSheet() {
...
+		if ($settingValue != null) return $settingValue;
+                        
+		// if a setting returns null, populate it with a default
+		switch ($settingName) {
```
asmecher: I think Chris already proposed a fix like this. I'd like to avoid it if at all possible -- having the settings updated when they're fetched is certain to have unintended consequences further down the road. If it's possible to reproduce the underlying problem, someone on the dev team can help to figure it out.

ubermichael:
I'm not sure if the bug is caused by going from 2.4.3 -> 2.4.5 (going from 2.4.4 didn't cause the problem) or if it's caused by having multiple journals in an OJS instance during the upgrade.

To recreate the upgrade bug:

`$ git checkout ojs-2_4_3-0`
create three journals
import data to each one.

Download http://pkp.sfu.ca/ojs/download/patch/ojs-2.4.3_to_2.4.5.patch.gz

`$ patch -p1 ~/Downloads/ojs-2.4.3_to_2.4.5.patch`

(The patch doesn't apply cleanly, but they are very minor)

`$ edit config.inc.php`
  installed = Off
`$ php tools/upgrade.php upgrade`

(There are some file permissions errors in the cache directory, but the upgrade completes)

`$ edit config.inc.php`
  installed = On

At this point, the plugin_settings table has no data for the plnplugin. As far as I can tell, it should.

Visit the plugin settings page for one of the journals and enable it. There will be no terms of service displayed. After saving the form, the plugin_settings table will only contain three rows for the journal:

```
'plnplugin','','2','enabled','1','bool'
'plnplugin','','2','journal_uuid','ECD14E56-B1C0-3B8D-950B-6F1B9446B4E5','string'
'plnplugin','','2','terms_of_use_agreement','s:2:\"N;\";','object'
```

The table is missing rows for checksum_type, object_type, object_threshold, pln_network, and pln_accepting. Those rows should be installed during the upgrade process.

plugins/generic/pln/PLNPlugin.inc.php
ubermichael: What are the unintended consequences that worry you?

asmecher: Logically, it doesn't make sense (to me) for a "get setting" operation to potentially introduce settings. But also there are a few modes of operation that have caused problems before WRT settings specifically. In some situations where getSetting(...) will be overridden to avoid a DB access (e.g. before the system is installed, when the database might not be present). This is the case e.g. for plugins that need to take effect in the installer. I don't think OTOH that there are intersections with this yet, but I'd rather not add another wrinkle if we can help it.

asmecher:
OK, looking at the code and refreshing my memory of how this works, I think the original diagnosis Chris made at <http://pkp.sfu.ca/bugzilla/show_bug.cgi?id=9016#c0> is correct: the only way that the default settings for the plugin get installed from xml/settings.xml is if a journal gets created while the plugin's code is in place.

None of the following activities will cause the default plugin settings to get installed:
- Upgrading an installation
- Installing the plugin via the Manager's "Plugins" interface

I would suggest using the plugin's settings form to introduce default settings, since I think administrators will already have to go there to configure things (am I correct?), and since having that form change journal settings will be expected behavior.

Does that make sense?

ubermichael:
That does make sense, but I don't think it will work. The plugin needs the default settings so it can fetch the terms of service in the depositor task, which runs before the the settings form is displayed.

asmecher:
Is this because the service URL is sort-of configurable? I'd suggest removing that -- there is only a single URL, it's already fixed in the code, and last I looked, there's no way of choosing or configuring a different one without modifying the code anyway.

If it becomes important to change the URL in the future, the administrator will already have to patch the code, so there's really nothing gained but extra complexity. And this problem.

ubermichael:
Ok. Then would you put all the settings into the form? Or would you load the defaults in the form's execute() function?

asmecher:
Michael, let me know when this is ready for review again.

ubermichael:
Hi Alec. I think it's ready.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -175,11 +173,11 @@ function getStyleSheet() {
 	 * @param $settingName string
 	 */
 	function getSetting($journalId,$settingName) {
-
+
```
asmecher: Still space-based indentation here; shouldn't have any whitespace characters on blank lines.

ubermichael:
Anything else to fix?

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -175,11 +173,11 @@ function getStyleSheet() {
 	 * @param $settingName string
```
asmecher: The type of the journalId parameter should be int.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -175,11 +173,11 @@ function getStyleSheet() {
...
 	 */
 	function getSetting($journalId,$settingName) {
-
+                        
 		// if there isn't a journal_uuid, make one
 		if ($settingName == 'journal_uuid') {
 			$uuid = parent::getSetting($journalId, $settingName);
-			if ($uuid) return $uuid;
+			if (!is_null($uuid) && $uuid != '') return $uuid;
```
asmecher: IMO the previous form of this was clearer.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -19,9 +19,7 @@
 
 define('PLN_PLUGIN_NAME','plnplugin');
 
-define('PLN_PLUGIN_NETWORKS', serialize(array(
-	'PKP' => 'pkp-pln.lib.sfu.ca'
-)));
+define('PLN_PLUGIN_NETWORK', 'pkp-pln.lib.sfu.ca');
```
asmecher: Why not include the http:// here? It looks (at a glance) like all uses of this constant prepend it anyway.

plugins/generic/pln/classes/form/PLNStatusForm.inc.php
```diff
@@ -50,9 +50,9 @@ function display() {
 		
 		if (!$networkStatusMessage) {
 			if ($networkStatus === true) {
-				$networkStatusMessage = __(PLN_PLUGIN_NOTIFICATION_PLN_ACCEPTING);
+				$networkStatusMessage = __("plugins.generic.pln.notifications.pln_accepting");
```
asmecher: Sorry, I'm getting trivial here -- but quoted strings that don't include variable replacements should use ' rather than " to delimit.

asmecher:
Done for now; just some small comments.

plugins/generic/pln/PLNPlugin.inc.php
ubermichael: That's how it was done in the past, and I've tried to keep the changes minimal.

asmecher: OK, but do you mind including the change? It's pretty quick.

asmecher: (I have no idea what your arrangement with Mark is, BTW, so if I'm going beyond that, let me know.)

ubermichael:
small comments addressed. please review.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -431,13 +428,12 @@ function termsAgreed($journalId) {
...
 	function getServiceDocument($journalId) {
 			
-		$plnNetworks = unserialize(PLN_PLUGIN_NETWORKS);
 		$journalDao =& DAORegistry::getDAO('JournalDAO');
 		$journal =& $journalDao->getById($journalId);
 		
 		// retrieve the service document
 		$result = $this->_curlGet(
-			'http://' . $plnNetworks[$this->getSetting($journalId, 'pln_network')] . PLN_PLUGIN_SD_IRI,
+			'http://' . PLN_PLUGIN_NETWORK . PLN_PLUGIN_SD_IRI,
```
asmecher: This is now double-appending the "http://", isn't it?

ubermichael:
It was.

PR #388, opened 2015-01-13
pkp/pkp-lib#282 Dataverse plugin: remove markup from dataset metadata

jwhitney:
Some markup accepted in dataset metadata in Dataverse 3, but not in 4.

### DIFF
--- a/plugins/generic/dataverse/DataversePlugin.inc.php
+++ b/plugins/generic/dataverse/DataversePlugin.inc.php
@@ -1225,13 +1225,17 @@ class DataversePlugin extends GenericPlugin {
 			if(!array_key_exists('holdingsURI', $pubIdAttributes)) {
 				$pubIdAttributes['holdingsURI'] = Request::url($journal->getPath(), 'article', 'view', array($article->getId()));
 			}
+			// Add copyright notice. 
+			if ($article->getCopyrightYear() && $article->getCopyrightHolder($article->getLocale())) {
+				AppLocale::requireComponents(LOCALE_COMPONENT_APPLICATION_COMMON);
+				$package->addMetadata('rights',	__('submission.copyrightStatement', array('copyrightYear' => $article->getCopyrightYear(), 'copyrightHolder' => $article->getCopyrightHolder($article->getLocale()))));
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


plugins/generic/dataverse/DataversePlugin.inc.php
```diff
@@ -1171,7 +1171,7 @@ function createMetadataPackage($article) {
 		$package->addMetadata('description', 
 						$article->getLocalizedData('studyDescription') ? 
 						$article->getLocalizedData('studyDescription') :
-						$article->getLocalizedAbstract()
+						String::html2text($article->getLocalizedAbstract())
```
asmecher: By the way, see http://pkp.sfu.ca/bugzilla/show_bug.cgi?id=9056 -- deposits generally shouldn't be using the getLocalized... functions because those are dependent on the current user's locale, which means a depositor working e.g. in English will deposit different content from someone working e.g. in Spanish.

Better to use, for journal-level content, e.g...
    $journal->getLocalizedSetting($journal->getPrimaryLocale())

For article-level content, e.g.:
    $article->getTitle($article->getLocale())

asmecher: (however, that's a general issue, so I won't hold back on merging this)

plugins/generic/dataverse/DataversePlugin.inc.php
```diff
@@ -1228,8 +1228,8 @@ function createMetadataPackage($article) {
 
 		// Journal metadata
 		$package->addMetadata('publisher', $journal->getSetting('publisherInstitution'));
-		$package->addMetadata('rights', $journal->getLocalizedSetting('copyrightNotice'));
-		$package->addMetadata('isReferencedBy', $this->getCitation($article), $pubIdAttributes);
+		$package->addMetadata('rights', String::html2text($journal->getLocalizedSetting('copyrightNotice')));
```
asmecher: Rights are now stored on a per-article basis; see http://pkp.sfu.ca/bugzilla/show_bug.cgi?id=8578

jwhitney:
Changes made for copyright notice & metadata deposit locale.

PR #389, opened 2015-01-13
pkp/pkp-lib#283 Dataverse plugin: remove Dataverse file from db when sup...

jwhitney:
...pfile deleted from OJS


asmecher:
Merged, thanks!

PR #391, opened 2015-01-15
pkp/pkp-lib#277 customBlockPlugIn localised

felixgruender:


plugins/generic/customBlockManager/CustomBlockPlugin.inc.php
```diff
@@ -176,13 +183,29 @@ function manage($verb, $args) {
...
 	function getContents(&$templateMgr) {
 		$journal =& Request::getJournal();
+		$journalId = $journal->getId();
+		
 		if (!$journal) return '';
+		
+		$locale = AppLocale::getLocale();
+		$primaryLocale = AppLocale::getPrimaryLocale();
+		$blockContent = $this->getSetting($journalId, 'blockContent');
+		$blockContentTxt = '';
```
asmecher: I don't think this variable is used anywhere.

plugins/generic/customBlockManager/CustomBlockEditForm.inc.php
```diff
@@ -48,7 +48,7 @@ function CustomBlockEditForm(&$plugin, $journalId) {
 	function initData() {
 		$journalId = $this->journalId;
 		$plugin =& $this->plugin;
-
+
```
asmecher: There are a couple of extra whitespace changes in the pull request; better remove those just to keep things clean.

plugins/generic/customBlockManager/CustomBlockPlugin.inc.php
```diff
@@ -176,13 +183,29 @@ function manage($verb, $args) {
...
+	
+		if (!is_array($blockContent)) {
+			$this->updateSetting($journalId, 'blockContent', array($primaryLocale => $blockContent));
+			$blockContentTxt = $blockContent;
+		}
```
asmecher: Generally we use...
    } else {
...rather than splitting onto two lines. Just a minor coding convention suggestion.

plugins/generic/customBlockManager/CustomBlockPlugin.inc.php
asmecher: Ah, actually, I see where it's used. Mind doing a little bit of cleanup around variable names? It's easy to misread at the moment.

asmecher: (I'd suggest adding an upgrade step so that content is standardized to the new localized form during the upgrade, so we don't have to consider both cases in the future.)

felixgruender:
Thanks for the hints regarding white spaces, coding conventions etc.! 
I added an upgrade step in classes/install/Upgrade.inc.php (function localizeCustomBlockSettings())

PR #394, opened 2015-01-19
pkp/pkp-lib#286 Dataverse plugin: publish Dataverses via API

jwhitney:
Publish draft Dataverses via API, if supported.
Added null check on study creation if request fails.


asmecher:
Looks good, Jen - merged!

PR #398, opened 2015-01-28
Update block.tpl

acuartango:
If you have OJS in spanish and english (for example), this is that happens at render the keyword cloud:

If i switch English i see ok the keyword cloud, because the component writes:
.....         font-size: 183.88888888889%;           ............

In Spanish the result is not css compatible, because the number in font-size has a comma:
.........      font-size: 183,88888888889%;       ...........

So i have rounded the real number to integer to have no comma nor point problem:
font-size: 183%;

Now you can see the keyword cloud without problem in english, spanish and other languages.


asmecher:
Merged, thanks! Will cherry-pick to ojs-dev-2_4, which is where this stuff would normally go for release in the next 2.4.x.

PR #410, opened 2015-01-31
Update file

josekarvalho:
Corrections on the email messages


mtub:
Hi Jose, thanks for your contributions. I'll have a look at the files. I would suggest that you try to put your future commits into one single pull request - it makes reviewing somewhat easier.

mtub:
Generally, new OJS 2.x translations should be made for the ojs-dev-2_4 branch as this is the branch  the new OJS 2.x release will come from.

PR #412, opened 2015-02-03
pkp/pkp-lib#309 Don't send duplicated emails to authors in /editor/notifyUsers

lepidus:
Fixes pkp/pkp-lib#309


asmecher:
@lepidus, thanks, that looks good! Merging.

PR #413, opened 2015-02-03
pkp/pkp-lib#297 Author is pre-selected in the register-as-section of the...

felixgruender:
... registration form if the referrer is information/authors or about/submissions, because in these cases the user who wants to register is very likely an author.

### DIFF
--- a/classes/user/form/RegistrationForm.inc.php
+++ b/classes/user/form/RegistrationForm.inc.php
@@ -157,6 +157,13 @@ class RegistrationForm extends Form {
 	 */
 	function initData() {
 		$this->setData('registerAsReader', 1);
+		
+		$referrer = (isset($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : null);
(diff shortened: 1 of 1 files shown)


asmecher:
@felixgruender, rather than looking at the HTTP_REFERER header, why not set a URL parameter on the link the user follows? That way one page doesn't have to know the URL of the other; the incoming URL just has to tell the form that it wants the author box to be pre-checked.

bozana:
Alec, how to do that? :-\

bozana:
Do you maybe mean to pass the parameter from those two pages (information/authors and about/submission) to the header template and further to the 'register' link, so the register handler and form know to pre-select the checkbox? :-)

asmecher:
Hmm, I forgot that would involve the header template. It is messy. But on the plus side it would work with path_info_disabled, where the current referrer check doesn't.

Can I suggest keeping this as a local modification until we hear of similar demand from other users?

bozana:
yes, sure :-)

asmecher:
Thanks!

PR #414, opened 2015-02-04
pkp/pkp-lib#313: Match refback size limit in form to size limit in database

ctgraham:
http://pkp.sfu.ca/support/forum/viewtopic.php?f=8&t=13690&p=52415#p52415


asmecher:
Merged, thanks!

PR #415, opened 2015-02-04
Update admin.xml

Jirrka:


Jirrka:
Update of Czech translation

PR #417, opened 2015-02-06
Attempt to commit Czech translation

Jirrka:


mtub:
Sorry, Jiri, this won't work like that because your locale repository is a copy of the OJS master branch. Please see my email.

Jirrka:
Aha,
OK, no problem, I have all files locally on my computer, so no any work is lost ☺

Seems I have still to correct heading of the files.

Jiri

PR #424, opened 2015-02-12
pkp/pkp-lib#335: fix checks in GeoLocationTool instanciation

ctgraham:
Add new isPresent() method and utilize it instead of check of (boolean) $object.

### DIFF
--- a/classes/statistics/StatisticsHelper.inc.php
+++ b/classes/statistics/StatisticsHelper.inc.php
@@ -297,7 +297,7 @@ class StatisticsHelper {
 
 	/**
 	* Get the geo location tool.
-	* @return GeoLocationTool
+	* @return mixed GeoLocationTool object or null
 	*/
 	function &getGeoLocationTool() {
 		$geoLocationTool = null;
--- a/plugins/generic/usageStats/GeoLocationTool.inc.php
+++ b/plugins/generic/usageStats/GeoLocationTool.inc.php
@@ -26,7 +26,8 @@ class GeoLocationTool {
 
 	/**
 	 * Constructor.
-	 * @param $argv array command-line arguments
+	 * If we cannot find the database file, an empty object will be constructed.
+	 * Use the method isPresent() to check if the database file is present before use.
 	 */
 	function GeoLocationTool() {
 		$geoLocationDbFile = dirname(__FILE__) . DIRECTORY_SEPARATOR . "GeoLiteCity.dat";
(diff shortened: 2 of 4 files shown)


beghelli:
Clinton, I think that if the tool can't be used, it shouldn't be initialized. So I would change the check for the db to the geo tool getter inside statistics helper class, and keep the old code checking for the tool or not. What do you think?

ctgraham:
I was leaning that direction until I saw the way the helper class was forcing the object.  I chose this isPresent() method for backwards compatibility, but if that isn't a concern, I can change the helper to return a null instead of the empty object.

beghelli:
The helper was already returning null right?

2015-02-16 19:30 GMT-02:00 Clinton Graham <notifications@github.com>:

ctgraham:
No, StatisticsHelper::getLocationTool() was returning an empty object:
https://github.com/pkp/ojs/blob/ojs-stable-2_4_5/classes/statistics/StatisticsHelper.inc.php#L306
calls
https://github.com/pkp/ojs/blob/ojs-stable-2_4_5/plugins/generic/usageStats/UsageStatsPlugin.inc.php#L225
which calls
https://github.com/pkp/ojs/blob/ojs-stable-2_4_5/plugins/generic/usageStats/GeoLocationTool.inc.php#L31
which instanciates as an object with
```
$this->_geoLocationTool == NULL
$this->_regionName == NULL
$this->_isDbFilePresent == FALSE
```
We could check for the empty object and return null at:
https://github.com/pkp/ojs/blob/ojs-stable-2_4_5/plugins/generic/usageStats/UsageStatsPlugin.inc.php#L232

But, how would we check?  Looking at the object's public properties is brittle, so I wrote the isPresent() method.  Once the method existed, why not use it everywhere?  That was my workflow.

What do you think?

beghelli:
I just don't think it makes sense to return an object that can't be used. I
was doing that, but now that you figured this out, I think we can use your
method to define if we return the object or not via helper. That would
facilitate the other places we use the object, we just need to check for
the object instead of remembering to check if the db is present. The
documentation in the getter method can warn that, it can return the object
or null if the object can't be used. That's enough information for
developers, they don't need to know that they have to check for the db file
presence.

2015-02-17 11:53 GMT-02:00 Clinton Graham <notifications@github.com>:

beghelli:
Already merged in. Thanks.

PR #428, opened 2015-02-18
Added li-ids for better css formatting in about subpages

ojsbsb:
Previously, it was not possible to select the li items seperately for formatting. Now, each has its own id that can be used.


asmecher:
Looks good; merged. Thanks for contributing!

PR #429, opened 2015-02-20
pkp/pkp-lib#245 Fix memory problem related with scheduled tasks log

beghelli:


plugins/generic/usageEvent/UsageEventPlugin.inc.php
```diff
@@ -286,6 +286,18 @@ function _buildUsageEvent($hookName, $args) {
 			$request, null, $canonicalUrlPage, $canonicalUrlOp, $canonicalUrlParams
 		);
 
+		// Make sure we log the server name and not aliases.
+		$configBaseUrl = Config::getVar('general', 'base_url');
+		$requestBaseUrl = $request->getBaseUrl();
+		if ($requestBaseUrl !== $configBaseUrl) {
+			// Make sure it's not an url override (no alias on that case).
+			if (!in_array($requestBaseUrl, Config::getContextBaseUrls()) && 
+					$requestBaseUrl !== Config::getVar('general', 'base_url[index]')) {
```
asmecher: Bruno, does this use of base_url[index] presume a relationship between the site-wide URL and a journal's URL (e.g. that the site-wide URL is a substring of the journal URL)? That may not be the case --- e.g. Coaction's setup.

beghelli: no, it's always searching for the entire base url string... I just want to know if the current request base url contains a server alias or not. So, if that's not equal to any base_url config (including the override ones) defined in config.inc.php, then I consider as an alias. Does that makes sense?

ctgraham: I'm struggling with this logic as well.  $_SERVER['HTTP_HOST'] will be reflected in $requestBaseUrl, right?  And if the requested HTTP_HOST doesn't match any of the known base_url[] entries, we set the canonical URL to config.inc.php's general base_url?  That doesn't sound right.

If HTTP_HOST doesn't match a known base_url[] entry, we need to look up the base_url[] entry via the current journal's shortname.

beghelli: we do. Config::getContextBaseUrls() returns all context base_url[], and we check $requestedBaseUrl against that array. We just move on (setting the general base url from config) if there's not match there. If there's match, we do nothing, because the $requestBaseUrl is already not an alias. And the canonical already uses it via router->url

ctgraham: Here's the scenario I'm considering:

config.inc.php
base_url = http://scholarlyexchange.org/ojs
base_url[index] = http://scholarlyexchange.org/ojs
base_url[myJournal] = http://myjournal.tld

Apache:
ServerName myjournal.tld
ServerAlias www.myjornal.tld
ServerAlias myjournal.myorg.tld

HTTP_HOST = http://www.myjournal.tld
or
HTTP_HOST = http://myjournal.myord.tld

beghelli: ok, so in both cases the config base_url will be different compared to the HTTP_HOST. Then the check will continues, is there anyother base_url setting that matches the HTTP_HOST? No, so the final logged base url will be the config base_url...

ctgraham: And will the config's general base_url work for stat's processing at that point, even if mod_rewrite obscured the journal's shortname in the URI?  If so, then we are fine.  If not, I think the code should look more like:
`$canonicalUrl = str_replace($requestBaseUrl, Config::getVar('general', 'base_url['.$journalKey.']') ? Config::getVar('general', 'base_url['.$journalKey.']') : $configBaseUrl, $canonicalUrl);`
(untested)

beghelli: if the mod_rewrite did that, users have to use base_url[context_path] right? otherwise the system links would be broken... if http_host will also present the end result of the mod_rewrite, then we are fine, because then the request base url will be found inside the config file, both here and when processing...

ctgraham: The Apache virtualhost uses mod_write to point a set of hosts to a particular journal.

For example

```
<VirtualHost *:80>
    ServerAdmin "webmaster@DOMAINNAME"
    ServerName DOMAINNAME
    #ServerAlias www.DOMAINNAME
    MIMEMagicFile /dev/null
    CustomLog logs/DOMAINNAME.access_log "%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-agent}i\""
    ErrorLog logs/DOMAINNAME.error_log

    # include our standard se exchange vhost file
    Include conf.d/se_vhost/se_standard_vhost/se_standard_vhost.conf

    RewriteEngine on
    RewriteRule ^/(index.php)?/?$ /ojs/index.php/JOURNALNAME/index [L]

    RewriteRule ^/JOURNALNAME/(.*)$ /$1 [R,L]

    RewriteCond %{REQUEST_URI} !/index.php/
    RewriteCond %{DOCUMENT_ROOT}ojs%{REQUEST_URI} !-d
    RewriteCond %{DOCUMENT_ROOT}ojs%{REQUEST_URI} !-f
    RewriteCond %{REQUEST_FILENAME} !-d
    RewriteCond %{REQUEST_FILENAME} !-f
    RewriteRule ^(.*)$ /ojs/index.php/JOURNALNAME/$1 [QSA,L]

    RewriteCond %{DOCUMENT_ROOT}ojs%{REQUEST_URI} -f
    RewriteRule ^(.*)$ /ojs/$1 [QSA,L]

</VirtualHost>
```

So, HTTP_HOST could end up as ANY of the ServerName or ServerAliases, and the REQUEST_URI will mask the /ojs/index.php/JOURNALNAME.

This is why I think `Config::getVar('general', 'base_url['.$journalKey.']')` is required.

beghelli: I think I finally got it Clinton... I will just wait for the call to finish and will change that. Thanks... :)

PR #432, opened 2015-02-23
pkp/pkp-lib#245 Use files to store scheduled task execution logs

beghelli:


pages/admin/AdminFunctionsHandler.inc.php
```diff
@@ -109,6 +109,31 @@ function clearDataCache() {
 
 		Request::redirect(null, 'admin');
 	}
+
+	/**
+	 * Download scheduled task execution log file.
+	 * @param $args array
```
asmecher: Function comments need fixing.

pages/admin/AdminFunctionsHandler.inc.php
```diff
@@ -109,6 +109,31 @@ function clearDataCache() {
...
+	 */
+	function downloadScheduledTaskLogFile() {
+		$this->validate();
+		$application =& Application::getApplication();
+		$request =& $application->getRequest();
+
+		$file = $request->getUserVar('file');
```
asmecher: This might need sanitizing. True, this is a site administrator function, but you could effectively download anything off the server this way, I think.

beghelli: only if a developer changes the code. The variable file here is only the file name, not the complete path. The scheduled task helper will only get this filename and search for it inside the scheduled task folder, in private ojs files folder. So basically one can download everything that's inside the scheduled tasks logs folder, which is the purpose.

beghelli: actually I dreamed that I did enforced that the file should be inside the scheduled tasks folder. One can pass something like '../../anything here' and access the other files. You're right, I will fix this.

PR #433, opened 2015-02-25
pkp/pkp-lib#305 DOAJ export enhanced

felixgruender:
Now it's possible to export not just a whole journal, but single issues or articles. Moreover, issues or articles can be marked as 'registered' if they have already been registered in DOAJ.


plugins/importexport/doaj/DOAJPlugin.inc.php
```diff
@@ -103,6 +144,314 @@ function exportJournal(&$journal, $outputFile = null) {
...
 
-?>
+	/**
+	 * @param $request PKPRequest
+	 * @param $selectedObjects array
+	 */
+	function _markRegistered($request, $selectedObjects) {
+
+		$target = $request->getUserVar('target');
+		$action = $target . 's';
```
asmecher: Generally speaking I'm not fond of programmatically building function names like this -- this case has aleady caused a mess in the CrossRef plugin when "all" was added as an option, then pluralized as "alls". I'd suggest using switch/case or whitelist or something clearer. It's also hard to grep for uses of functions when they're built this way.

plugins/importexport/doaj/DOAJPlugin.inc.php
```diff
@@ -103,6 +144,314 @@ function exportJournal(&$journal, $outputFile = null) {
...
+	 */
+	function _displayArticleList($templateMgr, $journal, $unregistered = false) {
+		$this->setBreadcrumbs(array(), true);
+
+		// Retrieve all published articles.
+		$articles = $this->_getAllPublishedArticles($journal);
+
+		// Retrieve array elements without index "doaj::registered"
+		if ($unregistered == true) {
+			$articles = array_filter($articles, function($article) { return !$article['article']->getData('doaj::registered'); });
```
asmecher: Check this syntax against our PHP version requirements -- consider using create_function instead.

plugins/importexport/doaj/DOAJPlugin.inc.php
```diff
@@ -103,6 +144,314 @@ function exportJournal(&$journal, $outputFile = null) {
...
+			// Prepare and display the article template.
+			$templateMgr->assign_by_ref('articles', $iterator);
+			$templateMgr->display($this->getTemplatePath() . 'articles.tpl');
+		}
+	}
+
+	/**
+	 * Retrieve all published articles.
+	 * @param $journal Journal
+	 * @return DAOResultFactory
```
asmecher: Return type is wrong.

plugins/importexport/doaj/classes/DOAJExportDom.inc.php
```diff
@@ -20,22 +20,35 @@ class DOAJExportDom {
 	 * Generate the export DOM tree for a given journal.
 	 * @param $doc object DOM object
 	 * @param $journal object Journal to export
+	 * @param $selectedObjects array
 	 */
-	function &generateJournalDom(&$doc, &$journal) {
-		$issueDao =& DAORegistry::getDAO('IssueDAO');
-		$sectionDao =& DAORegistry::getDAO('SectionDAO');
-		$pubArticleDao =& DAORegistry::getDAO('PublishedArticleDAO');
+	function generateJournalDom($doc, $journal, $selectedObjects) {
```
asmecher: I don't mind if we start changing these now -- but the reason the references are used so heavily is that PHP4 treats objects very differently than PHP5, and the references bridged that gap. I think our user community has (finally ... finally) moved beyond PHP4, so I'm content to let our compatibility slip.

plugins/importexport/doaj/classes/PubObjectCache.inc.php
```diff
@@ -0,0 +1,137 @@
+<?php
+
+/**
+ * @file plugins/importexport/doaj/classes/PubObjectCache.inc.php
```
asmecher: Is this a duplicate of the PubObjectCache class used elsewhere? If it's identical, better ot use a symlink rather than duplicating the class entirely.

plugins/importexport/doaj/templates/index.tpl
```diff
@@ -0,0 +1,27 @@
+{**
+ * plugins/importexport/doaj/index.tpl
+ *
+ * Copyright (c) 2013-2014 Simon Fraser University Library
```
asmecher: All copyrights need to be updated to 2015.

asmecher:
Thanks, Felix -- I've made some comments. (My apologies if they're a little terse.) Have a look and let me know when you've reviewed them; thanks as always for contributing!

felixgruender:
Alec, thanks for your comments! I've just made these corrections:

1) $action = $target . 's' replaced by switch/case
2) closure in callback function for array_filter(...) replaced by create_function(...)
3) return type for function _getAllPublishedArticles corrected
4) copyright years in index.tpl corrected
5) PubObjectCache class has been left as it is because it's not exactly the same as in the other plugins (some changes in add function)

asmecher:
Thanks, Felix, I'll take another review. But I worry about a modified PubObjectCache class hanging around -- this might cause the plugins that provide it to interact in bad ways. (See https://github.com/pkp/ojs/pull/433/files#diff-40cd2fc8a776db8ed4280d9787e84361R35 -- depending on which plugin imports the class first, you might get your modified copy or the version from the other plugins.) If it's possible, I'd suggest symlinking in the unmodified version and then subclassing it with your modifications, if that's not too convoluted to read?

plugins/importexport/doaj/DOAJPlugin.inc.php
```diff
@@ -103,6 +144,326 @@ function exportJournal(&$journal, $outputFile = null) {
...
+		}
+
+		// show message & redirect
+		$this->_sendNotification(
+			$request,
+			'plugins.importexport.doaj.markRegistered.success',
+			NOTIFICATION_TYPE_SUCCESS
+		);
+		
+		switch($request->getUserVar('target')) {
```
asmecher: Should include a default case with assert(false). Also, initialize $action before the switch() or else PHP Scrutinizer will complain about a potentially uninitiailized variable being used.

plugins/importexport/doaj/DOAJPlugin.inc.php
```diff
@@ -103,6 +144,326 @@ function exportJournal(&$journal, $outputFile = null) {
...
+
+		// Retrieve all published issues.
+		AppLocale::requireComponents(array(LOCALE_COMPONENT_OJS_EDITOR));
+		$issueDao = DAORegistry::getDAO('IssueDAO'); /* @var $issueDao IssueDAO */
+		$issueIterator = $issueDao->getPublishedIssues($journal->getId(), Handler::getRangeInfo('issues'));
+
+		// check whether all articles of an issue are doaj::registered or not
+		$issues = array();
+		while ($issue = $issueIterator->next()) {
+			$issueId = $issue->getData('id');
```
asmecher: Use $issue->getId() instead of getData('id').

plugins/importexport/doaj/DOAJPlugin.inc.php
```diff
@@ -103,6 +144,326 @@ function exportJournal(&$journal, $outputFile = null) {
...
+	}
+
+	/**
+	 * Process a request.
+	 * @param $request PKPRequest
+	 * @param $journal Journal
+	 */
+	function _process($request, $journal) {
+		$objectTypes = $this->_getAllObjectTypes();
+		$target = $request->getUserVar('target');
```
asmecher: Mind trying to move this into the switch() as well? As you can tell, I'm not a big fan of assembling e.g. variable names using string operations.

plugins/importexport/doaj/classes/DOAJExportDom.inc.php
```diff
@@ -20,22 +20,35 @@ class DOAJExportDom {
...
-			$issue =& $issueDao->getIssueById($pubArticle->getIssueId());
+		$pubArticles = $pubArticleDao->getPublishedArticlesByJournalId($journal->getId());
+		while ($pubArticle = $pubArticles->next()) {
+
+			// check for selected issues:
+			$issueId = $pubArticle->getIssueId();
+			$selectedIssues = $selectedObjects[DOAJ_EXPORT_ISSUES];
+			if (!empty($selectedIssues) && !in_array($issueId, $selectedIssues)) continue;
+
+			$issue = $issueDao->getIssueById($issueId);
```
asmecher: I don't think there's any validation of the supplied IDs to check e.g. if they're in the current journal. This could be used to generate exports by a manager of one journal for content from another. Most DAO getter functions permit the specification of a journal ID as an optional parameter, which is often used in cases like these.

plugins/importexport/doaj/classes/DOAJExportDom.inc.php
```diff
@@ -20,22 +20,35 @@ class DOAJExportDom {
...
 
-		$pubArticles =& $pubArticleDao->getPublishedArticlesByJournalId($journal->getId());
-		while ($pubArticle =& $pubArticles->next()) {
-			$issue =& $issueDao->getIssueById($pubArticle->getIssueId());
+		$pubArticles = $pubArticleDao->getPublishedArticlesByJournalId($journal->getId());
+		while ($pubArticle = $pubArticles->next()) {
+
+			// check for selected issues:
+			$issueId = $pubArticle->getIssueId();
+			$selectedIssues = $selectedObjects[DOAJ_EXPORT_ISSUES];
```
asmecher: This will cause an uninitialized index warning, I think, as that index won't be set in all cases. (Also elsewhere.)

felixgruender:
Alec, I tried to implement your suggestions.
As for the PubObjectCache file: the changes I've made there were in fact just some minor code reductions (some parts of the add function weren't necessary for me).
So it's actually no problem to use the original PubObjectCache class used in the other plugins; therefore I created a symbolic link (pointing to plugins/importexport/datacite/classes/PubObjectCache.inc.php).

plugins/importexport/doaj/DOAJPlugin.inc.php
```diff
@@ -103,6 +144,332 @@ function exportJournal(&$journal, $outputFile = null) {
 		}
 		return true;
 	}
-}
 
-?>
+	/**
+	 * @param $request PKPRequest
```
asmecher: Missing function description

plugins/importexport/doaj/DOAJPlugin.inc.php
```diff
@@ -103,6 +144,332 @@ function exportJournal(&$journal, $outputFile = null) {
...
+	 * @param $unregistered boolean
+	 */
+	function _displayArticleList($templateMgr, $journal, $unregistered = false) {
+		$this->setBreadcrumbs(array(), true);
+
+		// Retrieve all published articles.
+		$articles = $this->_getAllPublishedArticles($journal);
+
+		// Retrieve array elements without index "doaj::registered"
+		if ($unregistered == true) {
```
asmecher: Better to use if/else to set $articles.

plugins/importexport/doaj/classes/PubObjectCache.inc.php
```diff
@@ -0,0 +1,147 @@
+<?php
```
asmecher: Are you sure this got added as a symbolic link? OTOH this means a copy was added.

asmecher:
Thanks again, Felix, just a couple of small comments this time. We're already into the code freeze process for OJS 2.4.6; in your opinion is this tested and stable enough to sneak into that release, or would you rather we waited until e.g. OJS 2.4.7?

bozana:
Hi Alec, when Felix code is merged I would add the symlink for PubObjectCache.inc.php and make another pull request, if OK so...?
Thanks!!!

asmecher:
Bozana, would you mind making a pull request based off this branch with the symlink? When that's ready we can close this PR and merge the whole thing at once.

I'm content to merge this for OJS 2.4.6, if you're both content that it's stable enough. (I've read the code but not run any tests.)

bozana:
Yes, but I will have to cherry-pick Felix commit, add my commit and make a new pull request, right?

asmecher:
You could check out his branch locally, change the file to a symlink, push that to your github, then make a PR from that.

asmecher:
Superseded by https://github.com/pkp/ojs/pull/439

PR #436, opened 2015-02-27
pkp/pkp-lib#366 Fix first journal event 

beghelli:


classes/journal/JournalStatisticsDAO.inc.php
```diff
@@ -28,16 +28,12 @@ class JournalStatisticsDAO extends DAO {
...
 		$result =& $this->retrieve(
-			'SELECT	LEAST(a.date_submitted, COALESCE(pa.date_published, NOW()), COALESCE(i.date_published, NOW())) AS first_date
-			FROM	articles a
-				LEFT JOIN published_articles pa ON (a.article_id = pa.article_id)
-				LEFT JOIN issues i ON (pa.issue_id = i.issue_id)
-				LEFT JOIN articles a2 ON (a2.article_id < a.article_id AND a2.date_submitted IS NOT NULL AND a2.journal_id = ?)
-			WHERE	a2.article_id IS NULL AND
-				a.date_submitted IS NOT NULL AND
-				a.journal_id = ?',
+			'SELECT	u.date_registered AS first_date
```
asmecher: What about journals that import old content that predates the OJS incarnation of the journal? Or journals that are migrated using e.g. the import/export tools? OTOH that won't bring across the registration date. I'd rather we based the date on scholarly metadata, and the user registration date isn't intended for that.

beghelli: You're right. But using the current solution or this one doesn't fix the bug. The statistics listed on stats page are not only about published objects, they're also about workflow, and that's why Scielo use it. Maybe we can use the first article submission date?

asmecher: Yes, I think the earliest of all those would make sense.

PR #441, opened 2015-03-04
*8615* Deprecate submissions.user_id in favour of stage assignments

asmecher:


classes/user/UserAction.inc.php
```diff
@@ -155,6 +148,13 @@ function mergeUsers($oldUserId, $newUserId) {
 		}
 		$userGroupDao->deleteAssignmentsByUserId($oldUserId);
 
+		// Delete stage assignments outright.
+		$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO');
+		$stageAssignments = $stageAssignmentDao->getByUserId($oldUserId);
+		while ($stageAssignment = $stageAssignments->next()) {
+			$stageAssignmentDao->deleteObject($stageAssignment);
+		}
+
```
beghelli: Just to make sure, we also moved the stage assignments previously right?

dbscripts/xml/ojs_schema.xml
```diff
@@ -507,9 +504,6 @@
 			<DEFAULT VALUE="1"/>
 		</field>
 		<descr>Submissions</descr>
-		<index name="submissions_user_id">
-			<col>user_id</col>
-		</index>
```
beghelli: Don't we need to know who uploaded the article in the first place? This can be still retrieved using the article log?

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
```diff
@@ -234,10 +234,17 @@ function execute($request) {
 
 			$user = $request->getUser();
 
+			// Get a list of author user IDs
+			$authorUserIds = array();
+			$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO');
+			$submitterAssignments = $stageAssignmentDao->getBySubmissionAndRoleId($submission->getId(), ROLE_ID_AUTHOR);
+			$submitterAssignment = $submitterAssignments->next();
+			assert($submitterAssignment); // At least one author should be assigned
```
beghelli: If an editor unassigns all authors from the stage, this assert might fail... should we avoid that happening?

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
```diff
@@ -234,10 +234,17 @@ function execute($request) {
 
 			$user = $request->getUser();
 
+			// Get a list of author user IDs
+			$authorUserIds = array();
+			$stageAssignmentDao = DAORegistry::getDAO('StageAssignmentDAO');
+			$submitterAssignments = $stageAssignmentDao->getBySubmissionAndRoleId($submission->getId(), ROLE_ID_AUTHOR);
+			$submitterAssignment = $submitterAssignments->next();
```
beghelli: This might create a payment object related to a translator that has nothing to do with the article other than translating it... Why not get the principal contact of the article? I think we have that for OJS also right?

classes/user/UserAction.inc.php
asmecher: Previously we updated the article.user_id to the new user. I thought about this and decided that for the new structure the best thing to do would be to delete the stage assignment, but it's somewhat ambiguous whether that's best.

The most common use of the "merge users" tool is to delete unwanted accounts. In that case they Editor will merge the account into his/her own account. If we transfer author stage assignments, then the Editor will end up with a whole bunch of unwanted assignments to subimissions they already have access to.

A case in which transferring stage assignments might be useful is de-duplicating author accounts, but I don't think that's a very common use.

For the most part, Merge Users is used to delete spam accounts, which typically won't have any content, so it's irrelevant.

dbscripts/xml/ojs_schema.xml
asmecher: Yes, the log has it. Otherwise I don't think it's important to know.

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
asmecher: The payment stuff is in a terrible state right now -- mostly hidden from the interface and not well maintained. I've got that scheduled for attention for OJS 3.1 along with subscriptions.

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
asmecher: This is a temporary fix that'll need attention when we deal with payments. Personally I suspect we'll have to tear out all the current payment stuff and start from scratch to get it done cleanly. We do have principal contact, but only for author records, not user records, so we can't use that. I'd rather that we had correspondence go to all assigned authors, and if we find we need a principal contact, then we'll have to add a flag to the stage assignment for that.

PR #442, opened 2015-03-05
pkp/pkp-lib#377 Improve theme capabilities

asmecher:


templates/article/article.tpl
```diff
@@ -6,15 +6,25 @@
...
-
-{include file="article/header.tpl"}
+{strip}
+	{if $galley}
+		{assign var=pubObject value=$galley}
+	{else}
+		{assign var=pubObject value=$article}
+	{/if}
+	{include file="article/header.tpl"}
+{/strip}
```
jnugent: Noticed that in the OMP pull request you're removing the {strip} tags, but keeping them here, e.g. https://github.com/pkp/omp/pull/105/files#diff-ec5fc29cce8cc46581c96498521337b2R5
Is there a reason we don't need them for common/header.tpl?

asmecher: This is just a minor bit of cleanup. We use {strip} here and there to prevent whitespace from getting in ahead of the header template, but I've noticed a few spots where that's not necessary and I've been removing it in cases like that. (Whitespace ahead of the <!DOCTYPE> sometimes causes problems IIRC.)

PR #451, opened 2015-03-20
pkp/pkp-lib#410 Clean up submission file classes

asmecher:


classes/article/SubmissionFileDAO.inc.php
```diff
@@ -10,7 +10,7 @@
  * @class SubmissionFileDAO
  * @ingroup monograph
```
jnugent: @ingroup submission instead

jnugent: or article.

asmecher: Fixed (also others around this line)

PR #452, opened 2015-03-23
pkp/pkp-lib#405 Permit full indexing metadata for supplementary files to be entered

asmecher:


classes/article/SubmissionFileDAO.inc.php
```diff
@@ -64,6 +64,8 @@ function getGenreCategoryMapping() {
 	function fromRow($row) {
 		if (isset($row['artwork_file_id']) && is_numeric($row['artwork_file_id'])) {
 			return parent::fromRow($row, 'SubmissionArtworkFile');
+		} elseif (isset($row['supplementary_file_id']) && is_numeric($row['supplementary_file_id'])) {
```
beghelli: I thought we got rid of this concept of supplementary files in OJS/OMP 3.0, in favor of a more flexible solution using genres... ArtworkFile makes sense to me, but supplementary file could be anything that's not the submission file itself... that's right?

asmecher: Supplementary file in the OJS 3.0 sense would mean something that's got a full Dublin Core metadata set, i.e. is indexed separately from the article's metadata.

beghelli: Did you see that we have a metadata plugin that implements DC metadata for DataObjects?

asmecher: Ah, no -- let me consider this a little further.

asmecher: Bruno, thinking about this a little more: OJS doesn't provide supplementary files as their own OAI records; each OAI record is a single article. The OAI data refers to supplementary files as "relation" elements but that's all. I think that's the appropriate thing to do for articles, until we have a requirement for more. However, the DC metadata will be active in other areas for supplementary content, such as Dataverse.

beghelli: Hum, I see Dc11SchemaArticleAdapter now, and the how it handle galleys. I
overlooked that the plugin implemented the DcSchema in an article level.

Although the structure is ready to implement the schema in another object
level (see the assoc type parameter), not sure if we can use the general
PKPDc11Schema for galleys, subject doesn't make much sense to me for files.
I just wanted to make sure we avoid duplicating DC schema in the system
code (inside object classes, for example) instead of using what's ready.

2015-04-07 14:11 GMT-03:00 Alec Smecher <notifications@github.com>:

asmecher: Yes, a good way to tackle that might be to add an interface or something for a DC-exposing data object. But no requirement yet.

PR #463, opened 2015-04-02
problem with subscription type form (format is always invalid)

trik:
ValidatorInSet uses strict in_array, value from the form (string) is always in valid (array contains only integers)


asmecher:
Thanks, @trik, but this has already been corrected by relaxing the "strict" mode check throughout:
https://github.com/pkp/pkp-lib/issues/437

ctgraham:
I would like to note that this (converting the validation to strings, or the parameters to integers) is probably the "better" solution than relaxing the "strict" check.

When we do add InSet validation to master (it's not there yet per https://github.com/pkp/pkp-lib/issues/437#issuecomment-87754495), I hope we follow this example.

PR #466, opened 2015-04-07
Italian translation updates

marchitelli:


mtub:
update to admin.xml is also included in #479

PR #469, opened 2015-04-15
5598 multilingual authors april25

vodingithub:


asmecher:
I think this pull request was sent in against the wrong branch -- it should be ojs-dev-2_4, which is currently almost the same as OJS 2.4.6 and will later be released as OJS 2.4.7.

PR #470, opened 2015-04-15
pkp/pkp-lib#460 Fix getting issue galley by best id using cache 

beghelli:


classes/issue/IssueDAO.inc.php
```diff
@@ -23,7 +23,7 @@ function _cacheMiss(&$cache, $id) {
 		if ($cache->getCacheId() === 'current') {
 			$issue =& $this->getCurrentIssue($id, false);
 		} else {
-			$issue =& $this->getIssueById($id, null, false);
+			$issue =& $this->getIssueByBestIssueId($id, null, false);
```
asmecher: I think this moots the following check:
https://github.com/pkp/ojs/blob/ojs-stable-2_4_6/pages/issue/IssueHandler.inc.php#L286

asmecher:
I looked through existing calls to getIssueById that involve caching enabled, as this change will affect them. I suspect there will be problems when numeric custom IDs are supplied but that's already the case with articles, so I think it's OK to go ahead.

PR #472, opened 2015-04-16
Simple Run with Docker Compose

arturluizbr:
I've put a docker-compose.yml to quickly start webservice container running OJS on port 80.

**mysql.conf** and **php.ini** can be use to personalize service's mysql and php.

To install Docker and Docker Compose:
[Docker Installation](https://docs.docker.com/installation/)
[Docker Compose Installation](https://docs.docker.com/compose/install/)


.gitignore
```diff
@@ -1,4 +1,5 @@
 config.inc.php
+cache
```
asmecher: There's already a .gitignore in cache/ which should take care of this -- unless there are some file types missing from that? IIRC having .gitignore in cache/ will cause cache to be created empty on checkout, and this will cause it not to be created at all.

.gitmodules
```diff
@@ -1,6 +1,7 @@
 [submodule "lib/pkp"]
 	path = lib/pkp
 	url = git@github.com:pkp/pkp-lib.git
+	branch = ojs-alpha_3_0
```
asmecher: I guess this means you're working with the 3.0a branch? That's a very old branch -- the master branch is currently both more advanced and more stable. If you'd like to experiment I'd recommend checking it out :)

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+web:
+    image: arturluizbr/apache-php:php5.6-development
```
asmecher: Could some of this be generalized? image, database connection info...

asmecher:
Very interesting, @arturluizbr -- I think there are a few mistaken adds in the pull request, but I'll pass this around to one or two on our team who are working with Docker. Thanks!

arturluizbr:
I tried to run master branch on PHP 5.6, but more than 3000 errors weren't letting me install it.

So i tried 3.0 because new layout.

The 3.0 branch won't be update anymore?

asmecher:
@arturluizbr, did you install the composer dependencies when using the master branch?

    cd lib/pkp
    composer.phar update
    cd ../..

Otherwise, let me know what some of the important errors look like and I'll see if I can help.

The master branch will eventually be released as OJS 3.0 (we're planning a beta for August). The alpha also came from the master branch, but it's now quite old.

asmecher:
Thanks, @arturluizbr, but it looks like this was sent against the wrong branch -- have a look at the commit list.

PR #474, opened 2015-04-16
pkp/pkp-lib#469 Move pnotify to composer dependency

asmecher:


asmecher:
Updated submodule outside this PR.

PR #477, opened 2015-04-17
pkp/pkp-lib#427 Bulk editing for email templates

felixgruender:
Email templates can be exported as XML file and edited all at once.

### DIFF
--- a/pages/manager/EmailHandler.inc.php
+++ b/pages/manager/EmailHandler.inc.php
@@ -210,7 +210,189 @@ class EmailHandler extends ManagerHandler {
 
 		Request::redirect(null, null, 'emails');
 	}
+	
+	/**
+	 * Export the selected email templates as XML
+	 * @param $args array
+	 * @@param $request PKPRequest
+	 */
+	function exportEmails($args, $request) {
+		$this->validate();
+		import('lib.pkp.classes.xml.XMLCustomWriter');
+		
+		$selectedEmailKeys = (array) $request->getUserVar('tplId');
+		if (empty($selectedEmailKeys)) {
+			$request->redirect(null, null, 'emails');
+		}
+		
+		$journal = Request::getJournal();
+		$doc = XMLCustomWriter::createDocument();
+		$emailTexts = XMLCustomWriter::createElement($doc, 'email_texts');
+		$emailTexts->setAttribute('locale', AppLocale::getLocale());
+		$emailTexts->setAttribute('xmlns:xsi', 'http://www.w3.org/2001/XMLSchema-instance');
+		
+		$emailTemplateDao = DAORegistry::getDAO('EmailTemplateDAO');
+		$emailTemplates = $emailTemplateDao->getEmailTemplates(AppLocale::getLocale(), $journal->getId());
+		
+		foreach($emailTemplates as $emailTemplate) {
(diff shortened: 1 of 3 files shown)


pages/manager/EmailHandler.inc.php
```diff
@@ -211,6 +211,197 @@ function enableEmail($args) {
...
+		if (!$fileName) {
+			$request->redirect(null, null, 'emails');
+		}
+		
+		$filesDir = Config::getVar('files', 'files_dir');
+		$filePath = $filesDir . '/journals/' . $journalId . '/' . $fileName;
+		
+		if (!$fileManager->uploadError($uploadName)) {
+			if ($fileManager->uploadedFileExists($uploadName)) {
+				if ($fileManager->uploadFile($uploadName, $filePath)) {
```
asmecher: Why not work with the file where it lives in the temporary upload area?

pages/manager/EmailHandler.inc.php
```diff
@@ -211,6 +211,197 @@ function enableEmail($args) {
...
+	}
+	
+	/**
+	 * Check if an XML file contains valid XML
+	 * @param $filePath string
+	 * @return boolean
+	 */
+	function _isValidXML($filePath) {
+		if (is_file($filePath)) {
+			return is_object(simplexml_load_file($filePath));
```
asmecher: This results in double-parsing the file, and adds a simplexml dependency we don't have elsewhere.

felixgruender:
I made the following 2 changes:
1) The XML file is checked now in the temporary upload directory (using FileManager::getUploadedFilePath()) - so the actual file upload isn't necessary any more
2) The _isValidXML() function has been removed - instead the result of XMLParser::parseStruct() is checked in the _saveEmailTemplates() function

pages/manager/EmailHandler.inc.php
```diff
@@ -210,7 +210,188 @@ function enableEmail($args) {
...
 		Request::redirect(null, null, 'emails');
 	}
-
+	
+	/**
+	 * Export the selected email templates as XML
+	 * @param $args array
+	 * @@param $request PKPRequest
+	 */
+	function exportEmails($args, $request) {
```
asmecher: Be careful not to omit the calls to $this->validate() -- that's what checks that the user is a Journal Manager. Without it, anyone can access the new functions.

pages/manager/EmailHandler.inc.php
```diff
@@ -210,7 +210,188 @@ function enableEmail($args) {
...
+		$filePath = $filesDir . '/journals/' . $journalId . '/' . $fileName;
+		
+		if (!$fileManager->uploadError($uploadName)) {
+			if ($fileManager->uploadedFileExists($uploadName)) {
+				$uploadedFilePath = $fileManager->getUploadedFilePath($uploadName);
+				if ($this->_saveEmailTemplates($uploadedFilePath, $journal)) {
+					if ($fileManager->deleteFile($uploadedFilePath)) {
+						$this->_showMessage($request);
+						$request->redirect(null, null, 'emails');
+						return true;
```
asmecher: I don't think the return values are ever used...

felixgruender:
1) $this->validate() has been added to the new functions
2) superfluous return values have been removed in function uploadEmails()

PR #488, opened 2015-04-29
Portuguese (brazilian) translation fix

gustavotonini:


beghelli:
Obrigado Gustavo. Sorry Marco, not sure if I was too quick, I just read something correct in Portuguese and I merged. Let me know if you need something else...

PR #491, opened 2015-05-05
Merging Arabic Pack Files

vormia:
These are some of the files, not all. I faced some difficulties with the "lib" directory, it's structure may have been changed thoroughly. Also, I miss the "help" directory from OJS project!


mtub:
Thank you so much, @vormia, this really looks great. I understand you've been working with the files from the ojs-dev-2_4 branch, which is just the right thing. But you've commited your pull request against the OJS master branch. So if you could make your pull request against the correct branch 'ojs-dev-2_4' this should be okay.

And also, I wouldn't recommend translating missing help files. Those files won't be around much longer, they won't be transferred to OJS 3.

vormia:
Many thanks for replying @mtub.
@asmecher gave me a clue to fix the orientation problem by using the CSS link you mentioned. The lack of the remaining files (of translation) caused the error, as I think. I couldn't find my way in the new "lib" directory.
So please, what do you precisely suggest now? Should I "close" this pull request then "open" another one against branch 'ojs-dev-2_4' ? and I suppose it would deal with "lib" directory, right?

By the way, "help" files were very helpful for beginner users to learn the basics of this system, sorry to know they would be deprecated in version 3. Those 132 files took a significant time to be translated, nevertheless, it's the owner's decision, not mine.

mtub:
Sorry, @vormia, for being so brief about the help files. If you've already translated the help files than no harm is done as OJS 2 will be around for some time. It's just that OJS 3 will be quite different from OJS 2 and there might be better ways to document features and help users while using OJS.

Regarding the link to the css file please see my comments on your pkp-lib pull request, let's decide that first and then update the OJS pull request.

mtub:
Yes, please re-submit against the ojs-dev-2_4 branch. I suggest you remove registry/locales.xml from the pull request and I will add it again with the correct path after the pkp-lib changes have been merged (I can then include the submodule change, see http://pkp.sfu.ca/wiki/index.php?title=Information_for_Developers#PKP_library_submodule_changes)

PR #492, opened 2015-05-07
Fix call-time pass by reference ##mfelczak/ojs-stable-2_4_6##

mfelczak:
In Objects for Review plugin.


asmecher:
Merged both, thanks!

PR #497, opened 2015-05-14
pkp/pkp-lib#523 Solve jQuery conflict in the ALM plugin

lepidus:
Fix pkp/pkp-lib#523

The solution that I found to the problem is to disassociate $ from jQuery 1.10 before loading the tooltip plugin. It is important that the jQuery variable has jQuery 1.10 at line 46, because the tooltip plugin registers the plugin on the jQuery found in the variable "jQuery" . After the plugin is loaded, we can also disassociate the jQuery variable (line 53).

This hack solves the problem with pdf viewing. But it is not a generic solution because the jQuery var remains associated to the jQuery 1.10 until the tooltip plugin is loaded.

### DIFF
--- a/plugins/generic/alm/output.tpl
+++ b/plugins/generic/alm/output.tpl
@@ -43,13 +43,14 @@
 	// Import JQuery 1.10 version, needed for the tooltip plugin
 	// that we use below. jQuery.noConflict puts the old $ back.
 	$.getScript('{$jqueryImportPath}', function() {ldelim}
-		$.getScript('{$tooltipImportPath}', function() {ldelim}
+		var jquery_1_10 = jQuery.noConflict();
+		jquery_1_10.getScript('{$tooltipImportPath}', function() {ldelim}
 			// Assign the last inserted JQuery version to a new variable, to avoid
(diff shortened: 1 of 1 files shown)


asmecher:
@lepidus, have you seen this? https://forum.jquery.com/topic/multiple-versions-of-jquery-on-the-same-page
I suspect adding that to the ALM code will be the most compact and trouble-free solution.

gpieri:
Hi @asmecher,

This PR uses the same approach from this link. We use the noConflict() in the ALM to disassociate jQuery 1.10 from the global variables "$" and "jQuery". 

Actually, the ALM plugin already uses the noConflit function. 

The problem is that the noConflict is called after loading "bootstrap.tooltip.min.js". 

As the call to getScript is asynchronous, there is a window of opportunity for javascript dependent on the old jQuery to be run (e.g. checking the browser in PDF). Which leads to the problem.

Therefore, in the PR we call noConflict right after the browser finishes loading jQuery 1.10.

asmecher:
@gpieri, thanks, I should've looked closer. I've assigned the issue to @beghelli for a look.

PR #499, opened 2015-05-20
pkp/pkp-lib#503 MetadataExport Plugin created

felixgruender:
pkp/pkp-lib#503 Export tool for metadata of a journal/issue/article.
Supported formats (up to now): BibTeX, MARC XML, RDF, RIS.


pages/metadataExport/MetadataExportHandler.inc.php
```diff
@@ -0,0 +1,158 @@
...
+	
+	
+	/**
+	 * @function MetadataExportHandler - Constructor
+	 * @param $request Request
+	 */
+	function MetadataExportHandler($request) {
+		parent::ManagerHandler($request);
+		
+		$this->exportFunction = '_' . $request->getUserVar('exportScope');
```
asmecher: By convention, we're using "_" as a prefix for private functions, and here it's used to expose functions for arbitrary execution. I'd rather we didn't fetch function names from request variables -- there's a lot of security risk for not much benefit.

plugins/blocks/metadataExport/block.tpl
```diff
@@ -0,0 +1,49 @@
...
+	<br />
+	<form action="{url page="metadataExport"}">
+
+	{if $journalId}
+		<span class="blockTitle">{translate key="plugins.block.metadataExport.title"}</span>
+		<input type='radio' name='exportScope' value='exportJournal' checked />{translate key="plugins.block.metadataExport.exportJournal"}<br />
+	{/if}
+
+	{if $articleId || $issueId}
+		<input type='radio' name='exportScope' value={$op} />{translate key="plugins.block.metadataExport.$op"}<br />
```
asmecher: Be careful with unquoted / unescaped template variables -- should be `value="{$op|escape}"` for safety/consistency.

plugins/blocks/metadataExport/block.tpl
```diff
@@ -0,0 +1,49 @@
...
+		<select name='exportPluginName'>
+			{foreach from=$metadataExportPlugins item=thisMetadataExportPlugin}
+				<option 
+					{if $metadataExportPlugin && $metadataExportPlugin == $thisMetadataExportPlugin->getName()}selected="selected" {/if}
+					 value="{$thisMetadataExportPlugin->getName()|escape}">{$thisMetadataExportPlugin->getMetadataExportFormatName()|escape}
+				</option>
+			{/foreach}
+		</select>
+		
+		<input type='hidden' name='elementId' value={$elementId} />
```
asmecher: As above, and elsewhere -- be sure to quote and escape template variables.

plugins/blocks/metadataExport/metadataExportBlockPlugin.inc.php
```diff
@@ -0,0 +1,102 @@
...
+	function MetadataExportBlockPlugin() {
+		$this->metadataExportPlugins = PluginRegistry::loadCategory('metadataExportFormats');
+		uasort($this->metadataExportPlugins, create_function('$a, $b', 'return strcmp($a->getDisplayName(), $b->getDisplayName());'));
+	}
+
+	/**
+	 * Determine whether the plugin is enabled. Overrides parent so that
+	 * the plugin will be displayed during install.
+	 */
+	function getEnabled() {
```
asmecher: I think the installed checks were borrowed from another plugin; if this plugin isn't to be active until the system is installed, remove them.

asmecher:
Thanks, @felixgruender. This still needs some structural work, I think -- adding a whole new plugin category isn't justified for this case, and it would be better to decrease the amount of code duplication. I'd suggest working with the existing import/export plugin category to see whether it's possible to add a new one that presents the existing options for citation format plugins for export. Generally a single import/export plugin should be enough, I think, although it looks like you've got some additional requirements (e.g. sidebar block) that might make a generic plugin more suitable.

felixgruender:
As this export tool should be accessible via a sidebar block, I made a generic plugin (so the whole code resides now in plugins/generic/metadataExport).

plugins/generic/metadataExport/MetadataExportBlockPlugin.inc.php
```diff
@@ -0,0 +1,86 @@
...
+	var $metadataExportPlugins;
+
+	
+	/**
+	 * @function MetadataExportBlockPlugin - Constructor
+	 * @param $parentPluginName String
+	 */
+	function MetadataExportBlockPlugin($parentPluginName) {
+		$this->parentPluginName = $parentPluginName;
+		$this->metadataExportPlugins = PluginRegistry::loadCategory('generic/metadataExport/metadataExportFormats');
```
asmecher: Off the top of my head, I suspect this would be better done not when the plugin is instantiated, but when the metadata export plugins are needed. This is a fairly heavyweight operation.

plugins/generic/metadataExport/MetadataExportBlockPlugin.inc.php
```diff
@@ -0,0 +1,86 @@
...
+ * @ingroup plugins_generic_metadataExport
+ *
+ * @brief Class for block component of metadata export plugin
+ */
+
+import('lib.pkp.classes.plugins.BlockPlugin');
+
+class MetadataExportBlockPlugin extends BlockPlugin {
+	
+	/* @var parentPluginName String - name of the parent plugin object */
```
asmecher: For consistency (we're not perfect in this regard), `@var` declarations should be...

    /** @var string Name of the parent plugin object */
    var $parentPluginName;

Note that the variable name is only specified once.

plugins/generic/metadataExport/MetadataExportBlockPlugin.inc.php
```diff
@@ -0,0 +1,86 @@
...
+	var $parentPluginName;
+	
+	/* @var metadataExportPlugins Array - all plugins of the metadataexport category */
+	var $metadataExportPlugins;
+
+	
+	/**
+	 * @function MetadataExportBlockPlugin - Constructor
```
asmecher: For consistency elsewhere, don't use the `@function` directive; just describe the function in the first line. (We're using Doxygen for self-documentation.)

plugins/generic/metadataExport/MetadataExportBlockPlugin.inc.php
```diff
@@ -0,0 +1,86 @@
...
+	 * @param $parentPluginName String
+	 */
+	function MetadataExportBlockPlugin($parentPluginName) {
+		$this->parentPluginName = $parentPluginName;
+		$this->metadataExportPlugins = PluginRegistry::loadCategory('generic/metadataExport/metadataExportFormats');
+		uasort($this->metadataExportPlugins, create_function('$a, $b', 'return strcmp($a->getDisplayName(), $b->getDisplayName());'));
+	}
+	
+	/**
+	 * @see PKPPlugin::getTemplatePath()
```
asmecher: Wherever you're not changing the `@param` or `@return` self-documentation, use `@copydoc` instead of `@see`. (The current codebase is not terribly consistent here.)

plugins/generic/metadataExport/MetadataExportHandler.inc.php
```diff
@@ -0,0 +1,167 @@
...
+				assert(false);
+		}
+	}
+	
+	/**
+	 * @function _exportArticle - Export current article
+	 */
+	function _exportArticle() {
+		$articleId = $this->elementId;
+		$article = $this->publishedArticleDao->getPublishedArticleByArticleId($articleId);
```
asmecher: Beware of missing referential integrity checks; most DAO getter functions include optional additional parameters that do this for you, e.g. the second parameter to `getPublishedArticleByArticleId` allows for the optional provisoin of a journal ID. Without that, a Journal Manager could potentially grab data from another journal they don't manage.

plugins/generic/metadataExport/MetadataExportHandler.inc.php
```diff
@@ -0,0 +1,167 @@
...
+		if ($this->exportPlugin->isXML()) {
+			$this->exportPlugin->createXmlFile($this->journal, $this->articles, $this->fileName);
+		} else {
+			$this->exportPlugin->createTextFile($this->journal, $this->articles, $this->fileName);
+		}
+	}
+	
+	/**
+	 * @function _addIssueData - Add some issue data to the article object
+	 * @param $article ArticleDAO object
```
asmecher: The types here are wrong -- these are Article and Issue objects, not ArticleDAO/IssueDAO.

plugins/generic/metadataExport/MetadataExportHandler.inc.php
```diff
@@ -0,0 +1,167 @@
...
+	 * @param $issue IssueDAO object
+	 */
+	function _addIssueData($article, $issue) {
```
asmecher: Is this being done because the Issue object isn't available somewhere? IMO it would be nicer to let the issue describe itself, rather than injecting issue metadata into the article object.

plugins/generic/metadataExport/MetadataExportPlugin.inc.php
```diff
@@ -0,0 +1,421 @@
...
+ * @brief Metadata Export plugin class
+ */
+
+import('lib.pkp.classes.plugins.GenericPlugin');
+
+class MetadataExportPlugin extends GenericPlugin {
+	
+	/* @function MetadataExportPlugin - Constructor  */
+	function MetadataExportPlugin() {
+		parent::Plugin();
```
asmecher: The parent constructor is GenericPlugin, not Plugin.

plugins/generic/metadataExport/MetadataExportHandler.inc.php
asmecher: (Generally there don't seem to be any access controls here. What about unpublished content? Might need testing.)

plugins/generic/metadataExport/metadataExportFormats/bibtex/BibtexMetadataExportPlugin.inc.php
```diff
@@ -0,0 +1,98 @@
+<?php
```
asmecher: Rather than duplicating all the existing citation format plugins in a new category, is it not possible to just re-use the existing citation format plugins?

plugins/generic/metadataExport/MetadataExportPlugin.inc.php
```diff
@@ -0,0 +1,421 @@
...
+		return __('submission.copyrightStatement', array('copyrightHolder' => $copyrightHolder, 'copyrightYear' => $copyrightYear));
+	}
+
+	/**
+	 * @function getFileName - Fill white spaces in the article/issue/journal name
+	 * @param $name String
+	 * @return String
+	 */
+	function getFileName($name) {
+		return preg_replace('/\s/', '-', $name);
```
asmecher: I don't think this will safely sanitize filenames. What about reserved names like `.` or `..` or something containing a slash? It's better, if possible, to avoid using a temporary file entirely and just generate and serve the contents up from memory. I might be missing something that motivates the use of a temporary file, though.

felixgruender:
Thanks, @asmecher,  for your comments - I tried to implement your suggestions for improvement.

However - one remark on your question "Rather than duplicating all the existing citation format plugins in a new category, is it not possible to just re-use the existing citation format plugins?":

I don't think that's possible because the existing citation plugin and my new export tool actually pursue two different targets: the existing citation plugin displays the metadata of one single article, whereas the new export tool aims at downloading all article metadata of an issue or journal (as text file or XML, depending on the format).
Hence, in the existing citation plugin the metadata of an article are processed in a smarty template (e.g. plugins/citationFormats/bibtex/citation.tpl) - this approach would be rather impractical for me (as mentioned above, I want to generate downloadable text or XML files).

asmecher:
@felixgruender, I think it would be possible to delegate the work of formatting each article to the citation format plugin, and then have this code string the resulting information together into a downloadable archive. That would result in a lot less code duplication. It might require a little bit of tweaking of the existing citation format plugins, but that would be worthwhile rather than maintaining two separate copies of the code e.g. that formats an article from OJS's objects into BibTeX. I haven't looked closely at the details, so if you're stuck on how to do it, maybe I could make a few specific suggestions?

felixgruender:
OK, I've tried now to reuse the existing citation plugins (for BibTeX and RIS, as these two are part of both the citation plugins and the new export plugin).

See the new getFileContent() function in BibTeXMetadataExportPlugin.inc.php and RisMetadataExportPlugin.inc.php.
(I also made very little changes/additions in plugins/citationFormats/bibtex/citation.tpl and plugins/citationFormats/proCite/citation.tpl).

PR #500, opened 2015-05-22
Present user email validation notification when registering

davidjb:
This ensures no confusion when a user needs to verify their email address in order to proceed with registration for a journal.  Previously, OJS simply redirected the user home, with no indication at all as to what happened.  This now presents the user with a page explaining what to do next -- eg check and validate the email in their inbox.

### DIFF
--- a/pages/user/RegistrationHandler.inc.php
+++ b/pages/user/RegistrationHandler.inc.php
@@ -78,10 +78,18 @@ class RegistrationHandler extends UserHandler {
 
 		if ($regForm->validate()) {
 			$regForm->execute();
-			if (Config::getVar('email', 'require_validation')) {
-				// Send them home; they need to deal with the
-				// registration email.
-				$request->redirect(null, 'index');
+			if (!Validation::isLoggedIn()) {
+				if (Config::getVar('email', 'require_validation')) {
+					// Inform the user that they need to deal with the
+					// registration email.
+					$this->setupTemplate($request, true);
+					$templateMgr =& TemplateManager::getManager();
+					$templateMgr->assign('pageTitle', 'user.register.emailValidation');
+					$templateMgr->assign('errorMsg', 'user.register.emailValidationDescription');
+					$templateMgr->assign('backLink', $request->url(null, 'login'));
+					$templateMgr->assign('backLinkLabel', 'user.login');
+					return $templateMgr->display('common/error.tpl');
+				}
 			}
 
 			$reason = null;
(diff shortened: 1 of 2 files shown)


pages/user/RegistrationHandler.inc.php
```diff
@@ -78,10 +78,19 @@ function registerUser($args, &$request) {
...
+					// Send them home; they need to deal with the
+					// registration email.
+					$request->redirect(null, 'index');
+					$this->setupTemplate($request, true);
+					$templateMgr =& TemplateManager::getManager();
+					$templateMgr->assign('pageTitle', 'user.register.emailValidation');
+					$templateMgr->assign('errorMsg', 'user.register.emailValidationDescription');
+					$templateMgr->assign('backLink', $request->url(null, 'login'));
+					$templateMgr->assign('backLinkLabel', 'user.login');
+					$templateMgr->display('common/error.tpl');
```
asmecher: I think you'll need a `return` statement after `$templateMgr->display(...)`. The `$request->redirect` function that used to be here takes care of preventing further processing, but `$templateMgr->display(...)` doesn't.

pages/user/RegistrationHandler.inc.php
```diff
@@ -78,10 +78,19 @@ function registerUser($args, &$request) {
...
 			$regForm->execute();
-			if (Config::getVar('email', 'require_validation')) {
-				// Send them home; they need to deal with the
-				// registration email.
-				$request->redirect(null, 'index');
+			if (!Validation::isLoggedIn()) {
+				if (Config::getVar('email', 'require_validation')) {
+					// Send them home; they need to deal with the
+					// registration email.
+					$request->redirect(null, 'index');
```
asmecher: I think this redirect will prevent the code below from executing and the message won't be displayed.

asmecher:
Hi @davidjb -- thanks for contributing! This needs a couple of fixes (see comments) and some testing.

pages/user/RegistrationHandler.inc.php
davidjb: The code below definitely executes but I'll remove the line because the redirect isn't what should be happening now.

davidjb:
Okay, changes made as per the above; thanks for the suggestions!  The code's now operating exactly like the "account disabled" message code later in the same file, so that should do it.  Let me know if there's anything else I should add or do.

asmecher:
Merged -- thanks for contributing, @davidjb!

PR #501, opened 2015-05-22
Added basic Slovacian reader localization

ojsbsb:
@mtub In one of our current projects, we have translated the basic keys for the reader interface into Slovacian. Everything other than what readers will see is not translated, though, as the editorial workflow is not used in Slovacian.


locale/sk_SK/manager.xml
```diff
@@ -0,0 +1,741 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE locale SYSTEM "../../lib/pkp/dtd/locale.dtd">
+
+<!--
+  * locale/en_US/manager.xml
```
mtub: wrong path -> locale/sk_SK/manager.xml

locale/sk_SK/locale.xml
```diff
@@ -0,0 +1,1187 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE locale SYSTEM "../../lib/pkp/dtd/locale.dtd">
+
+<!--
+  * locale/en_US/locale.xml
```
mtub: wrong path

locale/sk_SK/editor.xml
```diff
@@ -0,0 +1,222 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE locale SYSTEM "../../lib/pkp/dtd/locale.dtd">
+
+<!--
+  * locale/en_US/editor.xml
```
mtub: wrong path

plugins/blocks/fontSize/locale/sk_SK/locale.xml
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE locale SYSTEM "../../../../../lib/pkp/dtd/locale.dtd">
+
+<!--
+  * plugins/blocks/fontSize/locale/en_US/locale.xml
```
mtub: wrong path

please check other plugins, too

mtub:
Thanks, @ojsbsb. This looks good. Please re-check the path names in the header comments (see my other comments). 

Furthermore, I recommend removing the untranslated parts from the files. This makes tracking them in the future easier, and the missing parts will be shown by the translator plugin's check routine.

ojsbsb:
Updated the locale paths, sorry we missed these. 

We are now deleting the untranslated lines. Seems we accidentally looked at exactly these few files which contained both translated and untranslated lines like https://github.com/pkp/ojs/blob/ojs-dev-2_4/locale/sr_SR/manager.xml.

ojsbsb:
Done, @mtub.

mtub:
Thanks @ojsbsb ! Will merge now. Will also look at the sr_SR code you've linked to.

PR #502, opened 2015-05-22
pkp/pkp-lib#526 Backport ContextDAO::getBySetting() to JournalDAO

ctgraham:
I'm looking for this getBySetting() function in 2.x for the forthcoming SUSHI-Lite plugin.

### DIFF
--- a/classes/journal/JournalDAO.inc.php
+++ b/classes/journal/JournalDAO.inc.php
@@ -418,6 +418,29 @@ class JournalDAO extends DAO {
 	function getInsertJournalId() {
 		return $this->getInsertId('journals', 'journal_id');
 	}
+
+	/**
+	 * Get journals by setting.  Backported from master pkp-lib's ContextDAO.
+	 * @param $settingName string
+	 * @param $settingValue mixed
+	 * @param $contextId int
+	 * @return DAOResultFactory
+	 */
+	function getBySetting($settingName, $settingValue, $contextId = null) {
+		$params = array($settingName, $settingValue);
+		if ($contextId) $params[] = $contextId;
+
+		$result = $this->retrieve(
+			'SELECT * FROM journals AS c
+			LEFT JOIN journal_settings AS cs
+			ON c.journal_id = cs.journal_id'.
+			' WHERE cs.setting_name = ? AND cs.setting_value = ?' .
+			($contextId?' AND c.journal_id = ?':''),
+			$params
+		);
+
+		return new DAOResultFactory($result, $this, '_returnJournalFromRow');
+	}
 }
 
 ?>
--- a/classes/journal/Journal.inc.php
+++ b/classes/journal/Journal.inc.php
@@ -49,7 +49,7 @@ class Journal extends DataObject {
 
 	/**
 	 * Set the primary locale of this journal.
-	 * @param $locale string
+	 * @param $primaryLocale string
 	 */
 	function setPrimaryLocale($primaryLocale) {
 		return $this->setData('primaryLocale', $primaryLocale);
@@ -440,7 +440,7 @@ class Journal extends DataObject {
 	 *
 	 * @param $metricType null|integer|array metrics selection
 	 * @param $columns integer|array column (aggregation level) selection
-	 * @param $filters array report-level filter selection
+	 * @param $filter array report-level filter selection
 	 * @param $orderBy array order criteria
 	 * @param $range null|DBResultRange paging specification
 	 *


classes/journal/JournalDAO.inc.php
```diff
@@ -204,7 +204,7 @@ function deleteJournalById($journalId) {
 	 * @param $enabledOnly boolean True iff only enabled jourals wanted
 	 * @param $rangeInfo object optional
 	 * @param $sortBy JOURNAL_FIELD_... optional sorting parameter
-	 * @param $searchField JOURNAL_FIELD_... optional filter parameter
+	 * @param $searchField JOURNAL_FIELD_... optional filter parameter; DEPRECATED - use getBySetting() instead
```
asmecher: I'd rather not deprecate function parameters; if it happens one day that we remove it, the risks of hard-to-debug parameter list mismatches are pretty high. If it's that a specific `$searchField` is not supported here, best to specifically assert that it's not used, I think.

classes/journal/JournalDAO.inc.php
```diff
@@ -418,6 +422,67 @@ function resequenceJournals() {
...
+	 */
+	function getBySetting($settingName, $settingValue, $contextId = null) {
+		$params = array($settingName, $settingValue);
+		if ($contextId) $params[] = $contextId;
+
+		$result = $this->retrieve(
+			'SELECT * FROM ' . $this->_getTableName() . ' AS c
+			LEFT JOIN ' . $this->_getSettingsTableName() . ' AS cs
+			ON c.' . $this->_getPrimaryKeyColumn() . ' = cs.' . $this->_getPrimaryKeyColumn() .
+			' WHERE cs.setting_name = ? AND cs.setting_value = ?' .
```
asmecher: May as well use `INNER JOIN` rather than `LEFT JOIN`. An `a LEFT JOIN b ... WHERE b.something = ?` is equivalent to the same with `INNER JOIN`.

classes/journal/JournalDAO.inc.php
```diff
@@ -418,6 +422,67 @@ function resequenceJournals() {
...
+
+		return new DAOResultFactory($result, $this, '_fromRow');
+	}
+
+	/**
+	 * Internal function to return a Journal object from a row.  Backported from master OJS's JournalDAO.
+	 * @param $row array
+	 * @return Journal
+	 */
+	function _fromRow($row) {
```
asmecher: Why not use `$this->_returnJournalFromRow`?

classes/journal/JournalDAO.inc.php
```diff
@@ -418,6 +422,67 @@ function resequenceJournals() {
...
+
+	//
+	// Protected methods
+	//
+	/**
+	 * Get the table name for this context.  Backported from master OJS's JournalDAO.
+	 * @return string
+	 */
+	function _getTableName() {
+		return 'journals';
```
asmecher: For consistency/brevity, I think it's better just to hard-code the table/column names in the one query that uses them.

classes/journal/JournalDAO.inc.php
ctgraham: I wanted to keep getBySetting() aligned to ContextDAO::getBySetting() for the sake of porting any future changes introduced in master's ContextDAO.  Thus the raw copy, including https://github.com/pkp/ojs/pull/502/files#r30907959 and https://github.com/pkp/ojs/pull/502/files#r30908188.  Still want to make this diverge?

classes/journal/JournalDAO.inc.php
ctgraham: Since the core code doesn't reference this parameter, I wanted to dissuade other plugins in development from choosing this function instead of getBySetting().  How about I remove the trigger_error() but keep the `use getBySetting() instead` in the documention?

classes/journal/JournalDAO.inc.php
asmecher: Personally, I don't think it's worth the 4 extra functions and decrease in readability -- but that's more subjective than objective, and I'll be fine to merge either way.

classes/journal/JournalDAO.inc.php
asmecher: The `getBySetting` function will deal with entries in `journal_settings`, but not `JOURNAL_FIELD_SEQUENCE`, which is also supported here. I'd suggest benign neglect (just leaving it as is). The entire function has already been removed from the master branch, I think.

classes/journal/JournalDAO.inc.php
ctgraham: I can add a feature branch to change this in master as well.  When there is a change this small, do we prefer to have an issue assigned, or just a PR?

asmecher: Frankly, that's the kind of tweak I'd commit directly to master and skip the PR hassle entirely. Since you're just back-porting, and it's entirely a stylistic thing (I think any DB optimizer will catch this), best just leave it as is for consistency, I think.

ctgraham:
Simplified backport.

asmecher:
Thanks, @ctgraham. If you do feel your approach is better, especially on stylistic matters, I'm more than happy to let it be. I tend to get nitpicky when there's nothing important to harp on.

PR #503, opened 2015-05-25
Switch to encrypted submodule URLs

davidjb:
The ``git://`` protocol is unencrypted so the suggested way of accessing GitHub repositories is via https:// (Git's Smart HTTP transport protocol).  This change will not affect or change the way this repo is cloned, just that the submodules' code is delivered securely.

In addition, in rare cases, the ``git://`` protocol may be blocked by some firewalls since it operates on an unusual port.  In the case of HTTPS, this is far less likely to be affected by such a firewall.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,15 @@
 [submodule "lib/pkp"]
 	path = lib/pkp
-	url = git://github.com/pkp/pkp-lib
+	url = https://github.com/pkp/pkp-lib.git
 [submodule "plugins/generic/customBlockManager"]
 	path = plugins/generic/customBlockManager
-	url = git://github.com/pkp/customBlockManager
+	url = https://github.com/pkp/customBlockManager.git
 [submodule "plugins/generic/staticPages"]
 	path = plugins/generic/staticPages
-	url = git://github.com/pkp/staticPages
+	url = https://github.com/pkp/staticPages.git
 [submodule "plugins/generic/translator"]
 	path = plugins/generic/translator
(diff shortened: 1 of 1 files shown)


asmecher:
@davidjb, does git officially suggest using https:// instead of git:// somewhere? That would be a surprise to me. As we're fully open source, encrypting the code isn't a priority. Are you running into a firewall problem?

davidjb:
https://help.github.com/articles/which-remote-url-should-i-use/ -- recommends HTTPS.  In 2013, I'd raised the question about why the ``git://`` protocol was removed from their clone URL selector (right-hand side of a repository) and their support team said:

> There are several problems with the git:// protocol (completely unencrypted and open to man in the middle attacks for instance). HTTPS should be beneficial in every way for your uses (it's actually just as efficient as git:// since the SmartHTTP update 3 years ago). For the security reasons, we've decided to stop promoting the protocol. That being said, we will still support it — but you'll have to manually type in the URLs.

Man-in-the-middle / tampering / surveillance during network transfer are the biggest concerns.  Even though your code's open source, security during cloning should just as important to ensure you end up running the code you were expecting to get.  With no downside to using HTTPS, it's an easy win to mitigate these potential issues.

I've asked GitHub again to highlight their above comments on their documentation page & I'm waiting to hear back.

asmecher:
Thanks, @davidjb, that makes sense. I'll give it a crack and see if it causes any problems.

asmecher:
(I've also made the change in the pkp-lib submodule, which has submodules of its own.)

PR #504, opened 2015-05-27
Show authors country in authors list

marchitelli:
Since ERIH PLUS criteria for inclusion want "Information on author affiliation and address" https://dbh.nsd.uib.no/publiseringskanaler/erihplus/about/criteria_for_inclusion and in a mail exchange they that "It is sufficient that this information is displayed in the list Scorri l'indice degli autori (i.e. authors list)" I propose to ever show  athor's countries and not only for disambiguation purposes.

PR #505, opened 2015-05-28
Author country in authors list

marchitelli:
The same of https://github.com/pkp/ojs/pull/504 but in ojs-dev-2_4 branch


asmecher:
Thanks, @marchitelli, merged!

PR #510, opened 2015-06-02
Attempt login first during registration before checking email validation

davidjb:
Without this change, attempting to login on the Registration form with standard authentication (eg ``ImplicitAuth = Off``) & a validated account will always result in the "Email Validation Required" message appearing, preventing login.  This is because the new "is logged in" check from my PR #500 happens before login is tried.

This change resolves this by attempting login with the user's submitted credentials.  If that fails, we know they're registering for the first time so prompt them to validate their email address.

Apologies for missing this in the original PR.


asmecher:
Merged. Thanks for following up, @davidjb!

PR #511, opened 2015-06-02
Allow implicit authentication to the be optional

davidjb:
As discussed at http://pkp.sfu.ca/support/forum/viewtopic.php?p=50029sid=d391a4ca8a1c1bb46ade0612db8e7fb4#p50029 onwards.

This patch for OJS comes from an install I've set up that required simultaneous local login and Shibboleth (implicit authentication) access.  It adds the ability to specify ``implicitAuth = Optional`` in the config.inc.php file.  This handles the use case of allowing users with Shibboleth/SSO users and users with local accounts (managers, anyone who can't login via Shibboleth, etc) to both access OJS via their respective methods.

This pull request is fully-backwards compatible with existing implicitAuth settings (``On`` and ``Off``). The only one difference that will be seen by users is a bugfix on the register form if ``implicitAuth = On``; see below.

This pull request consists of the following changes:

* Any checks for implicit auth (``On`` or ``Off``) now also check for ``Optional``, ensuring code paths or template content is shown in this case.
* User validation (Validation.inc.php) checks the type of auth being used before proceeding
* Updates config.TEMPLATE.inc.php with details about implicit auth
* Register form prompts users to login first when implicit auth set as ``On``.  Previously, when implicit auth was on, a non-functioning register form was shown (bugfix).  Other changes to this file are re-indentation.
* When ``implicitAuth = Optional``, display headings on login templates and sidebar to help users distinguish between implicit auth (Shibboleth) and local login.
* Updates pkp-lib to make the login.tpl compatible with optional implicit auth (see https://github.com/pkp/pkp-lib/pull/542 for the commit).
* Adds helper text to password change template when ``implicitAuth = Optional``; password change needs to be available for local accounts but can't be used for Shibboleth login so the page explains this.
* Adds relevant localisation for new strings.

### DIFF
--- a/pages/user/RegistrationHandler.inc.php
+++ b/pages/user/RegistrationHandler.inc.php
@@ -81,8 +81,16 @@ class RegistrationHandler extends UserHandler {
 
 			$reason = null;
 
-			if (Config::getVar('security', 'implicit_auth')) {
+			$implicitAuth = strtolower(Config::getVar('security', 'implicit_auth'));
+			if ($implicitAuth === true) {
 				Validation::login('', '', $reason);
+			} elseif ($implicitAuth === IMPLICIT_AUTH_OPTIONAL) {
+				// Try both types of authentication
+				if ($regForm->getData('username') && $regForm->getData('password')) {
+					Validation::login($regForm->getData('username'), $regForm->getData('password'), $reason);
+				} else {
+					Validation::login('', '', $reason);
+				}
 			} else {
 				Validation::login($regForm->getData('username'), $regForm->getData('password'), $reason);
 			}
--- a/classes/manager/form/UserManagementForm.inc.php
+++ b/classes/manager/form/UserManagementForm.inc.php
@@ -37,7 +37,7 @@ class UserManagementForm extends Form {
 			$this->addCheck(new FormValidatorCustom($this, 'username', 'required', 'user.register.form.usernameExists', array(DAORegistry::getDAO('UserDAO'), 'userExistsByUsername'), array($this->userId, true), true));
 			$this->addCheck(new FormValidatorAlphaNum($this, 'username', 'required', 'user.register.form.usernameAlphaNumeric'));
 
-			if (!Config::getVar('security', 'implicit_auth')) {
+			if (!Config::getVar('security', 'implicit_auth') || strtolower(Config::getVar('security', 'implicit_auth')) === IMPLICIT_AUTH_OPTIONAL) {
 				$this->addCheck(new FormValidator($this, 'password', 'required', 'user.profile.form.passwordRequired'));
 				$this->addCheck(new FormValidatorLength($this, 'password', 'required', 'user.register.form.passwordLengthTooShort', '>=', $site->getMinPasswordLength()));
 				$this->addCheck(new FormValidatorCustom($this, 'password', 'required', 'user.register.form.passwordsDoNotMatch', create_function('$password,$form', 'return $password == $form->getData(\'password2\');'), array(&$this)));
@@ -112,7 +112,7 @@ class UserManagementForm extends Form {
 		);
 
 		// Send implicitAuth setting down to template
-		$templateMgr->assign('implicitAuth', Config::getVar('security', 'implicit_auth'));
+		$templateMgr->assign('implicitAuth', strtolower(Config::getVar('security', 'implicit_auth')));
 
 		$site =& Request::getSite();
 		$templateMgr->assign('availableLocales', $site->getSupportedLocaleNames());
--- a/classes/security/Validation.inc.php
+++ b/classes/security/Validation.inc.php
@@ -16,24 +16,26 @@
 import('classes.security.Role');
 import('classes.security.Hashing');
 
+define('IMPLICIT_AUTH_OPTIONAL', 'optional');
+
 class Validation {
 
 	/**
 	 * Authenticate user credentials and mark the user as logged in in the current session.
-	 * @param $username string
+	 * @param $username string authenticating user's id; null if implicit auth is happening
 	 * @param $password string unencrypted password
 	 * @param $reason string reference to string to receive the reason an account was disabled; null otherwise
 	 * @param $remember boolean remember a user's session past the current browser session
 	 * @return User the User associated with the login credentials, or false if the credentials are invalid
 	 */
 	function &login($username, $password, &$reason, $remember = false) {
-		$implicitAuth = Config::getVar('security', 'implicit_auth');
+		$implicitAuth = strtolower(Config::getVar('security', 'implicit_auth'));
 
 		$reason = null;
 		$valid = false;
 		$userDao =& DAORegistry::getDAO('UserDAO');
 
-		if ($implicitAuth) { // Implicit auth
+		if ($implicitAuth && !$username) { // Implicit auth, and not regular auth
 			if (!Validation::isLoggedIn()) {
 				PluginRegistry::loadCategory('implicitAuth');
 
--- a/classes/template/TemplateManager.inc.php
+++ b/classes/template/TemplateManager.inc.php
@@ -34,7 +34,7 @@ class TemplateManager extends PKPTemplateManager {
 		assert(is_a($router, 'PKPRouter'));
 
 		// Are we using implicit authentication?
-		$this->assign('implicitAuth', Config::getVar('security', 'implicit_auth'));
+		$this->assign('implicitAuth', strtolower(Config::getVar('security', 'implicit_auth')));
 
 		if (!defined('SESSION_DISABLE_INIT')) {
 			/**
--- a/classes/user/form/RegistrationForm.inc.php
+++ b/classes/user/form/RegistrationForm.inc.php
@@ -41,10 +41,10 @@ class RegistrationForm extends Form {
 	 */
 	function RegistrationForm() {
 		parent::Form('user/register.tpl');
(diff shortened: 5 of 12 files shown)


davidjb:
Any thoughts on this being merged?

asmecher:
Sorry, @davidjb, I lost sight of this one. Stay tuned and I'll take a look shortly.

pages/user/RegistrationHandler.inc.php
```diff
@@ -81,8 +81,13 @@ function registerUser($args, &$request) {
 
 			$reason = null;
 
-			if (Config::getVar('security', 'implicit_auth')) {
+			$implicitAuth = Config::getVar('security', 'implicit_auth');
+			if ($implicitAuth === true) {
 				Validation::login('', '', $reason);
+			} elseif ($implicit_auth === 'Optional') {
```
asmecher: Looks like a typo... `$implicitAuth` vs `$implicit_auth`.

davidjb: Thanks, fixed!

asmecher:
@davidjb, are there any changes required to the lib/pkp submodule? If not, could you remove the submodule commit from the PR?

@ctgraham, IIRC you're working with Shibboleth over there -- would you be able to take a look at this too? I've used Shibboleth exactly once, and didn't use the built-in auth support in OJS.

davidjb:
@asmecher Yes, there's changes needed for lib/pkp.  The PR over there is https://github.com/pkp/pkp-lib/pull/542.

ctgraham:
@davidjb , can you rebase onto ojs-dev-2_4 to resolve the merge conflicts?  If so, I can test this PR.  We don't actively use Shib for our journals, but I do have a Shib-enabled VM for testing.

davidjb:
@ctgraham Rebased and merged.  The end result is that with `implicit_auth = Off` the functionality will remain the same; with `implicit_auth = On`, the standard login forms are hidden and registration form hidden until the user is logged in; and with `implicit_auth = Optional`, both aspects are shown at the same time.  Users can either login via Shibboleth or use the local login.

@asmecher I've rebased the PR for pkp-lib onto the current submodule checkout for OJS (see updated https://github.com/pkp/pkp-lib/pull/542).  No changes were required and the PR still merges fine; just that it's now up to date with what OJS needs.

davidjb:
FYI the Travis jobs will fail because of the change in pkp-lib submodule since my commit can't (yet) be found on that repo.

ctgraham:
Thanks, I'll check this out.

FWIW, you can tell Travis to use your submodule by including it in double-hashes in the commit message:
"Allow implicit auth to the be optional ##jcu-eresearch:shibboleth-optional##"
The repo of "pkp-lib" is implied between the user and branch.

http://pkp.sfu.ca/wiki/index.php?title=Github_Documentation_for_PKP_Contributors#PKP_library_submodule_changes

ctgraham:
I pulled this into my dev install of 2.4.7, configured implicit auth in config.inc.php and tried settings of On, Off, and Optional.  Each performed as expected.

One thing I would change is making the check for "Optional" to be case-insensitive.

asmecher:
(Or personally I prefer using defined constants rather than re-coding strings. In PHP, you'd use something like...

    define('IMPLICIT_AUTH_OPTIONAL', 'Optional');

...then in the code,

    if (... == IMPLICIT_AUTH_OPTIONAL) ...

...and in Smarty...

    {if $authType == $smarty.const.IMPLICIT_AUTH_OPTIONAL} ... {/if}

Bonus points for making that case-insensitive. Just my preference. -A)

ctgraham:
+1 to @asmecher 's suggestion.

davidjb:
All sounds good, easily updated. Is there a particular file to define directives in?  This constant will need to be  accessible / defined for both OJS and PKP lib as it's used in both.

asmecher:
Hmm, perhaps in `lib/pkp/classes/security/Validation.inc.php`, though I don't fully love it.

davidjb:
@asmecher I'll take your (or your team's) advice on where to define it.  I'll hang back from making any changes to this PR until you're happy.

asmecher:
@davidjb, I couldn't spot a better place to put it. Consider these suggestions -- case insensitivity and defined constants -- stylistic; if you're happy with things as they are, say the word and I'll merge it in.

davidjb:
@asmecher It's worth making the switch to a constant -- easy enough to implement now.  The only issue I've just hit is that the `Validation.inc.php` file isn't present within `pkp-lib` on the OJS branch.  Suggestions?

ctgraham:
Yes, in 2.x, Validation is still in `classes/security/Validation.inc.php`.  This should be fine for use in the submodule.

If I'm reading correctly, the changes in lib/pkp requiring this constant are just to lib/pkp/templates/user/login.tpl.  This is generally called from templates/user/login.tpl at the application level.

In the core codebase, templates/user/login.tpl is called from PKPLoginHandler in the shared library, for example:
https://github.com/pkp/pkp-lib/blob/ojs-dev-2_4/pages/login/PKPLoginHandler.inc.php#L60
You can tell Validation is already loaded because it is called statically just above.
https://github.com/pkp/pkp-lib/blob/ojs-dev-2_4/pages/login/PKPLoginHandler.inc.php#L28

Tracing back, you'll find it included in PKPApplication:
https://github.com/pkp/pkp-lib/blob/ojs-dev-2_4/classes/core/PKPApplication.inc.php#L85
So, it should be available pretty much everywhere.

davidjb:
Rebased and all updated.  The implicit_auth setting can now be specified as any case (`OpTiOnal` is valid) and converted to lowercase before being checked against the lower-case constant, which has been made available in the suggestion.

ctgraham:
@davidjb , yes, the magic ##submodule## reference does need to be in the first line of the commit:
https://github.com/pkp/ojs/blob/master/tools/startSubmodulesTRAVIS.sh#L25

davidjb:
Thanks @ctgraham, re-pushed again.

ctgraham:
@davidjb , sorry, I totally misled you in my earlier comment.  It is ##user/branch##, not ##user:branch##.

Somehow the Travis build is still running, despite the misinterpretation:
https://travis-ci.org/pkp/ojs/jobs/89217604#L117-L119

But if it comes out with test failure, it's probably my fault, not yours.

asmecher:
@davidjb, with apologies, I've just (re)discovered a useful feature in the configuration file parser: Something that shares the name of a defined constant will get that constant's value.

https://github.com/pkp/pkp-lib/blob/ojs-2_4_7-1/classes/config/ConfigParser.inc.php#L103

OTOH this will depend on having the constant available when the config file is read, so it might mean moving the `define` into the `PKPApplication` class or similar, but it would permit the constant to be referenced directly in the configuration file.

I think it's worth tinkering with that a little, but if you're tired of rearranging the handling of the config file option, let me know and I'd be happy to work with it myself.

ctgraham:
Wouldn't the user then need to enter the config.inc.php line as exactly the constant?
```
implicit_auth = IMPLICIT_AUTH_OPTIONAL
```
That looks case sensitive and obscure, unless I'm missing something.

asmecher:
@ctgraham, what I'm thinking is that we'd do...

    define('IMPLICIT_AUTH_DISABLED', 0);
    define('IMPLICIT_AUTH_REQUIRED', 1);
    define('IMPLICIT_AUTH_OPTIONAL', 2);

...and document those options properly within `config.TEMPLATE.inc.php`. Then we'd support the old `On`/`Off` options as legacy behavior.

Where the option is used, we'd do something like...
````
switch (Config:;getVar('security', 'implicit_auth') {
    case false: // legacy config value
    case IMPLICIT_AUTH_DISABLED:
        ...
    case true: // legacy config value
    case IMPLICIT_AUTH_REQUIRED:
        ...
    case IMPLICIT_AUTH_OPTIONAL:
        ...
}
````
I find that slightly cleaner than the current approach, but if you're not fully convinced, @ctgraham, I'm happy to continue as we already have it.

asmecher:
(Note that `switch` uses loose comparison in PHP, so OTOH the `case false` and `case true` options are unnecessary, but the order of `case` statements is important.)

ctgraham:
I feel like On|Off|Optional makes more sense for an Admin configuring the site.  It seems pretty in line with our existing configuration options.

OTOH, we have a prior case for CONSTANT_LIKE_VALUES in [smtp_auth](https://github.com/pkp/ojs/blob/ojs-dev-2_4/config.TEMPLATE.inc.php#L304).

asmecher:
Sounds like we have a matter of taste; I don't feel strongly about my proposal and don't want to further bog down @davidjb's good work with semi-relevant-at-best details. Merging, happily. Thanks, @ctgraham and @davidjb.

asmecher:
This PR was cherry-picked rather than merging because of a `lib/pkp` submodule hash commit conflict. Commit:
https://github.com/pkp/ojs/commit/977630f94b3162aa104bc6f6fa5d02e8cd2262fa

davidjb:
I don't have particularly strong feelings either way on how the option looks, but the current implementation is simpler in terms of both what @ctgraham has mentioned about admin config, but also conditional statements in the underlying code.  Having a switch statement definitely makes it more obvious as to what's happening to a reader, but I could see certain areas of my PR becoming quite complicated (eg requiring duplication of code) because of the way templates are rendered or controllers called.

Thanks for implementing this PR @asmecher.

PR #512, opened 2015-06-04
pkp/pkp-lib#546 Create empty locale files via the Translator plugin

ctgraham:
https://github.com/pkp/pkp-lib/issues/546
http://forum.pkp.sfu.ca/t/missing-key-disappears-from-list-of-missing-keys-although-it-is-not-translated/1125/10


asmecher:
Thanks, @ctgraham. Note that this is no longer an issue in the rewritten translator plugin for OJS 3.x/OMP 1.x.

PR #514, opened 2015-06-05
Remove invalid second parameter from TemplateManager::addStyleSheet() call.

ctgraham:
c.f. Error message in http://forum.pkp.sfu.ca/t/error-creating-new-issue-in-ojs-2-4-6/1206

$templateManager->addStyleSheet() resolves to PKPTemplateManager::addStyleSheet(), which has no second parameter:
https://github.com/pkp/pkp-lib/blob/ojs-dev-2_4/classes/template/PKPTemplateManager.inc.php#L259

Constant STYLE_SEQUENCE_CORE does not exist in the 2.4 codebase.


asmecher:
Thanks, @ctgraham -- also cherry-picked to ojs-stable-2_4_6.

PR #524, opened 2015-06-30
Remove arbitrary maxlengths from article titles and section titles.

ctgraham:
Arbitrary maxlengths on article and section titles are a source of client complaints.  Let's get rid of them.


asmecher:
Thanks, @ctgraham. These are remnants from the days whe we were using a "title" DB column, before things went multilingual.

PR #530, opened 2015-07-06
Remove pln plugin install date from the gateway output.

ubermichael:
Fixes pkp/pkp-lib#581


asmecher:
Merged and cherry-picked into ojs-stable-2_4_6, thanks!

PR #534, opened 2015-07-06
Update pt_PT locale for Referral Plugin

josekarvalho:


mtub:
Thanks, @josekarvalho

mtub:
Cherry-picked to ojs-dev-2_4 where locale updates should go

PR #535, opened 2015-07-09
pkp/pkp-lib#739 Create new Release/Report class structure in Counter plugin; Add COUNTER class.

ctgraham:
This abstracts the counter plugin to allow multiple COUNTER releases and reports.  Release 4.1 of JR1 and AR1 are provided as examples.

This provides the foundation for SUSHI-Lite calls via the forthcoming SUSHI-Lite plugin:
https://github.com/ulsdevteam/ojs-sushiLite-plugin

### DIFF
--- a/plugins/reports/counter/CounterReportPlugin.inc.php
+++ b/plugins/reports/counter/CounterReportPlugin.inc.php
@@ -15,7 +15,10 @@
 
 define('OJS_METRIC_TYPE_LEGACY_COUNTER', 'ojs::legacyCounterPlugin');
 
+define('COUNTER_CLASS_SUFFIX', '.inc.php');
+
 import('classes.plugins.ReportPlugin');
+import('plugins.reports.counter.classes.CounterReport');
 
 class CounterReportPlugin extends ReportPlugin {
 
@@ -29,6 +32,22 @@ class CounterReportPlugin extends ReportPlugin {
 			$this->addLocaleData();
 		}
 		return $success;
+	}
+
+	/**
+	 * @see PKPPlugin::getLocaleFilename($locale)
+	 */
+	function getLocaleFilename($locale) {
+		$localeFilenames = parent::getLocaleFilename($locale);
+
+		// Add dynamic locale keys.
+		foreach (glob($this->getPluginPath() . DIRECTORY_SEPARATOR . 'locale' . DIRECTORY_SEPARATOR . $locale . DIRECTORY_SEPARATOR . '*.xml') as $file) {
+			if (!in_array($file, $localeFilenames)) {
+				$localeFilenames[] = $file;
+			}
+		}
+
+		return $localeFilenames;
 	}
 
 	/**
@@ -58,6 +77,75 @@ class CounterReportPlugin extends ReportPlugin {
 	function getTemplatePath() {
 		return parent::getTemplatePath() . 'templates/';
 	}	 
+
+	/**
+	 * Get the latest counter release
+	 * @return string
+	 */
+	function getCurrentRelease() {
+		return '4.1';
+	}
+	
+	/**
+	 * List the valid reports
+	 * Must exist in the report path as {Report}_r{release}.inc.php
+	 * @return array multidimentional array release => array( report => reportClassName )
+	 */
+	function getValidReports() {
+		$reports = array();
+		$prefix = $this->getReportPath().DIRECTORY_SEPARATOR.COUNTER_CLASS_PREFIX;
+		$suffix = COUNTER_CLASS_SUFFIX;
+		foreach (glob($prefix.'*'.$suffix) as $file) {
+			$report_name = substr($file, strlen($prefix), -strlen($suffix));
+			$report_class_file = substr($file, strlen($prefix), -strlen(COUNTER_CLASS_SUFFIX));
+			$reports[$report_name] = $report_class_file;
+		}
+		return $reports;
+	}
+
+	/**
+	 * Get a COUNTER Reporter Object
+	 * Must exist in the report path as {Report}_r{release}.inc.php
+	 * @param $report string Report name
+	 * @param $release string release identifier
+	 * @return object
+	 */
+	function getReporter($report, $release) {
+		$reportClass = COUNTER_CLASS_PREFIX.$report;
+		$reportClasspath = 'plugins.reports.counter.classes.reports.';
+		$reportPath = str_replace('.', DIRECTORY_SEPARATOR, $reportClasspath);
+		if (file_exists($reportPath.$reportClass.COUNTER_CLASS_SUFFIX)) {
+			import($reportPath.$reportClass);
+			$reporter = new $reportClass('reports', $this->getName(), $release);
+			return $reporter;
+		}
+		return false;
+	}
+
+	/**
+	 * Get classes path for this plugin.
+	 * @return string Path to plugin's classes
+	 */
+	function getClassPath() {
+		return $this->getPluginPath() . DIRECTORY_SEPARATOR . 'classes';
+	}
+	
+
+	/**
+	 * Return the report path
+	 * @return string
+	 */
+	function getReportPath() {
+		return $this->getClassPath().DIRECTORY_SEPARATOR.'reports';
+	}
+
+	/**
+	 * Return the report path
+	 * @return string
+	 */
+	function getReleasePath() {
+		return $this->getClassPath().DIRECTORY_SEPARATOR.'releases';
+	}
 
 	/**
 	 * @see PKPPlugin::isSitePlugin()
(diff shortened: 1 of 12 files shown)


ctgraham:
@asmecher , @beghelli can I push for some review on this one?  I'd like to see this for 2.4.7, and I know this is a more substantial than average change.

asmecher:
@ctgraham, I think this got missed because there's no issue attached. Is there one that should be connected to this? If not, could you create one?

ctgraham:
I've opened a new issue as https://github.com/pkp/pkp-lib/issues/739 .

plugins/reports/counter/CounterReportPlugin.inc.php
```diff
@@ -59,6 +77,101 @@ function getTemplatePath() {
...
+	 * Must exist in the report path as {Report}_r{release}.inc.php
+	 * @return array multidimentional array release => array( report => reportClassName )
+	 */
+	function getValidReports() {
+		$reports = array();
+		$prefix = $this->getReportPath().DIRECTORY_SEPARATOR;
+		foreach ($this->getValidReleases() as $release) {
+			$suffix = '_r'.str_replace('.', '_', $release).COUNTER_CLASS_SUFFIX;
+			foreach (glob($prefix.'*'.$suffix) as $file) {
+				$reports[$release][substr($file, strlen($prefix), -strlen($suffix))] = substr($file, strlen($prefix), -strlen(COUNTER_CLASS_SUFFIX));
```
beghelli: it's minor, but I think that naming (assigning to variables with meaningful names) the result of those substrings operations would help a lot to read the code.

plugins/reports/counter/CounterReportPlugin.inc.php
```diff
@@ -59,6 +77,101 @@ function getTemplatePath() {
...
+		}
+		return $reports;
+	}
+
+	/**
+	 * Get a COUNTER Reporter Object
+	 * Must exist in the report path as {Report}_r{release}.inc.php
+	 * @return object
+	 */
+	function getReporter($report, $release) {
```
beghelli: documentation for the parameters is missing

plugins/reports/counter/CounterReportPlugin.inc.php
```diff
@@ -59,6 +77,101 @@ function getTemplatePath() {
...
+		if (strpos($report, '.') !== FALSE) {
+			$report = str_replace('.', '_', $report);
+		}
+		if (strpos($release, '.') !== FALSE) {
+			$release = str_replace('.', '_', $release);
+		}
+		$reportClass = $report.'_r'.$release;
+		$reportClasspath = 'plugins.reports.counter.classes.reports.';
+		$reportPath = str_replace('.', DIRECTORY_SEPARATOR, $reportClasspath);
+		if (file_exists($reportPath.$reportClass.'.inc.php')) {
```
beghelli: I think you can use the constant you defined for .inc.php

plugins/reports/counter/CounterReportPlugin.inc.php
```diff
@@ -59,6 +77,101 @@ function getTemplatePath() {
...
+	function getReporter($report, $release) {
+		if (strpos($report, '.') !== FALSE) {
+			$report = str_replace('.', '_', $report);
+		}
+		if (strpos($release, '.') !== FALSE) {
+			$release = str_replace('.', '_', $release);
+		}
+		$reportClass = $report.'_r'.$release;
+		$reportClasspath = 'plugins.reports.counter.classes.reports.';
+		$reportPath = str_replace('.', DIRECTORY_SEPARATOR, $reportClasspath);
```
beghelli: nice, it keeps the path readable :+1:

plugins/reports/counter/CounterReportPlugin.inc.php
```diff
@@ -94,346 +207,72 @@ function setBreadcrumbs() {
...
-				if ($i== $month) {
-					$metric = $entry[STATISTICS_METRIC];
-					$currTotal += $metric;
-					if ($entry[STATISTICS_DIMENSION_FILE_TYPE] == STATISTICS_FILE_TYPE_HTML) {
-						$htmlTotal += $metric;
-					} else if ($entry[STATISTICS_DIMENSION_FILE_TYPE] == STATISTICS_FILE_TYPE_PDF) {
-						$pdfTotal += $metric;
+					// Legacy COUNTER Release 3
+					import('plugins.reports.counter.classes.reports.JR1_r3_0');
+					$r3jr1 = new JR1_r3_0('reports', $this->getName());
```
beghelli: why not using this plugin to get this report?

beghelli: Is this Legacy Counter report able to produce the same old counter reports OJS used to produce?

plugins/reports/counter/CounterReportPlugin.inc.php
```diff
@@ -94,346 +207,72 @@ function setBreadcrumbs() {
...
+					$r3jr1->_display($request);
+					return;
+				case 'fetch':
+					// Modern COUNTER Releases
+					// must provide a release, report, and year parameter
+					if ($request->getUserVar('release') && $request->getUserVar('report') && $request->getUserVar('year')) {
```
beghelli: I think you can assign the requested parameters to variables and stop using $request->getUserVar() for every time you need them...

plugins/reports/counter/CounterReportPlugin.inc.php
```diff
@@ -94,346 +207,72 @@ function setBreadcrumbs() {
...
+						// release, report and year parameters must be sane
+						if (isset($available[$request->getUserVar('release')][$request->getUserVar('report')]) && in_array($request->getUserVar('year'), $years)) {
+							// try to get the report
+							$reporter = $this->getReporter($request->getUserVar('report'), $request->getUserVar('release'));
+							if ($reporter) {
+								// default report parameters with a yearlong range
+								$xmlResult = $reporter->getMetricsXML(array(), array(STATISTICS_DIMENSION_MONTH => array('from' => $request->getUserVar('year').'01', 'to' => $request->getUserVar('year').'12')));
```
beghelli: I think you have constants for the 'from' and 'to' stats filters, if I am not wrong.

plugins/reports/counter/classes/releases/CounterRelease.inc.php
```diff
@@ -0,0 +1,254 @@
...
+	 */
+	function getTitle() {
+		assert(false);
+	}
+
+	/**
+	 * Abstract method must be implemented by subclass
+	 * @return string COUNTER XML
+	 */
+	function getMetricsXML($columns = array(), $filters = array(), $orderBy = array(), $range = null) {
```
beghelli: parameters doc is missing

plugins/reports/counter/classes/releases/CounterRelease.inc.php
```diff
@@ -0,0 +1,254 @@
...
+		$request = PKPApplication::getRequest();
+		$journal = $request->getJournal();
+		$journalId = $journal ? $journal->getJournalId() : '';
+		// If the request context is at the journal level, the dimension context id must be that same journal id
+		if ($journalId) {
+			if (isset($filters[STATISTICS_DIMENSION_CONTEXT_ID]) && $filters[STATISTICS_DIMENSION_CONTEXT_ID] != $journalId) {
+				$this->setError(new Exception(__('plugins.reports.counter.generic.exception.filter'), COUNTER_EXCEPTION_WARNING | COUNTER_EXCEPTION_BAD_FILTERS));
+			}
+			$filters[STATISTICS_DIMENSION_CONTEXT_ID] = $journalId;
+			if (isset($filters[STATISTICS_DIMENSION_ASSOC_ID]) && $filters[STATISTICS_DIMENSION_ASSOC_ID] != ASSOC_TYPE_GALLEY) {
```
beghelli: I think you wanted to check if there's any context id inside the assoc id dimension right? So you need to first make sure the assocType is related to a context, and then check the assoc id. I think the check you have here will not work properly.

plugins/reports/counter/classes/releases/CounterRelease.inc.php
```diff
@@ -0,0 +1,254 @@
...
+	protected function filterForContext($filters) {
+		$request = PKPApplication::getRequest();
+		$journal = $request->getJournal();
+		$journalId = $journal ? $journal->getJournalId() : '';
+		// If the request context is at the journal level, the dimension context id must be that same journal id
+		if ($journalId) {
+			if (isset($filters[STATISTICS_DIMENSION_CONTEXT_ID]) && $filters[STATISTICS_DIMENSION_CONTEXT_ID] != $journalId) {
+				$this->setError(new Exception(__('plugins.reports.counter.generic.exception.filter'), COUNTER_EXCEPTION_WARNING | COUNTER_EXCEPTION_BAD_FILTERS));
+			}
+			$filters[STATISTICS_DIMENSION_CONTEXT_ID] = $journalId;
```
beghelli: you fix the filter here, but you don't below. Is there any reason?

plugins/reports/counter/classes/releases/CounterRelease4_1.inc.php
```diff
@@ -0,0 +1,77 @@
...
+	 * Construct a Reports result containing the provided performance metrics
+	 * @param $columns string|array column (aggregation level) selection
+	 * @param $filters array report-level filter selection
+	 * @param $orderBy array order criteria
+	 * @param $range null|DBResultRange paging specification
+	 * @see ReportPlugin::getMetrics for more details
+	 * @return string xml
+	 */
+	function getMetricsXML($columns = array(), $filters = array(), $orderBy = array(), $range = null) {
+		$reportItems = $this->getMetrics($columns, $filters, $orderBy, $range);
```
beghelli: This class is using a method (getMetrics) that's implemented only by the class that extends it. This is not a good design approach, the parent class should not know the details of the class that implement it.

One easy solution is to add the getMetrics() method as a method into the general report class, so it's clear that child classes should implement it. It could be abstract (which requires an implementation) or we can implement something by default and leave the other details to the child classes.

plugins/reports/counter/classes/releases/CounterRelease.inc.php
```diff
@@ -0,0 +1,254 @@
+<?php
+
+/**
+ * @file plugins/generic/sushiLite/classes/releases/CounterRelease.inc.php
```
beghelli: wrong path

plugins/reports/counter/classes/releases/CounterRelease.inc.php
```diff
@@ -0,0 +1,254 @@
...
+define('COUNTER_EXCEPTION_ERROR', 1);
+define('COUNTER_EXCEPTION_PARTIAL_DATA', 4);
+define('COUNTER_EXCEPTION_NO_DATA', 8);
+define('COUNTER_EXCEPTION_BAD_COLUMNS', 16);
+define('COUNTER_EXCEPTION_BAD_FILTERS', 32);
+define('COUNTER_EXCEPTION_BAD_ORDERBY', 64);
+define('COUNTER_EXCEPTION_BAD_RANGE', 128);
+define('COUNTER_EXCEPTION_INTERNAL', 256);
+
+class CounterRelease {
```
beghelli: I would call it just CounterReport. Counter Release just adds extra noise that's not much important for a general class. Your other more specific classes (CounterRelease3, for example) can bring the extra piece of information right next to where it's implemented.

plugins/reports/counter/classes/releases/CounterRelease3_0.inc.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @class CounterRelease3_0
+ * @ingroup plugins_reports_counter
+ *
+ * @brief A COUNTER report, release 3.0
+ */
+
+import('plugins.reports.counter.classes.releases.CounterRelease');
+
+class CounterRelease3_0 extends CounterRelease {
```
beghelli: it would make much more sense to have CounterRelease3_0 extends CounterReport, in my opinion. Right now it seems that we have a CounterRelease opposed to a CounterReport object, which is not true. Everything is a CounterReport, but we use some base classes to wrap reports specific to releases, which is absolutely fine. Just a naming thing.

plugins/reports/counter/classes/reports/AR1_r4_1.inc.php
```diff
@@ -0,0 +1,197 @@
...
+	/*
+	 * Convert an OJS metrics request to a COUNTER ReportItem
+	 * @param $columns string|array column (aggregation level) selection
+	 * @param $filters array report-level filter selection
+	 * @param $orderBy array order criteria
+	 * @param $range null|DBResultRange paging specification
+	 * @see ReportPlugin::getMetrics for more details
+	 * @return object COUNTER\ReportItem
+	 */
+	function getMetrics($columns = array(), $filters = array(), $orderBy = array(), $range = null) {
```
beghelli: I think it would be better to call this method getReportItems(). getMetrics is already used in other places to get the raw stats metrics, and this is not the case here.

plugins/reports/counter/classes/reports/AR1_r4_1.inc.php
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	/*
+	 * Convert an OJS metrics request to a COUNTER ReportItem
+	 * @param $columns string|array column (aggregation level) selection
+	 * @param $filters array report-level filter selection
+	 * @param $orderBy array order criteria
+	 * @param $range null|DBResultRange paging specification
+	 * @see ReportPlugin::getMetrics for more details
+	 * @return object COUNTER\ReportItem
```
beghelli: I think you return an array filled with COUNTER\ReportItem objects right?

plugins/reports/counter/classes/reports/AR1_r4_1.inc.php
```diff
@@ -0,0 +1,197 @@
+<?php
+
+/**
+ * @file plugins/generic/sushiLite/classes/reports/AR1_r4_1.inc.php
```
beghelli: wrong path

plugins/reports/counter/classes/reports/JR1_r3_0.inc.php
```diff
@@ -0,0 +1,395 @@
...
+ * @file plugins/reports/counter/classes/reports/JR1_r3_0.inc.php
+ *
+ * Copyright (c) 2013-2015 Simon Fraser University Library
+ * Copyright (c) 2003-2015 John Willinsky
+ * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+ *
+ * @class JR1_r3_0
+ * @ingroup plugins_reports_counter
+ *
+ * @brief Counter report plugin
```
beghelli: maybe a little more specification?

plugins/reports/counter/CounterReportPlugin.inc.php
ctgraham: Yes, this legacy JR1 (r3) report is exactly the report which this plugin used to deliver, COUNTER ready numbers in the old XML or XLS (CSV, really) format.  It also supports the double-legacy "legacy counter" report which is not COUNTER ready.

ctgraham: Can you clarify your question of "why not use this plugin to get this report?"?

plugins/reports/counter/classes/releases/CounterRelease.inc.php
ctgraham: Within the context of COUNTER, there is a distinct level of abstraction that happens at the Release level, and another that happens at the Report level.  I had a hard time distinguishing where changes have happened and will happen in the future because I was only really working with Release 4.1, the retracted Release 4.2, and the legacy Release 3.0.

As an example, Release 4.2 was proposed to add a new XML element "ReportAttributes" into the "Report" element.  This would have affected all reports delivered by this COUNTER Release.  I think a set of "Release" classes is needed, even if it's not quite clear yet what will change from release to release.  Thoughts?

plugins/reports/counter/classes/releases/CounterRelease3_0.inc.php
ctgraham: I struggled with this.  Yes, everything is ultimately a CounterReport, but it is implemented in a cross-product of Releases and Reports.  JR1 has a particular definition, for example, but is not only packaged differently in Release 3.0 than it is in Release 4.1, it is also potentially *formulated* differently.

I like the idea of everything being a CounterReport.  You propose that class is then implemented as a CounterRelease*NN*, and that release is then implemented as a Report*NNXX*?

plugins/reports/counter/CounterReportPlugin.inc.php
ctgraham: Naming as $report_name and $report_class_file.

plugins/reports/counter/CounterReportPlugin.inc.php
ctgraham: Adding docs.

plugins/reports/counter/CounterReportPlugin.inc.php
ctgraham: Agreed; fixing.

plugins/reports/counter/CounterReportPlugin.inc.php
ctgraham: @beghelli , looking for your feedback here.

plugins/reports/counter/CounterReportPlugin.inc.php
ctgraham: Cleaning up.

plugins/reports/counter/CounterReportPlugin.inc.php
ctgraham: I don't think these exist (at least not yet):
master: https://github.com/pkp/ojs/blob/09042fea007be16bc76d464434bdbd49566a1163/classes/statistics/MetricsDAO.inc.php#L110-L115

dev:https://github.com/pkp/ojs/blob/f400f097c2cb8618d6cff2efdb6d5a0e6bd55805/classes/statistics/MetricsDAO.inc.php#L122-L127

Think we should add them?

plugins/reports/counter/classes/releases/CounterRelease.inc.php
ctgraham: Added docs.

plugins/reports/counter/classes/releases/CounterRelease4_1.inc.php
ctgraham: Added abstract method.

plugins/reports/counter/classes/releases/CounterRelease.inc.php
ctgraham: Fixing.

plugins/reports/counter/classes/reports/AR1_r4_1.inc.php
ctgraham: FIxing.

plugins/reports/counter/CounterReportPlugin.inc.php
beghelli: @ctgraham, the first question was before understanding the report structure you created. Please, forget it :)

plugins/reports/counter/classes/releases/CounterRelease.inc.php
beghelli: My point is that the current way you name this class is confusing. Right now the name of this class gives the idea that reports are inheriting from a release class. Conceptually that's confusing, a report is not a more specific release version. 

But if you think of this class implementing a general report one, it makes sense that every report class should have the information about it's release. If that information and details are too much to stay inside the report class, then you can create separated release classes and let the report ones use them (composition instead of inheritance). But I don't think that's the case, maybe you can decide better.

But for me, to fix that, I suggest you to create a more general report class naming, leave the general implementation as it is now, and then create more specific release reports classes (as you already do) but make sure that the name is representing a report, and not a release. That way we can avoid this confusion.

plugins/reports/counter/classes/releases/CounterRelease3_0.inc.php
beghelli: As I mentioned above, if you think you will need to use composition (implement separated release classes and let the reports use them) instead of inheritance (implement only report classes that implement themselves their releases details) than that's ok. I can't decide too much because I don't see all the requirements, but any approach we take must be implemented in a clear way to avoid confusion.

Looking at the code right now, I think we don't need to use composition. Just implement a general report class that also contains all the release interface that a report needs, than a more specific one that implements an specific release detail, and then the reports of that specific releases can all inherit from that. 

It would be something like: CounterReport, Release3CounterReport, JR1CounterReport, JR2CounterReport, etc. The two last ones inheriting from the second one, that inherits from the first one. Does that makes sense?

plugins/reports/counter/classes/releases/CounterRelease.inc.php
ctgraham: Hmmm... My memory is fuzzy on this one.  I think my concern was ensuring that a COUNTER report would always be concerned with certain types of downloads. Not only did I get the code wrong, I think the scope is wrong.  It is the individual reports that should define what types of ASSOC_IDs they care about.  The abstract report shouldn't care.  I'll remove this check.

plugins/reports/counter/CounterReportPlugin.inc.php
beghelli: Yes, but I don't think it's high priority. Let create an issue for that to be solved later on, it's easy to grep the usage and replace by a constant. Thanks.

plugins/reports/counter/classes/releases/CounterRelease3_0.inc.php
ctgraham: Yes, I like this, though I think I should get the potential for composition ready now rather than later.

JR1, AR1, etc. is a
CounterReport
  has a CounterRelease

beghelli: Sure, you know better what's coming next :)

plugins/reports/counter/classes/reports/JR1_r3_0.inc.php
ctgraham: Correcting/clarifying this description.

plugins/reports/counter/classes/reports/AR1_r4_1.inc.php
ctgraham: Fixing.

ctgraham:
TODO before pull:
* refactor CounterRelease abstract to CounterReport abstract
* rename getMetrics method to getReportItems

ctgraham:
@beghelli , this should be ready for another review.  I believe I've addressed your comments.

ctgraham:
My read of the failed check is that there was a Selenium timeout, so I think this PR is still good to go.

plugins/reports/counter/classes/CounterReport.inc.php
```diff
@@ -0,0 +1,301 @@
...
+define('COUNTER_LITERAL_ARTICLE', 'Article');
+define('COUNTER_LITERAL_JOURNAL', 'Journal');
+define('COUNTER_LITERAL_PROPRIETARY', 'Proprietary');
+
+class CounterReport {
+
+	var $parentPluginCategory;
+	var $parentPluginName;
+	
+	var $_release;
```
beghelli: sorry, not sure if it wasn't here before, but the documentation is missing for those members...

plugins/reports/counter/classes/LegacyJR1.inc.php
```diff
@@ -0,0 +1,358 @@
...
+
+		$this->_assignTemplateCounterXML($templateManager, $begin, $end, $useLegacyStats);
+		$templateManager->display($this->getParentPlugin()->getTemplatePath() . 'reportxml.tpl', 'text/xml');
+	}
+
+	/**
+	 * Get the parent plugin
+	 * @return object
+	 */
+	function &getParentPlugin() {
```
beghelli: Reading it again, I think it's an unnecessary coupling you have between this class and the plugin one. The plugin class knowing the interface of this class is ok, but this class doesn't need to know the plugin interface. It can receive just the template path as a parameter in the constructor. Also, why not implementing this as a CounterReport?

plugins/reports/counter/classes/CounterReport.inc.php
```diff
@@ -0,0 +1,301 @@
...
+	
+	var $_errors;
+	
+	/**
+	 * Constructor
+	 * @param string $parentPluginCategory
+	 * @param string $parentPluginName
+	 */
+	function CounterReport($parentPluginCategory, $parentPluginName) {
+		$this->parentPluginCategory = $parentPluginCategory;
```
beghelli: see my comment below about coupling this object too much to the plugin object. You can pass the release object and the template path as parameters and let this class knowing nothing about the parent plugin interface.

plugins/reports/counter/classes/CounterReport.inc.php
```diff
@@ -0,0 +1,301 @@
...
+	 * Abstract method must be implemented by subclass
+	 * @param $columns string|array column (aggregation level) selection
+	 * @param $filters array report-level filter selection
+	 * @param $orderBy array order criteria
+	 * @param $range null|DBResultRange paging specification
+	 * @see ReportPlugin::getMetrics for more details on parameters
+	 * @return array COUNTER\ReportItem array
+	 */
+	function getReportItems($columns = array(), $filters = array(), $orderBy = array(), $range = null) {
+		assert(false);
```
beghelli: since you're already using some PHP 5 stuff, you can declare this class and method abstracts... But not sure about using PHP 5 things like this in ojs 2.x @asmecher are we doing this now?

plugins/reports/counter/classes/CounterReport.inc.php
```diff
@@ -0,0 +1,301 @@
...
+	function getVendorLogoUrl() {
+		return $this->_getVendorComponent('logo');
+	}
+
+	/**
+	 * Get the Vendor Componet by key
+	 * @param $key string
+	 * @return mixed
+	 */
+	protected function _getVendorComponent($key) {
```
beghelli: underscore just for private members...

plugins/reports/counter/classes/CounterReport.inc.php
asmecher: Generally speaking, no, we aren't using PHP5 stuff in OJS 2.x. I've gotten a lot more lax about it because of the low likelihood that anyone is running PHP4 on a modern version of OJS 2.x, but all the same, don't go out of your way to replace PHP4 constructs with PHP5 on the OJS2 codebase -- that's being done for OJS3.

plugins/reports/counter/index.php
```diff
@@ -16,8 +16,13 @@
  *
  */
 
+// Because of the use of Namespaces, this plugin now requires PHP 5.3 or better
+if (version_compare(PHP_VERSION, '5.3.0') >= 0) {
```
beghelli: but the legacy then will not work...

plugins/reports/counter/classes/CounterReport.inc.php
ctgraham: This PHP4/5 hybrid was the result of this pkp-dev conversation:

From: Graham, Clinton T 
Sent: Monday, January 05, 2015 2:36 PM
To: pkp-dev
Subject: Exceptions and OJS 2.x codebase

> I need a way to pass error conditions from one plugin (reports/counter) to another (generic/sushiLite). The error conditions should include a classification scheme and a human readable message.
 
> The clear way to go would be PHP5 Exceptions.  We incorporate Exceptions in the 2.4.x codebase via multiple third party libraries (ZendFramework, ADODB, htmlpurifier, lessphp, pqp, swordappv2, tinymce’s jbimages) and in our own code (tests, tools, dataverse plugin, pln plugin, sword plugin).
 
> Given the existing codebase, I’m not feeling very reluctant to make the COUNTER and SUSHI-Lite plugins PHP5 dependent as well.  I’m throwing this note out there to solicit any concerns with that approach.

From: Alec Smecher
Sent: Monday, January 05, 2015 2:44 PM
To: Graham, Clinton T; pkp-dev
Subject: Re: Exceptions and OJS 2.x codebase

> Would this just be in plugins? If so, please go for it -- and if you're giving some thought to exception handling in general, maybe we should put you on the spot during one of the tech committee calls to talk about establishing some best practices.

> Gotcha: watch out for the overridden error handler (set_error_handler); that's currently present to prevent PHP parse errors in situations where PHP4 and PHP5 had vastly different reference behavior.

From: Graham, Clinton T
Sent: Monday, January 05, 2015 2:55 PM
To: Alec Smecher; pkp-dev
Subject: RE: Exceptions and OJS 2.x codebase

>It would be just in the plugins.  I’m not convinced yet that I’m even going to throw the Exception from one plugin to the other; I may just instanciate and return an array of Exceptions via a getErrors() method, which would bypass the set_error_handler entirely.

asmecher: Thanks for tracking, @ctgraham. That thread still stands.

plugins/reports/counter/index.php
ctgraham: My intent here was to disable the plugin to avoid accidentally triggering fatal errors if someone was still using an obsolete version of PHP.

The [Exceptions conversation](https://github.com/pkp/ojs/pull/535/files#r44421855) would limit us to 5.1.0.

I also wanted, however, to create a reusable library which would output COUNTER XML based on structured input.  I searched unsuccessfully for an existing module or tool to transform an XSD to PHP DOMDocument calls.  I ended up writing this [COUNTER tool](https://github.com/ulsdevteam/COUNTER) instead.  I would be enthralled to have a better solution.

If we stick with this tool I wrote however, we're stuck with 5.3.0 as the minimum.  It went [end-of-life a year ago](http://php.net/eol.php), so I don't feel bad about it.

asmecher: I suggest we limit the amount of time we spend on obsolete hosting environments, even in OJS 2.x, especially in plugins. There are no perfect solutions for some of this, and our long-term solution is OJS 3 plus modern coding standards. I'm content to go ahead with @ctgraham's proposal.

plugins/reports/counter/classes/LegacyJR1.inc.php
ctgraham: I'll pass in the template path here.

The legacy report does not implement CounterReport because the core idea behind CounterReport is to transform a metrics request into a set of COUNTER\ReportItems.  These ReportItems can then be represented as XML.  COUNTER\ReportItems are now tied to the current COUNTER Release.  I had intended originally to support multiple COUNTER releases in the COUNTER classes and in the plugin, but this is unnecessary because COUNTER itself does not support old releases.  Rather, they give a deadline for all implementations to support the latest release: http://www.projectcounter.org/code_practice.html

plugins/reports/counter/classes/CounterReport.inc.php
ctgraham: Oops.  I think I should have omitted any function scope.  Just `function _getVendorComponent($key)`.

plugins/reports/counter/classes/CounterReport.inc.php
ctgraham: Adding docs.

plugins/reports/counter/classes/CounterReport.inc.php
ctgraham: Simplifying to just pass the release number.

beghelli:
Tested locally, everything works good. Merging.

PR #547, opened 2015-07-24
pkp/pkp-lib#615 Updates to CrossRef plugin

jmacgreg:


plugins/importexport/crossref/classes/CrossRefExportDom.inc.php
```diff
@@ -328,14 +328,17 @@ function &_generateJournalArticleDom(&$doc, &$journal, &$issue, &$section, &$art
...
 
 		/* AuthorList */
-		$contributorsNode =& XMLCustomWriter::createElement($doc, 'contributors');
-		$isFirst = true;
-		foreach ($article->getAuthors() as $author) {
-			$authorNode =& $this->_generateAuthorDom($doc, $author, $isFirst);
-			$isFirst = false;
-			XMLCustomWriter::appendChild($contributorsNode, $authorNode);
+		// Contributor list must not be empty
+		if ($article->getAuthors()) {
```
asmecher: Each call to `$article->getAuthors()` will hit the database; better to store it to a variable and only fetch it once.

plugins/importexport/crossref/classes/CrossRefExportDom.inc.php
```diff
@@ -328,14 +328,17 @@ function &_generateJournalArticleDom(&$doc, &$journal, &$issue, &$section, &$art
 		XMLCustomWriter::appendChild($journalArticleNode, $titlesNode);
 
 		/* AuthorList */
-		$contributorsNode =& XMLCustomWriter::createElement($doc, 'contributors');
-		$isFirst = true;
-		foreach ($article->getAuthors() as $author) {
-			$authorNode =& $this->_generateAuthorDom($doc, $author, $isFirst);
-			$isFirst = false;
-			XMLCustomWriter::appendChild($contributorsNode, $authorNode);
+		// Contributor list must not be empty
```
asmecher: Are you working with articles without authors? Could you describe the use case a little more?

plugins/importexport/crossref/templates/settings.tpl
```diff
@@ -34,9 +35,12 @@
...
 		{/if}
 
-	</ul>
 	{else}
-		{translate key="plugins.importexport.crossref.requirements.satisfied"}
+		<li>{translate key="plugins.importexport.crossref.requirements.satisfied"}</li>
+		{if $currentJournal->getSetting('onlineIssn')|escape || $currentJournal->getSetting('printIssn')|escape}
+			<li>{translate key="plugins.importexport.crossref.requirements.issnWarning" issnUrl=$issnUrl}</li>
+		{/if}
+	</ul>
```
asmecher: I think this `</ul>` is in the wrong place.

plugins/importexport/crossref/classes/CrossRefExportDom.inc.php
jmacgreg: No problem. This came from a conversation with Karl @ CrossRef - they do occasionally see empty <contributor/> nodes, which fails on their end. Technically this shouldn't be possible in OJS, as authors are mandatory (for now at least), but there are ways around it eg. via article import (I think?). There's probably a larger justification one could make for having articles without authors be possible, but that's another story.

plugins/importexport/crossref/templates/settings.tpl
jmacgreg: It looks good to me - the indent for the <li> above looks kind of weird, but that's only because it's indented to visually clarify the if statement.

plugins/importexport/crossref/classes/CrossRefExportDom.inc.php
asmecher: OK, that makes sense.

plugins/importexport/crossref/templates/settings.tpl
asmecher: It looks like it's currently...

    {if ...}
        <ul>
        ...
    {else}
        ...
        </ul>
    {/if}

Am I reading that wrong?

jalperin: i see the same thing alec does.

PR #551, opened 2015-07-28
Check copyAccessLogFileTool exec status

ctgraham:
Prevent report of success after shell failure.   Also, align `glob` syntax for cross-platform use, including Solaris.

### DIFF
--- a/tools/copyAccessLogFileTool.php
+++ b/tools/copyAccessLogFileTool.php
@@ -113,7 +113,7 @@ class CopyAccessLogFileTool extends CommandLineTool {
 
 		if ($fileMgr->fileExists($filePath, 'dir')) {
 			// Directory.
-			$filesToCopy = glob($filePath . DIRECTORY_SEPARATOR . '*.*');
+			$filesToCopy = glob($filePath . DIRECTORY_SEPARATOR . '*');
 			foreach ($filesToCopy as $file) {
 				// If a base filename is given as a parameter, check it.
 				if (count($this->argv) == 2) {
--- a/tools/copyAccessLogFileTool.php
+++ b/tools/copyAccessLogFileTool.php
@@ -175,19 +175,41 @@ class CopyAccessLogFileTool extends CommandLineTool {
 		}
 
 		// Uncompress it, if needed.
-		$gunzipPath = escapeshellarg(Config::getVar('cli', 'gunzip'));
+		$gunzipPath = Config::getVar('cli', 'gunzip');
 		if ($isCompressed) {
-			exec($gunzipPath . ' ' . $tmpFilePath);
+			if (!is_executable($gunzipPath)) {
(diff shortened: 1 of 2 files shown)


ctgraham:
Dammit.  This isn't right.  `is_executable("'/path/to/file'")` fails.

ctgraham:
Should work now.

asmecher:
Bruno, can you take care of this one?

tools/copyAccessLogFileTool.php
```diff
@@ -175,19 +175,41 @@ function _copyFile($filePath) {
 		}
 
 		// Uncompress it, if needed.
-		$gunzipPath = escapeshellarg(Config::getVar('cli', 'gunzip'));
+		$gunzipPath = Config::getVar('cli', 'gunzip');
 		if ($isCompressed) {
-			exec($gunzipPath . ' ' . $tmpFilePath);
+			if (!is_executable($egrepPath)) {
```
beghelli: I think you meant $gunzipPath right?

tools/copyAccessLogFileTool.php
```diff
@@ -175,19 +175,41 @@ function _copyFile($filePath) {
...
+			printf(__('admin.copyAccessLogFileTool.error.executingUtil', array('utilPath' => $egrepPath, 'utilVar' => 'egrep')) . "\n");
+			exit(1);
+		}
+		$egrepPath = escapeshellarg(Config::getVar('cli', 'egrep'));
+		$output = array();
+		$returnValue = 0;
 		// Each journal path is already escaped, see the constructor.
-		exec($egrepPath . " -i '" . $this->_journalPaths . "' " . escapeshellarg($tmpFilePath) . " > " . escapeshellarg($destinationPath));
+		exec($egrepPath . " -i '" . $this->_journalPaths . "' " . escapeshellarg($tmpFilePath) . " > " . escapeshellarg($destinationPath), $output, $returnValue);
+		if ($returnValue > 2) {
```
beghelli: egrep manual tells that 2 is returned on error, so we should be checking for equal or greater than 2, I think.

tools/copyAccessLogFileTool.php
```diff
@@ -175,19 +175,41 @@ function _copyFile($filePath) {
...
-			exec($gunzipPath . ' ' . $tmpFilePath);
+			if (!is_executable($egrepPath)) {
+				printf(__('admin.copyAccessLogFileTool.error.executingUtil', array('utilPath' => $gunzipPath, 'utilVar' => 'gunzip')) . "\n");
+				exit(1);
+			}
+			$gunzipPath = escapeshellarg($gunzipPath);
+			$output = array();
+			$returnValue = 0;
+			exec($gunzipPath . ' ' . $tmpFilePath, $output, $returnValue);
+			if ($returnValue > 1) {
```
beghelli: guzip returns 1 when an error occurs and 2 in case of warning, I think that this check will only get the warning one.

ctgraham:
@beghelli , correct on all comments.  Fixed in the latest push.

PR #552, opened 2015-07-30
EDITOR_DECISION_* don't include the decision taken

marcbria:
emailTemplates for EDITOR_DECISION_whatever don't include the decision taken.

This file need to keep sync with the decision status defined in editor.xml:

        <message key="editor.article.decision.accept">Accept Submission</message>
        <message key="editor.article.decision.pendingRevisions">Revisions Required</message>
        <message key="editor.article.decision.resubmit">Resubmit for Review</message>
        <message key="editor.article.decision.decline">Decline Submission</message>

PD: Editing directly from github (tabs, 8, no wrapp) to test if this also works fine to pull a request.

### DIFF
--- a/locale/en_US/emailTemplates.xml
+++ b/locale/en_US/emailTemplates.xml
@@ -438,11 +438,11 @@ Please confirm your ability to complete this vital contribution to the work of t
 <br />
 We have reached a decision regarding your submission to {$contextName}, &quot;{$submissionTitle}&quot;.<br />
 <br />
-Our decision is to:<br />
+Our decision is to: Accept Submission<br />
 <br />
 {$editorialContactSignature}<br />
 ]]></body>
-		<description>This email from the Editor or Section Editor to an Author notifies them of a final decision regarding their submission.</description>
+		<description>This email from the Editor or Section Editor to an Author notifies them of a final decision regarding their submission is "Accept Submission".</description>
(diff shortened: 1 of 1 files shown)


asmecher:
Marc, not that I'm sure this is best, but the omission was intentional. We decided that these emails were critical enough that the journal really should word the boilerplate for themselves.

marcbria:
> Marc, not that I'm sure this is best, but the omission was intentional. We decided that these emails were critical enough that the journal really should word the boilerplate for themselves.

I think the omission is not a good idea.

Just a few editors know enough OJS to play with the emailTemplates so in 90% of the cases those are keep delivered. 

For example, in our service, only one of our 30 journals noticed this and after 5 years I discovered today that all my editors were "filling" the decision by hand... typing it every time they call the templates are called.

This is not only a waste of time for them, it's also a source of inconsistency (they can make typo errors or mistaken the decision).

IMHO it's better to have a "standard" email template (including the decision) that editors could modify (both, when the template is called or at email templates) than offering "incomplete" ones.

asmecher:
OK, that makes sense. And it's not as though adding the decison here really changes much. Will merge and back-port.

jalperin:
check with John, he's the one that thought it should be as it is. I never
quite understood why.

Juan

PR #559, opened 2015-08-06
pkp/pkp-lib#653: don't try to delete a non-existent issue in the native importexport plugin's cleanup phase after error.

ctgraham:
https://github.com/pkp/pkp-lib/issues/653

I think the whole `foreach ($issues as $issue)` can be deleted, but I'm not confident.


asmecher:
Good spotting, @ctgraham -- and it does look like the entire clause is redundant because of existing issue cleanup in `cleanupFailure`. I'll remove the whole clause.

PR #566, opened 2015-08-10
small bug fixes from making the screen cast.

ubermichael:
Fix pkp/pkp-lib#676

* Use $this instead of $task (undefined) when logging messages in
   DepositPackage.
* Log the HTTP status of an error when available in DepositPackage.
* Check that $parts['fragment'] is set explicitly in PLNSettingsForm,
   to prevent an error in the logs.
* Accept staging server URLs that include paths.


asmecher:
Committed and cherry-picked. The video went over very well, by the way, @ubermichael.

PR #574, opened 2015-08-13
Change DC11SchemaArticleAdapter to "proper" format creator's tag

samueloph:
//This PR is here to be discussed, since the devteam may disagree with my considerations and I may be wrong because I've only worked with VuFind and don't know if OHS would have this same problem.

The format of the DC for creator generated for OJS was 'authorname;affiliation', this causes stranges harvests from Vufind.

For example, same authors will be treated differently in VuFind if it harvest some articles from DSpace and others from OJS because the OJS's ones will have its creators declared as 'authorname;affiliation' instead of only 'authorname'.

We've also had problems with users writing wrongs affiliations and then causing a mess in everyone who harvest from us without considering the different creator tag with ';', but this kind of problem is not meant to be a proof of concept since it's bad user input.

Currently we have a VuFind harvesting a lot of OJS in Brazil (http://oasisbr.ibict.br/vufind/), and they seem to be troubleshooting this by forcing VuFind to get the creator field only until the ';'.

Making OJS export DC with a clean creator tag should make it easier for people to harvest from it, without making changes to VuFind or other harvesters.

Also, if you open the web page's source of an article that is on OJS [the XHTML], you'll find that the creator tag there will be clean [only the names].

And I think this PR will make OJS more compliant to DC usage as seem in
http://dublincore.org/documents/usageguide/elements.shtml#creator

### DIFF
--- a/plugins/metadata/dc11/filter/Dc11SchemaArticleAdapter.inc.php
+++ b/plugins/metadata/dc11/filter/Dc11SchemaArticleAdapter.inc.php
@@ -86,13 +86,7 @@ class Dc11SchemaArticleAdapter extends MetadataDataObjectAdapter {
 		// Creator
 		$authors = $article->getAuthors();
 		foreach($authors as $author) {
-			$authorName = $author->getFullName(true);
-			$affiliation = $author->getLocalizedAffiliation();
-			if (!empty($affiliation)) {
-				$authorName .= '; ' . $affiliation;
-			}
-			$dc11Description->addStatement('dc:creator', $authorName);
-			unset($authorName);
+			$dc11Description->addStatement('dc:creator', $author->getFullName(true));
 		}
 
 		// Subject


asmecher:
Thanks, @samueloph -- yes, it looks like we're definitely not using a best practice here. One option would be to use the suggested `.` for institutional hierarchies, but I don't much like that either.

The best approach would probably be to remove the affiliation as you've done, though I suspect we originally shoehorned it in there at someone's request, so perhaps we should look for a namespace in order to do this properly. Suggestions are of course welcome.

asmecher:
BTW, I've filed this as an issue:
https://github.com/pkp/pkp-lib/issues/686

samueloph:
Thanks @asmecher, I've send this PR link to some people here in Brazil that also works with VuFind and OJS to have a look, should we concentrate the discussion on the Issue?

diegojmacedo:
Hi,
I think a valid argument, because it really is a question with two paths, as pure oai_dc has limitation on the amount of metadata and very difficult to have a complete breakdown of the item as if it is really necessary to have the entire item of information that will that group the data in oai_dc fields, as is done today in the author field and affiliation, regarding the amount of information that can is available, if you remove the affiliation information of the dc: creator, and any supplementary information may be loss in detail records.

In Vufind and harvester systems as Repox and XOAI, you can configure the mapping of fields and extract only the necessary information using XSLT and xpath, facilitating the the extraction of information in the desired fields, what needs to be done for example in Vufind just set up XSLT import file in Vufind, to make the extraction of information from the desired fields.

In the case of the Sword OJS plugin is joining as author affiliation and this is reflected in the import into the dspace. There could be at Sword metadata setting an option that enables to import additional information property mapping with DSpace scheme.

asmecher:
@mjordan, any thoughts on this WRT both best and common practices?

mjordan:
I can't see anything in the DC 1.1 documentation or in the OAI-PMH spec to justify appending the affiliation to the creator's name in general usage. If the DSpace SWORD server expects the OJS SWORD plugin to use that convention, that's an entirely platform-interaction-specific requirement. I'd vote an enthusiastic:+1: to removing the affiliation from the DC creator element except in cases where platform-to-platform interoperability expects it.

asmecher:
Thanks, @mjordan! I think we have consensus.

asmecher:
@samueloph and @diegojmacedo, I've merged the pull request and cherry-picked it over to ojs-stable-2_4_6 and ojs-dev-2_4. It'll be released with the next OJS. Thanks for contributing!

PR #575, opened 2015-08-18
pkp/pkp-lib#690: user login block should not expose a form via http if force_login_ssl is set

ctgraham:
See https://github.com/pkp/pkp-lib/issues/690

Also, move hardcoded English string to locale key.

### DIFF
--- a/plugins/blocks/user/UserBlockPlugin.inc.php
+++ b/plugins/blocks/user/UserBlockPlugin.inc.php
@@ -61,9 +61,16 @@ class UserBlockPlugin extends BlockPlugin {
 			$templateMgr->assign_by_ref('userSession', $session);
 			$templateMgr->assign('loggedInUsername', $session->getSessionVar('username'));
 			$loginUrl = Request::url(null, 'login', 'signIn');
+			// if the page is not ssl enabled, and force_login_ssl is set, this flag will present a link instead of the form
+			$forceSSL = false;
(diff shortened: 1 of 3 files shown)


plugins/blocks/user/UserBlockPlugin.inc.php
```diff
@@ -61,9 +61,16 @@ function getContents(&$templateMgr) {
 			$templateMgr->assign_by_ref('userSession', $session);
 			$templateMgr->assign('loggedInUsername', $session->getSessionVar('username'));
 			$loginUrl = Request::url(null, 'login', 'signIn');
+			// if the page is not ssl enabled, and force_login_ssl is set, this flag will present a link instead of the form
+			$forceSSL = 0;
```
asmecher: Looks good, but do you mind changing the flag to use true/false rather than numbers?

ctgraham:
Changed flag: int 0/1 to boolean false/true.

plugins/blocks/user/block.tpl
```diff
@@ -28,7 +28,9 @@
 		</ul>
 	{else}
 		{if $implicitAuth}
-			<a href="{url page="login" op="implicitAuthLogin"}">Journals Login</a>
```
asmecher: Good catch on this too.

asmecher:
Thanks, @ctgraham -- merged, and I'll also cherry-pick it to ojs-stable-2_4_6.

PR #576, opened 2015-08-18
Synchronize profile forms

ctgraham:
Synchronize common fields for user profiles across registration, my profile, and manager interfaces; add hooks to allow plugins to inject additional profile functionality; link from my profile to password change form.

### DIFF
--- a/templates/manager/people/userProfileForm.tpl
+++ b/templates/manager/people/userProfileForm.tpl
@@ -73,6 +73,8 @@
 
 <h3>{if $userId}{translate key="manager.people.editProfile"}{else}{translate key="manager.people.createUser"}{/if}</h3>
 
+{call_hook name="Templates::Manager::User::Profile"}
+
 <form id="userForm" method="post" action="{url op="updateUser"}" onsubmit="enablePasswordFields()">
 <input type="hidden" name="source" value="{$source|escape}" />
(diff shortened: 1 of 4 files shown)


asmecher:
Clinton, we did this for OMP/OJS 3.0, and it's definitely an improvement -- but there were lots of subtle differences between the forms and that led to a bunch of small bugs that needed later reconciliation. I'm hesitant to merge this in for a minor release. How well tested are these changes?

ctgraham:
I tried to push most of the differences to the individual forms.  This reorders some fields, which creates the biggest known change.  I manually verified the input names for each change and did some testing myself, but my testing was not comprehensive.  I wouldn't cherry-pick it to stable-2_4_6 without additional testing.

asmecher:
Thanks, @ctgraham, that makes sense. The improvement in code reuse here is obvious but I'm on the fence about whether it should be merged into 2.x. Do you mind if I leave this open for some consideration? I've filed an issue: https://github.com/pkp/pkp-lib/issues/696

ctgraham:
I think 2.4.7 is a good target.  I almost put "requires 2.4.7" in the readme earlier this week for https://github.com/ulsdevteam/pkp-orcidprofile/ , which is of course the motivation for this.

PR #577, opened 2015-08-18
 pkp/pkp-lib#691 Fill UserManagementForm if the user can be managed by logged in user

lepidus:
Fix  pkp/pkp-lib#691 by filling the form if the user can be managed by the current logged in user.


asmecher:
Thanks! Merged into ojs-dev-2_4 and cherry-picked into ojs-stable-2_4_6.

PR #578, opened 2015-08-24
pkp/pkp-lib#700 Clean up plugin API

asmecher:


controllers/grid/plugins/PluginGridRow.inc.php
```diff
@@ -34,7 +34,7 @@ function PluginGridRow($userRoles, $contextLevel) {
 	 * @param $plugin Plugin
 	 * @return boolean
 	 */
-	function _canEdit(&$plugin) {
+	protected function _canEdit($plugin) {
```
beghelli: I see now why underscore

PR #579, opened 2015-08-26
Proposed changes to RU locale

ojsbsb:
From one of our library's OJS projects, we have three changes to the Russian locale that might be of general interest, as they make the interface appear less formal: c78fd65


ppv1979:
These translations of terms are shorter, but unfortunately they doesn't express the same meaning as previous translations. For example, "Editorial Policies" is not the same as "About Project", etc. I think short version is better only when we have exact meaning for the term.

mtub:
These are partly matters of style, and Pavel has reviewed the changes and wasn't convinced, so I tend to follow him here. From what I understand using google translate, I got the impression that 

* the proposed change to *about.editorialPolicies* is, as @ppv1979 points out, misleading
* the proposed change to *about.focusAndScope* seems useful, @ppv1979, can you have another look?

ppv1979:
@mtub, I think that _about.focusAndScope_ translation from @ojsbsb can be used here. My version of translation is more narrow than the @ojsbsb's one: in my translation it means "themes, list of subjects", in @ojsbsb's version it means "Journal's concept", so I think the @ojsbsb's version is wider and can be more useful than mine. Let's change.

mtub:
will update that key manually

PR #584, opened 2015-08-31
Implemented improved password hashing to resolve pkp/pkp-lib#695

UTSLibraryITS:
fixes pkp/pkp-lib#695

We have implemented [password_compat](https://github.com/ircmaxell/password_compat) library for improved OJS password hashing.

Although PHP 5.3 is required for password_compat, all changes are fully backwards compatible. If the version of PHP is supported, the more secure bcrypt encryption will automatically be used.

Password reset hashes have also been improved with better hashing (and HMAC where available) and the addition of a unique salt (configurable via the config file). Reset hashes also have a basic expiration component, based off the user's last login time.

All existing \ legacy password hashes will continue to work. They will be seamlessly updated to new bcrypt-based hashes the next time the user logs in.

Please let us know if there is any issues or comments.


classes/security/password.inc.php
```diff
@@ -0,0 +1,18 @@
...
+ * @see https://github.com/ircmaxell/password_compat
+ */
+
+// password_compat uses namespaces, so ensure PHP version supports this
+if (version_compare(phpversion(), '5.3.0', '>=')) {
+	require_once(BASE_SYS_DIR . '/lib/password_compat/lib/password.php');
+}    
+
+// set a flag to indicate whether modern encryption is supported
+define('LEGACY_ENCRYPTION', !(
```
asmecher: Rather than making this a defined constant, I think it would be cleaner to make it a function in the Validation class.

dbscripts/xml/upgrade.xml
```diff
@@ -11,7 +11,7 @@
   *
   -->
 
-<install version="2.4.6.0">
+<install version="2.4.6.1">
```
asmecher: The convention that we're following for releases is this:
- Builds (2.4.6-0 to 2.4.6-1) are released periodically from the stable git branch (ojs-stable-2_4_6) so that users who aren't hosting with the git branch directly can receive periodic updates.
- Revisions (2.4.6 to 2.4.7) involve creating a new branch (ojs-stable-2_4_7), from which new builds will be periodically released.

Because of that, we typically wouldn't permit any database structure changes in builds. Those would need a new revision. So I'd suggest this be released in OJS 2.4.7 rather than 2.4.6-1.

dbscripts/xml/upgrade/2.4.6_update.xml
```diff
@@ -0,0 +1,18 @@
...
+  * Copyright (c) 2013-2015 Simon Fraser University Library
+  * Copyright (c) 2003-2015 John Willinsky
+  * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+  *
+  * 2.4.6 database updates XML file.
+  -->
+<data>
+	<!-- Increase length of password column for newer hashing algorithms -->
+	<sql>
+		<query>ALTER TABLE users MODIFY COLUMN password varchar(255)</query>
```
asmecher: The upgrade script syncs the XML schema descriptors against the database, and should automatically alter tables if a characteristic like the column length needs to change.

classes/security/password.inc.php
UTSLibraryITS: Moved the password library stuff into it's own class.

dbscripts/xml/upgrade/2.4.6_update.xml
UTSLibraryITS: Deleted 2.4.6_update.xml

classes/security/Validation.inc.php
```diff
@@ -257,7 +311,33 @@ function generatePasswordResetHash($userId) {
...
 		}
-		return substr(md5($user->getId() . $user->getUsername() . $user->getPassword()), 0, 6);
+		// create hash payload
+		$salt = Config::getVar('security', 'salt');
+		
+		// use last login time to create a basic expiry coefficient
+		$lastLogin = strtotime($user->getDateLastLogin());
+		$time = time();
+		$normalised = ($time - ($time % 3600)) + 7200;
+		$diff = floor(($normalised - $lastLogin) / 3600);
```
ctgraham: Won't this restrict reset hashes to only working within the same hour-of-the-day?  That is, if I reset my password in the 59th minute, my hash would expire in the next 60 seconds.
```
$lastLogin = 02:30:55
$time = 05:59:00
$normalized = (05:59:00 - :59:00) + 02:00:00 = 07:00:00
$diff = 07:00:00 - 02:30:55 \ 01:00:00 = 4

$lastLogin = 02:30:55
$time = 06:01:00
$normalized = (06:01:00 - :01:00) + 02:00:00 = 08:00:00
$diff = 08:00:00 - 02:30:55 \ 01:00:00 = 5
```

For clarity, it might be helpful to define 3600 and 7200 in terms of a constant PASSWORD_RESET_SECONDS_UNTIL_EXPIRY or similar.

UTSLibraryITS: Good point. Thanks for picking up on that.

I've removed the expiration component for the time being. I was trying to avoid having to change the lib/pkp stuff.
The added salt (and the use of the last login time) still adds an extra layer of security.

A proper expiry could be added with a modification to the following lines which actually use the reset hash:
- https://github.com/pkp/ojs/blob/ojs-stable-2_4_6/classes/user/form/LoginChangePasswordForm.inc.php#L35
- https://github.com/pkp/pkp-lib/blob/13fc909010d53be7957afd07f772fdd783da3fbb/pages/login/PKPLoginHandler.inc.php#L195

implementation like so:
```
// create hash
$expiry = time() + PASSWORD_RESET_SECONDS;
$data   = $user->getUsername() . $user->getPassword() . $user->getDateLastLogin() . $expiry;
$hash   = hash_hmac('sha256', $data, $salt) . ':' . $expiry;

// Example: be987f298ca0f8e7017119b9e76084a666e7181116838e4c968a8fcc02383ba9:1441181037

// ---- //

// check hash
list($hash, $expiry) = explode(':', $hash);

if ((int) $expiry > time()) {
  // not expired
  $data = $user->getUsername() . $user->getPassword() . $user->getDateLastLogin() . $expiry;
  if ($hash === hash_hmac('sha256', $data, $salt)) {
    // hash is valid
  } 
}
```

asmecher:
On submodules, this might be useful: https://pkp.sfu.ca/wiki/index.php?title=Github_Documentation_for_PKP_Contributors#PKP_library_submodule_changes

dankobiaka:
I think I worked out the submodule stuff...:S. Please let know if there is any issues.

UTSLibraryITS:
I think I worked out the submodule stuff.

Please let me know if there is any issues.

asmecher:
Thanks, @UTSLibraryITS, and FWIW we just got another similar feature request (issue https://github.com/pkp/pkp-lib/issues/736) that this should resolve. Would you mind changing the pull request to apply to ojs-dev-2_4 rather than ojs-stable-2_4_6?

UTSLibraryITS:
Hey @asmecher,

How do I change the pull request to ojs-dev-2_4?

Thanks,

Josh.

samueloph:
Regarding to https://github.com/pkp/pkp-lib/issues/736.

I think OJS should force at least the use of SHA1, this would change the PHP requirement to >=4.30, this should have no impact on userbase since PHP 4.3 is pretty old.

The problem remaining is how to change from MD5 to SHA1.

Anyway, I'm gonna go further and sugest that OJS3 requires PHP>=5.3, considering that the oldest Operating System server based still receiving support that I can think of is Debian Squeeze, released on February 6th, 2011, and that Squeeze has PHP>5.3.

It should reduce the amount of code changing. And in the same time ensure that every OJS3 has stronger (a lot more!) security than its oldest versions.

One could say that there could be people using PHP<5.3, I think that could be yes (a very very few people), but if they don't are using PHP>=5.3, well, I bet they woundn't upgrade to OJS3 anyway.

What do you guys think?

@ctgraham @asmecher @UTSLibraryITS

asmecher:
Good discussion, @samueloph.

We've been very slow to discard old techology in part because of our focus on the developing world. With OJS 3 (and OMP) we're finally discarding PHP < 5.2 and I'd be in favour of considering a bump to 5.3. For OJS 2.4.x, we still have a long legacy, and fully dropping support for MD5 is problematic because it'll require all users to reset their passwords. That's too much of an imposition for a point-level release, IMO.

I'd argue that continuing support for outdated hashes is still a good idea, as long as the defaults are sane, and if we can support them without a lot of legacy code. Again, that requirement to reset every user password will be tough -- we'll get a lot of push-back from users. I like the approach @UTSLibraryITS has proposed, where users can make a gradual shift over to the new hash, recognizing that it will never be a complete shift -- some old accounts will never have their hashes converted at a password change.

samueloph:
Sorry Alec, I mean to drop MD5 and SHA1 support only in OJS3.

This will not cause people to reset passwords, it'll just remove the need to make checks like this https://github.com/utslibrary/ojs/blob/password-fix/classes/security/Validation.inc.php#L271-L279

ctgraham:
+1 to bumping to PHP 5.3 rather than just PHP 5.2.

Also +1 to defaulting to the latest available (note in this case that would be bcrypt) methods, and supporting backwards compatibility for the older methods.

I believe this current PR will automagically migrate older md5 and sha1 hashes to bcrypt at login (not just at password change).  C.f.:
https://github.com/utslibrary/ojs/blob/password-fix/classes/security/Validation.inc.php#L134

asmecher:
Just to make sure I have it figured, are you proposing that the new behavior introduced by this PR would become the default, and that this PR's additions that gradually convert md5/sha1 hashes over to the new hashing would take care of accounts predating this approach? If so, yes, that sounds like a good solution to me. It would simplify this code and promote better security.

samueloph:
Yes @asmecher, this would also imply that OJS3 won't have this https://github.com/pkp/ojs/blob/master/config.TEMPLATE.inc.php#L248-L251 anymore.

ctgraham:
+1 to removing the encryption config.inc.php setting from master, and thanks @samueloph for illustrating the multiline select in Github links.  Didn't know about that!

asmecher:
@samueloph, wouldn't the old encryption setting still be necessary so that the legacy code would know what algorithm to use when checking un-migrated hashes?

samueloph:
Well, I would have to give a further look, but I think that its possible to check the stored hash itself as a valid MD5 or valid SHA1, thus without depending on config.inc.php.

samueloph:
@UTSLibraryITS 

I think you will have to make another PR to get the right target branch.
You could do this using your own branch of OJS but it looks like you deleted it.
So I guess you'll have to make a new fork, apply tha changes, and make the PR against the branch[es] that @asmecher asked.

asmecher:
It's possible to distinguish MD5 from SHA1 by looking at the string length, but that does make me a little bit nervous about ambiguity down the road.

UTSLibraryITS:
replaced by pkp/ojs#611

PR #585, opened 2015-09-01
Fix for the User XML Import's username creation function

jmacgreg:
The function wasn't correctly checking whether a firstname was in place before generating a username.
This was leading to a situation where imported users had "user1", "user2", etc. as their username.


asmecher:
@jmacgreg, I've cherry-picked your commit into ojs-dev-2_4:
https://github.com/pkp/ojs/commit/6393d493f191c941aa8322aa10fcb5eaadd64439

Is that all that needs doing for this entry? If so, please close it.

PR #588, opened 2015-09-02
pkp/pkp-lib#715 fix authors disambiguation scheme in browse plugin

bozana:
s. https://github.com/pkp/pkp-lib/issues/715
Pleas cherry-pick to other branches if necessary...

### DIFF
--- a/classes/article/AuthorDAO.inc.php
+++ b/classes/article/AuthorDAO.inc.php
@@ -120,9 +120,9 @@ class AuthorDAO extends PKPAuthorDAO {
 				aa.middle_name,
 				aa.last_name,
 				CASE WHEN asl.setting_value = \'\' THEN NULL ELSE SUBSTRING(asl.setting_value FROM 1 FOR 255) END AS affiliation_l,
-				asl.locale,
+				CASE WHEN asl.setting_value = \'\' THEN NULL ELSE asl.locale END AS locale,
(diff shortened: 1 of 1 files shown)


classes/article/AuthorDAO.inc.php
```diff
@@ -120,9 +120,9 @@ function &getAuthorsAlphabetizedByJournal($journalId = null, $initial = null, $r
 				aa.middle_name,
 				aa.last_name,
 				CASE WHEN asl.setting_value = \'\' THEN NULL ELSE SUBSTRING(asl.setting_value FROM 1 FOR 255) END AS affiliation_l,
-				asl.locale,
+				CASE WHEN asl.setting_value = \'\' THEN NULL ELSE asl.locale END AS locale,
```
asmecher: I think this (and below) misses the case where asl.setting_value is NULL. It should be...

    CASE WHEN asl.setting_value = \'\' OR asl.setting_value IS NULL

Also below.

bozana: Yes, I was also thinking so, but wenn asl.setting_value is NULL the locale and primary_locale are also however NULL :-) So do you still think I should consider it in the case statement?

asmecher: Ah, that makes sense.

PR #591, opened 2015-09-02
Suppress warning from custom block plugin when new custom block has no content

ctgraham:
Fetch of block content can be empty at:
https://github.com/ulsdevteam/ojs/blob/090c3f257077d9057c17c44efb0266bf16640e0a/plugins/generic/customBlockManager/CustomBlockPlugin.inc.php#L192

Don't use `array_key_exists()` if $blockContent isn't an array.


asmecher:
Merged, thanks!

PR #597, opened 2015-09-09
fix - access to submission information even self registration is disabled

juansieira:
Now, when the user doesn't have the ability to self registration, the user doesn't have access to the Online Submissions (index of about template). 

I think that users should have access to it. In "Online Submissions" template the only thing that should be cutted is the Self-Registration, but the other information could be relevant (Submission Preparation Checklist, Privacy statement ...) because the user could be registered by the Journal Manager.

submissions

### DIFF
--- a/templates/about/index.tpl
+++ b/templates/about/index.tpl
@@ -51,7 +51,7 @@
 <div id="aboutSubmissions">
 <h3>{translate key="about.submissions"}</h3>
 <ul>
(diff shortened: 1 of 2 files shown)


asmecher:
Looks good! Merged and cherry-picked over to ojs-dev-2_4 as well. Thanks for contributing, @juansieira.

PR #598, opened 2015-09-09
Correctly set date_assigned and date_notified in EditAssignments

RogerAKemp:
Currently, the date_assigned field in edit_assignments is always null and the date_notified field is set regardless if the SectionEditor is emailed or not about an edit assignment.  This change corrects this behavior by adding get/set methods for dateAssigned to the EditAssignment class and using them in the EditorAction class.

### DIFF
--- a/classes/submission/editAssignment/EditAssignment.inc.php
+++ b/classes/submission/editAssignment/EditAssignment.inc.php
@@ -125,6 +125,22 @@ class EditAssignment extends DataObject {
 	}
 
 	/**
+	 * Get date editor assigned.
+	 * @return timestamp
+	 */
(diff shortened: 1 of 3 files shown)


asmecher:
Looks good to me, though it's not going to have meaningful data for existing users. Thanks, @RogerAKemp.

RogerAKemp:
Thanks Alec.

I fixed our date_assigned field by issuing the MySQL command:

UPDATE edit_assignments SET date_assigned=date_notified;

...which does properly correct the date_assigned field in all cases (since
the date was incorrectly placed into the notified column).  The
date_notified values could be teased out of the email_log if someone really
wants the exact values for that field.
Roger

PR #599, opened 2015-09-10
pkp/pkp-lib#741 Enable Google Analytics to extend the journal settings to the site level

ctgraham:
See http://forum.pkp.sfu.ca/t/google-analytics-homepage/3561

This allows the Site Admin, when looking at a journal's Google Analytics settings, to enable those settings within the Site context as well.

PR #605, opened 2015-09-14
pkp/pkp-lib#741: Apply field labels and data sanitization to new Google Analytics options.

ctgraham:
C.f. https://github.com/pkp/ojs/pull/599/files#r39314902


asmecher:
Thanks, @ctgraham!

PR #618, opened 2015-09-22
AuthorList to support diacritics in author names

ojsbsb:
Added some code to authorIndex.tpl so that author last names that start with diacritics don't trigger unwanted h3 headings. This will be especially interesting to journals with authors coming from non-english countries. I'm writing up a forum post about this and will post the link here.

### DIFF
--- a/templates/search/authorIndex.tpl
+++ b/templates/search/authorIndex.tpl
@@ -22,7 +22,50 @@
 
 	{if $lastFirstLetter|lower != $firstLetter|lower}
 			<div id="{$firstLetter|escape}">
-		<h3>{$firstLetter|escape}</h3>
+
+		{php}
+			/**
+			 * Complex letter heading for authorIndex
+			 *
+			 * This code offers an improved handling of letter headings in the alphabetically
+			 * sorted authorIndex. While the sorting order is coming from the database, this code allows
+			 * to avoid new headings appearing when a last name's first letter contains diacritics.
+			 *
+			 * to do:
+			 *  - find a better way to deal with the settings string
+			 *  - adapt to fit in with OJS architecture
+			 */
+
+			/* Accessing variables from the smarty loop outside this php code. */
+			$firstLetter = $this->get_template_vars('firstLetter');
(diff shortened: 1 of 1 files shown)


asmecher:
@ojsbsb, thanks for contributing. There are a few ways I'd suggest tinkering with this before it'll be suitable for inclusion as core code:
- PHP should be moved into the controller code (rather than the template file) as much as possible
- The grouping used should be language-specific, not hard-coded
Let me know if I can provide any guidance with this!

bkroll:
This is the forum post I wrote for this PR: http://forum.pkp.sfu.ca/t/handling-of-author-last-names-starting-with-a-diacritic-in-author-list/4225

Thanks for your comments, @asmecher! I will work on that in the next couple of days and then post here again.

bkroll:
oh sorry: @ojsbsb and me are the same person - usually posting from @ojsbsb for OJS stuff...

bkroll:
With the last two commits I have just addressed your second comment, @asmecher, regarding the settings string. The hard-coded version was only for demonstration. The sorter setting can now be read from the currently active localisation or, if there is no such setting for the active locale, the default alphabet is used.

I'd like to move most of the PHP code somewhere else, at least the lines until $settingsArray is created. Could you give me a hint where the appropriate place would be in the OJS structure? Thanks!

asmecher:
@bkroll, thanks -- as for moving the PHP out of the template, I'd suggest tinkering with putting it into `pages/search/SearchHandler.inc.php` in the `authors` function. You'll see near the bottom of that function that this is where the `authorIndex.tpl` template you're working with is invoked.

bkroll:
Having played with this a bit, it turns out that all other solutions lead to much more lines of code than this. The solution is based on arrays, and as arrays can't be passed to smarty directly, moving PHP out of the template would lead to larger modifications.

The localization is now implemented by #6c3091c as required.

bkroll:
So, before we forget about this - is there anything else we should do to make this fit in?

PR #628, opened 2015-10-05
Use canonical URLs in sitemap

marchitelli:
When a journal uses custom identifiers for articles URL, this URL must
be used for sitemap generation; otherwise search engine harvest
non-canonical URLs.


asmecher:
Thanks, @marchitelli! Merged.

PR #631, opened 2015-10-06
Fix DataCite and CrossRef scheduled tasks error e-mails

bozana:
S. https://github.com/pkp/pkp-lib/issues/799


asmecher:
Thanks, Bozana!

PR #633, opened 2015-10-06
pkp/pkp-lib#798 and *8494* undefined index at datacite export

bozana:
S. https://github.com/pkp/pkp-lib/issues/798
This is only a fix of this issue. I believe that those DOI export/registration plug-ins need some bigger fixing in the master branch :-(


asmecher:
Code needs re-syncing from ojs-dev-2_4 rather than checking to make sure fixes are ported. Closing.

PR #634, opened 2015-10-07
pkp/pkp-lib#685 the opt-out block is not a site plugin

bozana:
S. https://github.com/pkp/pkp-lib/issues/685
This is a fix: the opt-out block plug-in should be only displayed on the journal pages.


asmecher:
@beghelli, could you take a look?

PR #635, opened 2015-10-08
Cleanup crossref locale

ctgraham:
Addresses https://github.com/pkp/pkp-lib/issues/804 by removing incorrect locale keys.


asmecher:
@mtub, could you take a peek?

PR #639, opened 2015-10-14
Import submission editor decision constants

selwyntcy:
Import the several undefined submission editor decision constants.


asmecher:
Thanks, @selwyntcy!

PR #647, opened 2015-10-28
pkp/pkp-lib#845: Reuse canBeExported() for DOIExportPlugin and CrossRefExportPlugin

ctgraham:
Resolves https://github.com/pkp/pkp-lib/issues/845, though we have to pass a disposable `$errors` array for each call.


asmecher:
Merged, thanks!

PR #649, opened 2015-10-29
pkp/pkp-lib#847 fix download file logging

beghelli:


beghelli:
@bozana can you review it? I tested the privacy function, but it would be good to have your eyes on this also. Thanks.

bozana:
@beghelli very nice! :-) Works like a charm... :-) THANKS A LOT!!!

bozana:
It should maybe also be cherry-picked to ojs-stable-2_4_7, or?

PR #651, opened 2015-10-29
Version control for OJS3

felixgruender:


felixgruender:
See also:
https://github.com/pkp/pkp-lib/pull/849

classes/article/ArticleDAO.inc.php
```diff
@@ -81,10 +81,11 @@ function getBySetting($settingName, $settingValue, $journalId = null, $rangeInfo
 	/**
 	 * Internal function to return an Article object from a row.
 	 * @param $row array
+	 * @param $submissionRevision int
 	 * @return Article
 	 */
-	function _fromRow($row) {
-		$article = parent::_fromRow($row);
+	function _fromRow($row, $submissionRevision = null) {
```
asmecher: Is this new parameter used anywhere? I don't see it.

asmecher: Never mind, still reading!

classes/article/ArticleGalley.inc.php
```diff
@@ -218,6 +218,26 @@ function getFirstGalleyFile($fileType = null, $allFiles = false) {
 
 		return null;
 	}
+	
+	/**
+	 * Retrieve not the first, but all other revisions assigned to this submission file.
+	 * @param $fileId int
+	 * @return Array
+	 */
+	function getOtherRevisions($fileId) {
```
asmecher: It's better, if possible, to include some further qualification than just fetching the file by ID. The ID isn't checked against e.g. submission ID, so this could be abused to fetch files belonging to other submissions than intended.

controllers/modals/submissionMetadata/IssueEntryHandler.inc.php
```diff
@@ -60,6 +60,10 @@ function fetch($args, $request) {
 
 		$tabContentUrl = $dispatcher->url($request, ROUTE_COMPONENT, null, 'tab.issueEntry.IssueEntryTabHandler', 'galleyMetadata', null, array('submissionId' => $submission->getId(), 'stageId' => $this->getStageId()));
 		$templateMgr->assign('tabContentUrl', $tabContentUrl);
+		
+		$submissionDao = Application::getSubmissionDAO();
```
asmecher: For code within OJS, feel free to use ArticleDAO directly. It's only necessary to avoid naming it for shared code in lib-pkp.

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
```diff
@@ -348,6 +355,27 @@ function execute($request) {
...
+			$submissionDao = Application::getSubmissionDAO();
+			if ($request->getUserVar('submissionRevision')) {
+				$revision = $request->getUserVar('submissionRevision');
+				
+				$revisionSubmissionData = $submissionDao->getById($submission->getId(), null, false, $revision);
+				// get field names of the submission tab (= getLocaleFieldNames except 'copyrightHolder') and update them
+				$submissionTabFieldNames = $submissionDao->getLocaleFieldNames();
+
+				foreach($submissionTabFieldNames as $submissionTabFieldName) {
+					if ($submissionTabFieldName == 'copyrightHolder') continue;
```
asmecher: Hmm, it's messy to have specific exemptions like this; what is this for?

plugins/themes/default/styles/articleRevisions.less
```diff
@@ -0,0 +1,32 @@
+/**
+ * @file lib/pkp/styles/articleRevisions.less
```
asmecher: This needs a little work to get it standardized alongside the rest of our Javascript, which gets linted, optionally compiled and minified, etc.

asmecher: (Oops! That comment was supposed to be attached to the .js file.)

templates/controllers/modals/submissionMetadata/form/issueEntrySubmissionReviewForm.tpl
```diff
@@ -54,3 +54,13 @@
 	{/if}
 
 </form>
+
+{if $formParams.revision && ($formParams.revision < $latestRevisionId)}
```
asmecher: Javascript should be written in external controllers, not embedded in templates.

plugins/themes/default/styles/articleRevisions.less
ojsde: I created these 2 new javascript handlers:
js/pages/ArticleDownloadHandler.js
js/controllers/tab/issueEntry/SubmissionRevisionsHandler.js

Moreover, I made some additions to this file:
js/controllers/tab/issueEntry/IssueEntryTabHandler.js

felixgruender: (I just realized that I was logged in as 'ojsde' - that was not intentional...)

bozana:
Hi Felix, I reviewed and commented your code as good as I could ;-) 
Additionally to the comments in the code I have some additional once, that I realized/found when I try to test in all on my local installation:
-- save as new version is also possible for the not published articles.
-- by editing  metadata of an older article version: not all fields are read-only (e.g. languages, subjects, disciplines, supporting agencies).
-- date published is always the same as for the first version.
-- the new keywords, that are added for a new article version, are displayed in the form also for the old article versions. also entries for languages, subjects, disciplines, supporting agencies, references.
-- when viewing the article pages, I have no possibility to see the earlier versions.
-- if the last article version doesn't have any galley files, the old article version galley file will be displayed, when viewing the article galley page.
-- when viewing the article (abstract + galley) pages, the asserion on the line 355 in the function getPublishedArticleByPubId in classes/article/PublishedArticleDAO.inc.php fails, when a public article ID is used.
-- I don't know if this is the bug in this OJS version but: if a galley or file is not approved it is still displayed on the article pages. I can double check it in the current master branch.
Thanks a lot!

felixgruender:
@bozana, you justifiably noted that the "date published" is always the same - I'm just wondering how to solve that.
This date is stored in the table "published_submissions" for which I haven't provided a version column yet - and doing so would increase the complexity, I suppose...
Instead, I'd prefer to move the "date published" to the "submission_settings" table... would that be favourable?

bozana:
@felixgruender, here I wanted to answer :-\ Thus copying it here: "Sorry for the late response! :-\ I think Alec already answered your question, right?" :-)

felixgruender:
@bozana, right, Alec has already answered this question!

PR #654, opened 2015-11-02
pkp/pkp-lib#855 fix DataCite export link for issues

bozana:
S. https://github.com/pkp/pkp-lib/issues/855
Could it be cherry-picked to ojs-dev-2_4?


asmecher:
Cherry-picked to ojs-dev-2_4, thanks!

PR #663, opened 2015-11-04
pkp/pkp-lib#869: Only show plugin install/upgrade/delete options in plugin management if filesystem permissions allow for it.

ctgraham:
Iterate each directory under plugins and examine the plugins directory itself to see if "Install" should be shown.

Poll each plugin to see if "Upgrade" and "Delete" should be shown.

### DIFF
--- a/pages/manager/PluginHandler.inc.php
+++ b/pages/manager/PluginHandler.inc.php
@@ -68,6 +68,21 @@ class PluginHandler extends ManagerHandler {
 		$templateMgr->assign('helpTopicId', 'journal.managementPages.plugins');
 
 		$site =& $request->getSite();
+		$pluginDir = dirname(dirname(dirname(__FILE__))).'/plugins/';
+		// Don't show the install link unless something can be installed somewhere
+		$preventPluginInstall = true;
+		// must be able to write into one of the existing category directories
+		foreach (glob($pluginDir.'*', GLOB_ONLYDIR) as $file) {
+			if (is_writable($file)) {
+				$preventPluginInstall = false;
+			}
+		}
+		// or create a new category directory (unlikely)
+		if (is_writable($pluginDir)) {
+			$preventPluginInstall = false;
+		}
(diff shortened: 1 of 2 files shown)


ctgraham:
Resolves https://github.com/pkp/pkp-lib/issues/869

pages/manager/PluginHandler.inc.php
```diff
@@ -68,6 +68,21 @@ function plugins($args, &$request) {
 		$templateMgr->assign('helpTopicId', 'journal.managementPages.plugins');
 
 		$site =& $request->getSite();
+		$pluginDir = dirname(dirname(dirname(__FILE__))).'/plugins/';
+		// Don't show the install link unless something can be installed somewhere
```
asmecher: This approach isn't particularly reliable; I'd prefer checking just before an uploaded plugin is installed. That way the check could be made specific and safer.

ctgraham: I agree there should be another check when the plugin is uploaded (if there isn't already).

My concern was to hide the Install link if installation was impossible.  I think that check is reliable, no?

Can you elucidate the "safer" concern?

asmecher: If there's a check on the specific plugin, then this check is good too. Re: "safer", I was worried that permissions might be mixed inside an installed plugin, potentially allowing the installation of some parts but failing on others.

ctgraham: The "safer" component is a legitimate critique of https://github.com/ulsdevteam/pkp-lib/commit/68b8bf10c082d74b82bc12bd5fd0410d48f92d08#diff-77788b18f15e41b72c24abb00cba4e07R318 for Upgrade purposes, but I think the Install should be fine.  Unless an "Install" of a existing plugin is really an "Upgrade"?  I haven't actually checked.

Not using the web interface for plugin Installs/Upgrades/Deletes leaves me a bit less familiar; I was primarily responding to recent internal and external questions about the appearance of these links in contexts when they were known not be be relevant.

PR #667, opened 2015-11-09
warn if the acron plugin is not enabled.

ubermichael:
The PLN plugin requires the Acron plugin, in order to check the service document, make deposits, and check the status of the deposits in the PLN.

Check if the Acron plugin is enabled, and show a warning on the settings form if it isn't.

Closes pkp/pkp-lib#875

### DIFF
--- a/plugins/generic/pln/PLNPlugin.inc.php
+++ b/plugins/generic/pln/PLNPlugin.inc.php
@@ -329,6 +329,12 @@ class PLNPlugin extends GenericPlugin {
 				$this->import('classes.form.PLNSettingsForm');
 				$form = new PLNSettingsForm($this, $journal->getId());
 				
+				$application =& PKPApplication::getApplication();
+				$products =& $application->getEnabledProducts('plugins.generic');
+				if( ! isset($products['acron']) && ! Config::getVar('scheduled_tasks', false)) {
+					$templateMgr->assign('acronRequired', __('plugins.generic.pln.settings.acron_required'));
+				}
+				
 				if (Request::getUserVar('save')) {
 					$form->readInputData();
 					if ($form->validate()) {
--- a/plugins/generic/pln/locale/en_US/locale.xml
+++ b/plugins/generic/pln/locale/en_US/locale.xml
@@ -18,6 +18,7 @@
 	
 	<message key="plugins.generic.pln.settings_page">PKP PLN Plugin - Settings</message>
 	<message key="plugins.generic.pln.settings">Settings</message>
+	<message key="plugins.generic.pln.settings.acron_required">The PLN plugin requires either the Acron plugin or a periodic scheduling tools such as 'cron.'</message>
 	<message key="plugins.generic.pln.settings.saved">PKP PLN settings saved.</message>
 	<message key="plugins.generic.pln.settings.terms_of_use">Terms of Use</message>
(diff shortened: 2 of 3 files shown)


ubermichael:
@asmecher Should the wording be changed to allow for server-side cron configurations? I'm not sure that is something that OJS can check for in code. Maybe just change the wording to include "This message can be ignored if you use a unix cron or similar."

asmecher:
We can't check whether cron is actually configured, but part of the instructions for setting that up involve turning on the `scheduled_tasks` option in `config.inc.php`. If that's on, it's a fairly reliable indicator that someone has at least tried to set up server-side cron for OJS.

ubermichael:
Thanks Alec. I think I've got the check for scheduled_tasks in place now.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -329,6 +329,12 @@ function manage($verb, $args, &$message, &$messageParams) {
 				$this->import('classes.form.PLNSettingsForm');
 				$form = new PLNSettingsForm($this, $journal->getId());
 				
+				$application =& PKPApplication::getApplication();
+				$products =& $application->getEnabledProducts('plugins.generic');
+				if( ! isset($products['acron']) && Config::getVar('scheduled_tasks', 'Off') == 'Off') {
```
asmecher: Config::getVar will never return 'Off', I don't think. It should return PHP booleans when On/Off/True/False (case insensitive) are used.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -329,6 +329,12 @@ function manage($verb, $args, &$message, &$messageParams) {
...
+					$templateMgr->assign('acron_required', __('plugins.generic.pln.settings.acron_required'));
```
asmecher: For consistency, use lowerCamelCase.

ubermichael:
It would be nice if the phpDoc for Config::getVar() mentioned that it returns booleans for some values. The patch is updated.

asmecher:
Fixed the Config::getVar() self-documentation.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -329,6 +329,12 @@ function manage($verb, $args, &$message, &$messageParams) {
 				$this->import('classes.form.PLNSettingsForm');
 				$form = new PLNSettingsForm($this, $journal->getId());
 				
+				$application =& PKPApplication::getApplication();
+				$products =& $application->getEnabledProducts('plugins.generic');
+				if( ! isset($products['acron']) && ! Config::getVar('scheduled_tasks', 'Off')) {
```
asmecher: The default value should be `false`, not `'Off'`, if you're expecting a `boolean` out the back.

ubermichael:
But it is 'Off' in the config file. That's confusing.

ctgraham:
@ubermichael, better to be confusing to the programmer than to the user!  :-)

ubermichael:
Will the user ever be setting a default value *in code*?

ctgraham:
I was meaning that it's good to have the clarity of "Off" or "off" in config.inc.php for the user, even if it comes at the expense of a mixed boolean-string return for us a programmers.

ubermichael:
Off is now false.

asmecher:
Merged, thanks!

PR #668, opened 2015-11-09
Update the terms of use preamble

ubermichael:
The terms of use preamble needs to be more legalish.


ubermichael:
In the future, we plan to move this into the staging software. That way we can change it the same way the terms of use are changed, without sending a PR.

asmecher:
Merged for OJS 2.4.8.

PR #670, opened 2015-11-16
Enhance usage stats settings form

ctgraham:
Resolves https://github.com/pkp/pkp-lib/issues/890.


beghelli:
this is already merged, both in stable and master. Thanks @ctgraham

PR #683, opened 2015-12-10
pkp/pkp-lib#880: compress stats archive

beghelli:


bozana:
Everything looks good to me, I think it can be merged :-)

PR #685, opened 2015-12-14
Update the message about the PLN Staging URL

ubermichael:
OJS journal managers don't need to create an account or login to
it. In general, they won't be able to do that at all.


asmecher:
Merged!

PR #686, opened 2015-12-16
New plugin mEDRA2CR

marchitelli:
plugin for DOI registration to mEDRA - Crossref endpoint

### DIFF
--- a/plugins/importexport/medra2cr/Medra2crExportPlugin.inc.php
+++ b/plugins/importexport/medra2cr/Medra2crExportPlugin.inc.php
@@ -1,0 +1,182 @@
+<?php
+
+/**
+ * @file plugins/importexport/medra2cr/Medra2crExportPlugin.inc.php
+ *
+ * Copyright (c) 2013-2015 Simon Fraser University Library
+ * Copyright (c) 2003-2015 John Willinsky
+ * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+ *
+ * @class Medra2crExportPlugin
+ * @ingroup plugins_importexport_medra2cr
+ *
+ * @brief mEDRA2cr Onix for DOI (O4DOI) export/registration plugin.
+ */
+
+
+if (!class_exists('DOIExportPlugin')) { // Bug #7848
+	import('plugins.importexport.medra2cr.classes.DOIExportPlugin');
+}
+
+// O4DOI schemas.
+define('O4DOI_ISSUE_AS_WORK_2cr', 0x01);
+define('O4DOI_ISSUE_AS_MANIFESTATION_2cr', 0x02);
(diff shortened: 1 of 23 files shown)


asmecher:
@bozana, would you mind reviewing this?

marchitelli:
@bozana @asmecher : the plugin is based on the work by bozana for the medra plugin, obviously!

bozana:
@asmecher @marchitelli, yes I will take a look...

bozana:
@marchitelli, as far as I could see there are just a few changes/differences to the old mEDRA plug-in: different error response/handling, different date format and a few new elements (ORCID, NamesBeforeKey and KeyNames), right? Thus, why is a complete new plug-in necessary? Unfortunately, this means lots of maintenance work :-( Could those changes go into the old mEDRA plug-in? E.g. which service should be used (medra or medra2cr) could be a settings/configuration option?

bozana:
@asmecher would you agree that it would be better to add the additional service into the old mEDRA plug-in instead of having a new and almost the same plug-in?

marchitelli:
@bozana @asmecher I agree, too, that is preferable to have an additional service into the old plugin, but I can not modify the old one for managing another added endpoint and some (very little) different rules because I think that some changes in the logic of the plugin are needed.
This is the principal matter because we have developed a new (almost similar) instead of modify the old one.

bozana:
@marchitelli, could you please help me and point what would be necessary to change in the old one?

asmecher:
Agreed that duplicating the code is a bad thing. Another option, if the two of you decide not to modify the existing plugin, would be to extend it -- that way you could have a second plugin but not duplicate the code. But I'm not convinced it wouldn't be better to add this functionality to the existing plugin.

ctgraham:
Strong +1 to adding an option to the existing plugin, or to extending the existing plugin.

marchitelli:
@bozana medra2cr is different from the other plugin because:
- it uses a different endpoint https://www.medra.org/servlet/ws/CRProxy
- it uses different soap request and response messages

bozana:
@marchitelli: couldn't all this be solved as different cases or functions or ... in the same plugin? For example: if "service = medra2cr" is chosen in the plugin settings, then use "medra2cr endpoint".

bozana:
Hi @marchitelli, haven't heard from you for a long time... What do you think? Would you like to try to add those changes to the current plug-in? Do you need any help?
Best!

marchitelli:
Hi @bozana, sorry for the delay. We would like to modify the plugin according to your suggestion, after the end of February. Is it ok? 
I can close this PR and open another one later?

bozana:
That's totally fine! Thanks!!!
I don't know how to proceed best with this PR, but @asmecher will surely know :-)

asmecher:
@marchitelli, either approach is fine with me, just make sure to let us know when it's ready so we don't miss it!

PR #693, opened 2015-12-31
updated locale of crossRef plugin

mtub:
* renamed an option in CrossRef plugin (see https://github.com/pkp/pkp-lib/issues/872#issuecomment-156455947)
* renamed task name in CrossRef and DataCite plugins as discussed with @beghelli and @bozana


asmecher:
Merged, thanks!

PR #699, opened 2016-01-09
pkp/pkp-lib#1004: removed unused plugin code (including translations) 

mtub:
These plugins were never part of the application's code, locale files slipped into the code and got translated. Now removed, see pkp/pkp-lib#1004


asmecher:
Merged -- thanks, @mtub.

PR #706, opened 2016-01-11
Remove the UI elements for the pln_url config variable.

ubermichael:
It isn't necessary and it causes quite a bit of confusion for the journal managers. Also removes the pln_network* locale keys and settings.xml config. 

Fixes pkp/pkp-lib#1043

### DIFF
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -506,3 +506,22 @@ deprecation_warnings = Off
 ; Log web service request information for debugging
 log_web_service_info = Off
 
+;;;;;;;;;;;;;;;;
+; PLN Settings ;
+;;;;;;;;;;;;;;;;
+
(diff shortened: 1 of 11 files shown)


asmecher:
Is there no use case for different journals in the same installation to have different PLN URLs? (I don't know of one, but this does remove support for it.)

ubermichael:
We couldn't think of one. If it becomes important to someone in the future, we could make something like this work:

    pln_url[abc] = http://...
    pln_url[def] = http://...

But at this point it doesn't seem to be worth it.

asmecher:
Agreed. Could you add a `docs/dev/release-notes/README-2.4.8 describing the new configuration options? Have a look at e.g. `docs/release-notes/README-2.3.0` for an example.

ubermichael:
Done and done.

PR #709, opened 2016-01-16

vormia:
This update matches the changes in English locale files that differ versions **2.4.5** and **2.4.7** of OJS.
Another pull request is still waiting for response in `PKP-Lib` repository.


mtub:
Thanks for these updates and for updating the registry file. There was a merge conflict because of four lines that had been removed from the pln plugin locale. Fixed it and will upload now.

PR #710, opened 2016-01-16
Add licensing terms to the deposit XML.

ubermichael:
Fixes pkp/pkp-lib#1057

### DIFF
--- a/plugins/generic/pln/classes/DepositPackage.inc.php
+++ b/plugins/generic/pln/classes/DepositPackage.inc.php
@@ -195,6 +195,30 @@ class DepositPackage {
 
 		$entry->appendChild($pkpDetails);
 		$atom->appendChild($entry);
+
+		$locale = $journal->getPrimaryLocale();
+		$license = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'license');
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'openAccessPolicy', $journal->getLocalizedSetting('openAccessPolicy', $locale)));
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'licenseURL', $journal->getSetting('licenseURL')));
+		
+		$mode = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'publishingMode');
+		switch($journal->getSetting('publishingMode')) {
+			case PUBLISHING_MODE_OPEN:
+				$mode->nodeValue = 'Open';
+				break;
+			case PUBLISHING_MODE_SUBSCRIPTION:
+				$mode->nodeValue = 'Subscription';
+				break;
+			case PUBLISHING_MODE_NONE:
+				$mode->nodeValue = 'None';
+				break;
+		}
+		$license->appendChild($mode);
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'copyrightNotice', $journal->getLocalizedSetting('copyrightNotice', $locale)));
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'copyrightBasis', $journal->getSetting('copyrightYearBasis')));
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'copyrightHolder', $journal->getSetting('copyrightHolderType')));
+		
+		$entry->appendChild($license);
 		$atom->save($atomFile);
 		
 		return $atomFile;


plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -192,9 +192,36 @@ function generateAtomDocument() {
 				$pkpDetails->setAttribute('checksumValue', md5_file($packageFile));
 				break;
 		}
-
+
```
asmecher: (accidental whitespace added)

plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -192,9 +192,36 @@ function generateAtomDocument() {
...
 		$atom->appendChild($entry);
+		
+		$license = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'license');
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'openAccessPolicy', $journal->getLocalizedSetting('openAccessPolicy')));
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'licenseURL', $journal->getSetting('licenseURL')));
+		
+		$mode = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'publishingMode');
+		$mode->setAttribute('mode', $journal->getSetting('publishingMode'));
+		switch($journal->getSetting('publishingMode')) {
+			case 0:
```
asmecher: Use the PUBLISHING_MODE_... constants defined in classes/journal/Journal.inc.php rather than magic numbers.

plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -192,9 +192,36 @@ function generateAtomDocument() {
...
+				$mode->nodeValue = 'Open';
+				break;
+			case 1:
+				$mode->nodeValue = 'Subscription';
+				break;
+			case 2:
+				$mode->nodeValue = 'None';
+				break;
+			default:
+				$mode->nodeValue = 'Unknown';
```
asmecher: Do you ever see this case in practice?

plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -192,9 +192,36 @@ function generateAtomDocument() {
 				$pkpDetails->setAttribute('checksumValue', md5_file($packageFile));
 				break;
 		}
-
+		
 		$entry->appendChild($pkpDetails);
 		$atom->appendChild($entry);
+		
+		$license = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'license');
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'openAccessPolicy', $journal->getLocalizedSetting('openAccessPolicy')));
```
asmecher: I'd recommend getting the setting in the journal's primary locale, rather than the current locale. I don't know OTOH what the execution path to trigger this looks like, but if it's piggy-backed on a user request, it may result in the license information getting picked depending on the current user's language, which would be weird. (Possibly elsewhere as well)

plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -192,9 +192,36 @@ function generateAtomDocument() {
...
+		$license->appendChild($atom->createElementNS('http://pkp.sfu.ca/SWORD', 'licenseURL', $journal->getSetting('licenseURL')));
+		
+		$mode = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'publishingMode');
+		$mode->setAttribute('mode', $journal->getSetting('publishingMode'));
```
asmecher: I don't see the value in including the magic number in the XML as well as a text equivalent. IMO the magic number should not be exposed outside of OJS.

plugins/generic/pln/classes/DepositPackage.inc.php
ubermichael: I haven't, but it is the sensible thing to do.

asmecher: I'd rather use an `assert(false);` or even `fatalError('...');` -- IMO this will introduce uncertainty.

plugins/generic/pln/classes/DepositPackage.inc.php
ubermichael: This will run from an acron job, so it will pick up the user's locale. I don't think I quite understand OJS's locale system.

I've used getSetting() for things that don't seem translatable/localizable, eg. licenseURL, copyrightYearBasis, and copyrightHolderType. Should I have used the localized setting instead?

plugins/generic/pln/classes/DepositPackage.inc.php
ubermichael: If I understand correctly, fatalError() will call die(), which will stop the package generation and deposit process. The error will be logged to the scheduledTaskLog file, but won't be shown to a user.

asmecher: On 25/01/16 09:35 AM, Michael Joyce wrote:
> If I understand correctly, fatalError() will call die(), which will 
> stop the package generation and deposit process. The error will be 
> logged to the scheduledTaskLog file, but won't be shown to a user.
The situation should never arise, so I'm not so concerned about how an 
error message is logged or presented -- just that the XML shouldn't be 
coded to represent nonexistent options.

ubermichael:
@asmecher I think this one's ready to go.

ubermichael:
@asmecher I think this one's ready to go.

plugins/generic/pln/classes/DepositPackage.inc.php
asmecher: The situation should never arise, so I'm not so concerned about how an error message is logged or presented -- just that the XML shouldn't be coded to represent nonexistent options.

plugins/generic/pln/classes/DepositPackage.inc.php
asmecher: Some settings are localized (may be different in different languages), some are not. In cases where they're localized, you're generally guaranteed to have a value in the journal's primary locale, and may optionally have a value in other locales.

When using a `getLocalizedXyz` function, it first tries the user's current locale, and if a value exists for that locale, it gets used. If not, it falls back on the value for the journal's primary locale, which should always exist.

In most import/export scenarios the user's current locale is irrelevant -- the export should represent the content the same regardless of the user's locale. So it makes sense to force the use of the journal's primary locale, which should be something like...

    $journal->getSetting('settingName', $journal->getPrimaryLocale());

plugins/generic/pln/classes/DepositPackage.inc.php
ubermichael: I removed the default.

ubermichael: Well, I think I did. Changed in the source code, committed, rebased, and pushed. But the default clause still shows up above.

SFU is having some network errors, maybe that's the problem?

asmecher:
Hmm, not sure if Github hiccoughed or what. I see the default case gone (personally would've preferred an `assert(false)`, but OK) but I had another question that I don't see anymore: are you maintaining a DTD or XSD for the deposit XML?

ubermichael:
The deposit XML is based on the SWORD 2.0 spec, which is based on an Atom profile. There's no schema for it.

asmecher:
No custom namespacing was required or anything?

ubermichael:
The custom elements are defined in the namespace http://pkp.sfu.ca/SWORD.

asmecher:
You might consider setting those out in a DTD or XSL. If the plugin is going to evolve -- which is likely -- it'll help to have this defined explicitly in something other than the codebase.

ubermichael:
In this case, the additions are so minimal that I don't think we'll need to define a DTD.

asmecher:
Merged, thanks!

PR #714, opened 2016-01-18
Make better use of the status information from the staging server.

ubermichael:
fixes pkp/pkp-lib#1045

Replace the deposit status table with one a little more comprehensible.
The deposit status functions now return locale keys which are slightly
more descriptive than YES/NO.

 * Also uses common.* for translation keys where applicable.
 * Separates the status into local (packaged, transferred), processing
   (received, validated, sent to lockss), and LOCKSS (received, syncing,
   agreement).
 * Adds locale keys as necessary
 * Removes redundant locale keys.
 * Softens the table display slightly. My tired old eyes couldn't take
   it any more.
 * Adjust SQL queries for new statuses.
 * Fix typo in status check code.
 * Have the PLN Plugin cleanup deposits that are complete.
 * Add some missing locale keys
 * Ask the user for confirmation before resetting a deposit.
 * Remove the deposit item count in the status table.
 * Fix colspans in the status table.
 * fix whitespace


plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -97,6 +97,95 @@ function setStatus($status) {
 	}
 
 	/**
+	 * Return a string representation of the local status.
+	 * 
+	 * @param $translatable boolean
+	 * @return string
+	 */
+	function getLocalStatus($translatable = true) {
```
asmecher: Is `$translatable=false` ever used? I don't see it in this PR.

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -97,6 +97,95 @@ function setStatus($status) {
...
+		return $status;
+	}
+	
+	/**
+	 * Return a string representation of the processing status.
+	 * 
+	 * @param $translatable boolean
+	 * @return string
+	 */
+	function getProcessingStatus($translatable = true) {
```
asmecher: Likewise -- `$translatable` unused?

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -97,6 +97,95 @@ function setStatus($status) {
...
+		return $status;
+	}
+	
+	/**
+	 * Return a string representation of the LOCKSS status.
+	 * 
+	 * @param $translatable boolean
+	 * @return string
+	 */
+	function getLockssStatus($translatable = true) {
```
asmecher: Likewise -- `$translatable` unused?

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -97,6 +97,95 @@ function setStatus($status) {
...
+			$status = 'received';
+		}
+		if($this->getValidatedStatus()) {
+			$status = 'validated';
+		}
+		if($this->getSentStatus()) {
+			$status = 'sent';
+		}
+		if($translatable) {
+			return PLN_PLUGIN_LOCALE_PREFIX . $status;
```
asmecher: I'm not a big fan of building locale keys programmatically, because it makes it hard to find where a key is used in the code (e.g. by grepping). Suggest rewriting these functions into a switch, e.g.:

````
switch (true) {
    case $this->getSentStatus(): return 'plugins.generic.pln.status.sent';
    case $this->getValidatedStatus(): return 'plugins.generic.pln.status.validated';
    ...etc...
}
return 'plugins.generic.pln.status.unknown';
````
Nice and compact, but doesn't render locale strings un-greppable.

File: plugins/generic/pln/templates/status.tpl [hunk 5]
```diff
@@ -13,50 +13,43 @@
...
-				<th>{translate key="plugins.generic.pln.status.transferred"}</th>
-				<th>{translate key="plugins.generic.pln.status.received"}</th>
-				<th>{translate key="plugins.generic.pln.status.syncing"}</th>
-				<th>{translate key="plugins.generic.pln.status.synced"}</th>
-				<th>{translate key="plugins.generic.pln.status.updated"}</th>
-				<th>{translate key="plugins.generic.pln.status.local_failure"}</th>
-				<th>{translate key="plugins.generic.pln.status.remote_failure"}</th>
+				<th>{translate key="common.id"}</th>
+				<th>{translate key="common.type"}</th>
+				<th>Checked</th>
```
asmecher: English language hard-coded in template.

plugins/generic/pln/templates/status.tpl
```diff
@@ -13,50 +13,43 @@
...
-				<td>{if $deposit->getUpdateStatus()}{translate key="plugins.generic.pln.status.yes"}{else}{translate key="plugins.generic.pln.status.no"}{/if}</td>
-				<td>{if $deposit->getLocalFailureStatus()}{translate key="plugins.generic.pln.status.yes"}{else}{translate key="plugins.generic.pln.status.no"}{/if}</td>
-				<td>{if $deposit->getRemoteFailureStatus()}{translate key="plugins.generic.pln.status.yes"}{else}{translate key="plugins.generic.pln.status.no"}{/if}</td>
-				<td><input type="submit" name="reset[{$deposit->getId()}]" class="button" value="{translate key="plugins.generic.pln.status.reset"}"/></td>
+				<td>{$deposit->getLastStatusDate()}</td>
+				<td>{translate key=$deposit->getLocalStatus()}</td>
+				<td>{translate key=$deposit->getProcessingStatus()}</td>
+				<td>{translate key=$deposit->getLockssStatus()}</td>
+				<td>{translate key=$deposit->getComplete()}</td>
+				<td><input type="submit" name="reset[{$deposit->getId()}]" class="button" value="{translate key="common.reset"}" onclick="return confirm('{$confirmReset}')" /></td>
```
asmecher: `$confirmReset` needs escaping (e.g. in case an apostrophe appears in the locale key).

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -112,11 +201,35 @@ function setNewStatus() {
...
+	 * @return int
+	 */
+	function _getStatusField($field) {
+		return $this->getStatus() & $field;
+	}
+	
+	/**
+	 * Set a status value.
+	 * 
+	 * @param boolean $value one of the PLN_PLUGIN_DEPOSIT_STATUS_* constants.
```
asmecher: Type annotations should appear after the field name.

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -112,11 +201,35 @@ function setNewStatus() {
...
+	 * @param int $field
+	 */
+	function _setStatusField($value, $field) {
+		if($value) {
```
asmecher: I'm not sure why `$value` is being used this way...

ubermichael:
Thanks for the comments @asmecher. I think I've addressed all your concerns.

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -61,6 +68,8 @@
 define('PLN_PLUGIN_NOTIFICATION_TYPE_ZIP_MISSING',		PLN_PLUGIN_NOTIFICATION_TYPE_PLUGIN_BASE + 0x0000005);
 define('PLN_PLUGIN_NOTIFICATION_TYPE_TAR_MISSING',		PLN_PLUGIN_NOTIFICATION_TYPE_PLUGIN_BASE + 0x0000006);
 
+define('PLN_PLUGIN_LOCALE_PREFIX', 'plugins.generic.pln.status.');
```
asmecher: This is now dead code, I think.

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -47,7 +47,7 @@ function &getDepositObjects() {
 	}
 	
 	/**
-	 * Get deposit uuid
+	 * Get deposit uuid
```
asmecher: Accidental spacing change

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -112,11 +179,35 @@ function setNewStatus() {
...
+	}
+	
+	/**
+	 * Set a status value.
+	 * 
+	 * @param boolean $value 
+	 * @param int $field one of the PLN_PLUGIN_DEPOSIT_STATUS_* constants.
+	 */
+	function _setStatusField($value, $field) {
+		if($value) {
```
asmecher: I still find this use of `$value` strange. What's it doing?

plugins/generic/pln/classes/DepositDAO.inc.php
```diff
@@ -124,7 +124,7 @@ function &getNeedPackaging($journalId) {
 			array(
 				(int) $journalId,
 				(int) PLN_PLUGIN_DEPOSIT_STATUS_PACKAGED,
-				(int) PLN_PLUGIN_DEPOSIT_STATUS_LOCAL_FAILURE
+				(int) PLN_PLUGIN_DEPOSIT_STATUS_LOCKSS_AGREEMENT
```
asmecher: Will this status check change break content that's already in the `pln_deposits` table from before the new version of the plugin?

ubermichael: There is only one journal in the wild that has deposits in the pln_deposits table. We'll ask them to reset the deposits so they get regenerated.

ubermichael:
Ok. @asmecher lets have another round.

plugins/generic/pln/classes/Deposit.inc.php
ubermichael: Status is stored as a bit field. If `$value` is true, it sets the corresponding bit to one, otherwise it sets it to false.

asmecher: Ah, obviously. Thanks.

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -97,6 +97,72 @@ function setStatus($status) {
...
+		if($this->getTransferredStatus()) {
+			return 'plugins.generic.pln.status.transferred';
+		}
+		if($this->getPackagedStatus()) {
+			return 'plugins.generic.pln.status.packaged';
+		}
+		if($this->getNewStatus()) {
+			return 'plugins.generic.pln.status.new';
+		}		
+	}
```
asmecher: else?

plugins/generic/pln/classes/Deposit.inc.php
```diff
@@ -97,6 +97,72 @@ function setStatus($status) {
...
+	 */
+	function getLockssStatus() {
+		if($this->getLockssAgreementStatus()) {
+			$status = 'plugins.generic.pln.status.agreement';
+		}
+		if($this->getLockssSyncingStatus()) {
+			$status = 'plugins.generic.pln.status.syncing';
+		}
+		if($this->getLockssReceivedStatus()) {
+			$status = 'plugins.generic.pln.status.received';
```
asmecher: `$status` is set but never used. Should be `return` instead.

ubermichael: Thanks for catching this. I should have had coffee before trying to fix it.

ubermichael:
Ok. I think I've fixed everything.

PR #715, opened 2016-01-22
Update ru_RU locale

ppv1979:


mtub:
Thanks, @ppv1979. Still would like those locale commits rebased to get rid of the additional merge commits, though ;)

PR #722, opened 2016-01-26
pkp/pkp-lib#1087: Update Crossref schema for export to 4.3.6

ctgraham:
We are primarily interested in:
>   4.3.5, 4.3.6 (PDF) 5/14/2015 added text/rtf as mime type
[Source](http://doi.crossref.org/schemas/common4.3.6.xsd)

There are a number of backwards compatible changes in 4.3.3 -> 4.3.6, and some *not* backwards compatible changes in the updated Common reference of 4.3.2 -> 4.3.6.

The only change we found in the common include that affected our export was depositor/name -> depositor/depositor_name, but others may exist.


ctgraham:
Resolves https://github.com/pkp/pkp-lib/issues/1087 for Pitt ULS test cases.  A second look is needed for any schema elements which changed which we are not using locally but could be referenced in the current code.

PR #727, opened 2016-01-27
Report plugin prerequisites in the ping response.

ubermichael:
Some journals seem to stop contacting the PLN once they've been whitelisted. It's probably a problem of missing prerequisites, but we have no way to tell without asking the journal manager to look for log files, check the prerequisites, or do other things.

This PR will add the prerequisites to the ping response, so we can check directly.

/cc @mjordan


asmecher:
Looks good to me; merged.

PR #729, opened 2016-01-28
fix issue paging for the doi export/registration plugins 

bozana:
S. https://github.com/pkp/pkp-lib/issues/1088


plugins/importexport/crossref/CrossRefExportPlugin.inc.php
```diff
@@ -120,19 +120,22 @@ function displayIssueList(&$templateMgr, &$journal) {
 		$this->registerDaoHook('IssueDAO');
 		$issueIterator =& $issueDao->getPublishedIssues($journal->getId(), Handler::getRangeInfo('issues'));
```
asmecher: When reusing variable names, generally it's a good idea to `unset($issueIterator);` to avoid reference wierdness.

bozana: Thanks @asmecher. I added and pushed it together with the first commit i.e. all again -- for users to be able to apply only one patch...

PR #730, opened 2016-01-29
Fix typo in pkp/pkp-lib#617 fix

selwyntcy:
There is a typo in the fix for pkp/pkp-lib#617


asmecher:
Thanks -- good spotting, @selwyntcy!

PR #731, opened 2016-01-31
Add a configurable URL pointing to documentation for the PLN status page

ubermichael:
The URL defaults to $PLN_URL/docs/status.html where $PLN_URL is the
config setting locks:pln_network, and can be overridden with
locks:pln_status_docs.

Usually, that URL will be http://pkp-pln.lib.sfu.ca/docs/status (which
is 404 at the moment).

Closes pkp/pkp-lib#1045.


asmecher:
Merged -- thanks, @ubermichael.

PR #733, opened 2016-02-01
Check for missing prerequisites and report them in the plugin settings page

ubermichael:
* Refactor checking for acron/scheduled tasks into a plugin method.
 * Check and report acron and the other prerequisites in a unified
   manner.
 * Add a locale key for bagit loading errors.
 * Be more careful about how the bagit library is imported, to prevent
   WSOD when Archive_Tar isn't available. Only import bagit as needed.

Fixes pkp/pkp-lib#1105

I'll add some notes to the commit below.

### DIFF
--- a/plugins/generic/pln/classes/form/PLNSettingsForm.inc.php
+++ b/plugins/generic/pln/classes/form/PLNSettingsForm.inc.php
@@ -68,6 +68,36 @@ class PLNSettingsForm extends Form {
 			$this->setData('terms_of_use_agreement', $terms_agreed);
 		}
 	}
+	
+	/**
+	 * Check for the prerequisites for the plugin, and return a translated 
+	 * message for each missing requirement.
+	 * 
+	 * @return array
+	 */
+	function _checkPrerequisites() {
+		$messages = array();
+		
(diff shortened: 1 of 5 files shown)


plugins/generic/pln/classes/form/PLNSettingsForm.inc.php
```diff
@@ -68,6 +68,33 @@ function readInputData() {
...
+	/**
+	 * Check for the prerequisites for the plugin, and return a translated 
+	 * message for each missing requirement.
+	 * 
+	 * @return array
+	 */
+	function _checkPrerequisites() {
+		$messages = array();
+		
+		if( ! @include_once('Archive/Tar.php')) {
```
asmecher: This check should happen after the PHP5 check, I think.

asmecher: (and perhaps that check shouldn't be run if this one fails)

ubermichael: I think you're right @asmecher - there's no need to do the other checks if php5 isn't available.

PR #738, opened 2016-02-04
Replace template variables

selwyntcy:
Replace all template and locale variables according to the db upgrade script.
Add queries to the db upgrade script for the missing replacement of `weekLaterDate` by `responseDueDate`.

Not sure if it's appropriate to submit this PR right now, as more changes to the email template handling are pending.


asmecher:
Thanks, @selwyntcy, we haven't reviewed the upgrade process yet but this looks like it'll be a helpful preparatory piece.

@mtub, can you look at the translation aspects? This is generally adapting OJS 2.x template variables to the current 3.x templates.

mtub:
OJS 3.0 translations won't be easy and will involve comparing the two branches thoroughly. But I don't see anything  that keeps us from exchanging the email template variable names now. I'm okay with it.

asmecher:
Thanks, @selwyntcy -- this plays nicely with the code shared by both OJS and OMP, and will help translators to focus on the important stuff rather than getting bogged down in code-centric changes like variable renames. That's really helpful.

@mtub, when it comes time to start thinking about OJS 3.0 translations (after OMP 1.2 is out), we'll have to review the translator plugin to see what we can expect translators to work with. Hopefully the OMP 1.2 process will decrease the amount of re-translation. But of course we'll have to watch for OMP-centric language creeping into pkp-lib.

PR #742, opened 2016-02-09
Add a Preserved In block to the PLN plugin.

ubermichael:
Also fix PLNPlugin::getTemplatePath() include the directory separator
at the end of the path string. It was not consistent with other plugins.


asmecher:
@ubermichael, can you create an issue for this on the pkp-lib repository and link this to it? That way we can schedule it against a release so it doesn't get forgotten.

ubermichael:
See pkp/pkp-lib#1173

plugins/generic/pln/PLNPlugin.inc.php
```diff
@@ -163,7 +163,7 @@ function getHandlerPath() {
 	 * @return string
 	 */
 	function getTemplatePath() {
-		return parent::getTemplatePath() . DIRECTORY_SEPARATOR . 'templates';
+		return parent::getTemplatePath() . DIRECTORY_SEPARATOR . 'templates/';
```
asmecher: I'm ambivalent about `DIRECTORY_SEPARATOR` vs. `/`, but suggest keeping consistent on any given line of code :)

PR #750, opened 2016-02-15
Add reply-to headers to scheduled tasks

jnugent:
These two scheduled tasks specifically set the reply-to header to null which led to some rather broken sender envelopes.  Set the header to the primary contact for the journal instead.

### DIFF
--- a/classes/tasks/OpenAccessNotification.inc.php
+++ b/classes/tasks/OpenAccessNotification.inc.php
@@ -42,8 +42,16 @@ class OpenAccessNotification extends ScheduledTask {
 			$email = new MailTemplate('OPEN_ACCESS_NOTIFY', $journal->getPrimaryLocale(), false, $journal, false, true);
 
 			$email->setSubject($email->getSubject($journal->getPrimaryLocale()));
-			$email->setReplyTo(null);
-			$email->addRecipient($journal->getSetting('contactEmail'), $journal->getSetting('contactName'));
+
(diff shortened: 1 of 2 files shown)


asmecher:
Did you find where those empty string reply-to values were getting added? I looked through the code and couldn't spot it.

jmacgreg:
Hey guys, just checking on the status of this. We've patched two hosting journals already, and no complaints. We have a few other journals that would likely be able to make use of this, so it'd be good to include in the stable branch as well. Thanks!

asmecher:
It looks like the patch should work, but I think we need to further investigate the cause to make sure it's not also happening elsewhere.

jnugent:
My current thought is that the setReplyTo method in Mail.inc.php is not doing what it should when it's handed null, like what classes/tasks/ReviewReminder.inc.php does.  Testing it.

jnugent:
Ironically, it appears that this regression may have been introduced into OJS when we added SPF enhancements:

https://github.com/pkp/ojs/commit/2a4f6775b466c09b67346fced7358b5e971b4bd7

We replaced a lot of the setReplyTo() calls that originally set the reply to address to the journal primary contact to null.

jnugent:
I realized this because I am in the middle of porting the SPF enhancements to OCS.

jnugent:
Closing this in favour of fixing the test in pkp-lib

PR #753, opened 2016-02-17
automatic review reminders

bozana:
s. https://github.com/pkp/pkp-lib/issues/562

Could you please check if the upgrade changes are OK so?


asmecher:
Looks good, thanks!

PR #754, opened 2016-02-18
Missing 'else' in booksForReview plugin

tgc:
http://forum.pkp.sfu.ca/t/problems-with-books-for-review/8259/11


asmecher:
Thanks, @tgc. Note that the "books for review" plugin has been retired in favour of the "objects for review" plugin, which is a lot more flexible.

PR #770, opened 2016-03-01
Remove duplicate translation key from objectsForReview

tgc:
While working on the Danish translation we found one key that seemingly could not be translated.
Inspecting the locale.xml file revealed a duplicate translation key as the cause of the problem.
This patch removes the duplicate key at the source and in also corrects the translations that this error has propagated into.


bozana:
Thanks a lot @tgc!
@mtub, would you like to merge or shall I do it?

mtub:
Thanks for spotting and fixing, @tgc.

PR #777, opened 2016-03-03
pkp/pkp-lib#1209 decrease the size of the url field for referrals table again

bozana:
S. https://github.com/pkp/pkp-lib/issues/1209#issuecomment-192004801


asmecher:
Merged, thanks!

PR #785, opened 2016-03-09
Portuguese (Portugal) translation for DOI plugin

josekarvalho:


mtub:
Thank you, @josekarvalho !

PR #787, opened 2016-03-09
Issue/1195

crism:
Addresses pkp/pkp-lib#1195. Issue and article views only include galleys which have been approved, and which have at least one approved file. Individual galleys may still show unapproved files, though.


asmecher:
@NateWr, would you mind taking a look at this? Relates to https://github.com/pkp/pkp-lib/issues/1231 and may need consideration in OMP as well, though this is a PR for OJS.

asmecher:
(And thanks as always, @crism!)

NateWr:
The code for OMP is completely different and [considerably more complicated](https://github.com/pkp/omp/blob/master/templates/frontend/objects/monograph_full.tpl#L160-L233). I'm not really sure why the OJS code seems to be so much more simple, but I suspect it hasn't been very well tested.

For me to do code review on this I'd probably need to spend most of my time next week diving back into OJS and remembering how things work. I'd suggest we table this until OMP 1.2 is out the door and I can come back and take a closer look.

It's a valuable contribution (thanks @crism!), but the whole publication format/galley structure is the most complex part of our software and working things out for OMP took a long time.

crism:
Thanks for the comments, @NateWr. The galley structure did seem strangely oversimplified, and it probably does need a deeper rewrite than I gave it.

NateWr:
@crism, I've looked at this now and I think it's good. OJS has only a single sign-off for the publication formats, so it's a lot cleaner than OMP. The main difference, though, is that we're not linking directly to files. Instead we're linking to a galley page which then links to files. This setup may need some further consideration but not as part of this PR.

There is one more thing that needs to be considered. When you click a galley link, it takes you to a page for that galley that then lists each galley file available. These files are visible even if unapproved.

selection_047

selection_048

The code is in [article.tpl](https://github.com/pkp/ojs/blob/master/templates/frontend/pages/article.tpl#L29). Would you be able to update your PR with code that handles the visibility there as well?

crism:
@NateWr, sure. I’ll poke at that today.

crism:
Updated and tested. Pull, please!

PR #789, opened 2016-03-13
removed (dead) link to Google Doc Viewer

mtub:
* Google Doc Viewer web page/description does not exist anymore
* but service is still up
* see <http://googlesystem.blogspot.de/2015/02/google-docs-viewer-page-no-longer.html>


asmecher:
Thanks, @mtub. Let's leave the plugin present for existing users, but I doubt we'll want to support it for OJS 3.0 and on. The PDF.js plugin has fairly handily solved our PDF embedding woes and Google's service was never all that well supported (and didn't work for subscription journals etc).

PR #795, opened 2016-03-15
Add the article's sponsor(s) node(s) in the XML tree generated by the…

nddf:
Add the article's sponsor nodes when generating XML using the plugins/importexport/native plugin.

### DIFF
--- a/plugins/importexport/native/NativeExportDom.inc.php
+++ b/plugins/importexport/native/NativeExportDom.inc.php
@@ -307,6 +307,17 @@ class NativeExportDom {
 
 		/* --- */
 
+		
+		/* --- Sponsors --- */
+		if(is_array($article->getSponsor(null))) {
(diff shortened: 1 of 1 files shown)


ctgraham:
It looks like the check for this failed because an empty `<sponsor>` element was added to the export in testing:
```
   </permissions>
+  <sponsor locale="en_US"/>
   <galley locale="en_US">
     <label>PDF</label>
     <file>
```
Did you intend to allow for empty elements in the output?

nddf:
I didn't see #PCDATA implies non-empty string. But it is true it is more
meaningful to avoid it...

Sorry for the issue and bad pull-request... 1rst time.

2016-03-15 15:26 GMT+01:00 Clinton Graham <notifications@github.com>:

ctgraham:
No worries.  You can push changes on your existing branch to automatically update this pull request.

PR #798, opened 2016-03-19
Enable Issues images in archives only

nddf:
if $issue->getShowCoverPage($coverLocale) decides whether or not to print the Issue cover page, should it also decide whether to print the image in archives list? It seems !$issue->getHideCoverPageArchives($coverLocale) is responsible for that.


asmecher:
Thanks, @nddf -- I'll cherry-pick this forward to `ojs-dev-2_4` as well.

PR #799, opened 2016-03-21
Add missing words in emailTemplates

samueloph:
Could you please cherrypick this to ojs_dev_2-4 too?

I can make another PR against that branch if its needed.


mtub:
Thanks, @samueloph . Could you file a second  PR against ojs-dev-2_4? Please note that ojs-dev-2_4 has plain text emails, while the master branch has HTML emails.

samueloph:
Thanks @mtub, I noticed the difference, that's why i thought it wouldn't be a trivial cherry pick and suggested doing another PR. I'm doing it right now.

PR #803, opened 2016-03-23
pkp/pkp-lib#1303: fix Reload Default Email Templates: only erase main email templates, reload any missing templates

ctgraham:
Resolves https://github.com/pkp/pkp-lib/issues/1303

Clearing the old default data is already handled by `installEmailTemplateData()`.
Reloading missing templates will be handled by `installEmailTemplates()`


asmecher:
Looks good, thanks!

PR #806, opened 2016-03-29
add check to ensure empty article object is not returned

geocode64:
Prevents fatal error in template file if $returner = null
ie. PHP Fatal error:  Call to a member function getPublished() on a non-object in /xxx/journals/cache/t_compile/%%38^38D^38D7420B%%article.tpl.php on line 154


asmecher:
(Superseded by https://github.com/pkp/ojs/pull/807)

PR #807, opened 2016-03-29
Add check to ensure empty article object is not returned

geocode64:
Apologies for the other pull request to master.

Was recieving this PHP error when viewing a particular journal article:
PHP Fatal error: Call to a member function getPublished() on a non-object in  xxx/journals/cache/t_compile/%%38^38D^38D7420B%%article.tpl.php on line 154

The $returner variable was being set to null due to a true result for $journalId != $returner->getJournalId(); 

The modified code means that the check will fail gracefully and allow the article to load properly if $returner is null.

### DIFF
--- a/classes/article/PublishedArticleDAO.inc.php
+++ b/classes/article/PublishedArticleDAO.inc.php
@@ -357,8 +357,11 @@ class PublishedArticleDAO extends DAO {
 		if ($useCache) {
 			$cache =& $this->_getPublishedArticleCache();
 			$returner = $cache->get($articleId);
-			if ($returner && $journalId != null && $journalId != $returner->getJournalId()) $returner = null;
-			return $returner;
+			$skip = $returner && $journalId != null && $journalId != $returner->getJournalId();
+			
+                        if ($returner && !$skip) {
+                            return $returner;
+                        }
 		}
 
 		$primaryLocale = AppLocale::getPrimaryLocale();


asmecher:
@geocode64, that check (`$journalId != $returner->getJournalId()`) is intended to ensure that the article is within the journal being requested. It looks as though `$journalId` is set, but the `$articleId` corresponds to another journal. Do you know why that might be happening?

geocode64:
@asmecher , do you know what is the reason for this check? Are there expected situations where `$journalId != $returner->getJournalId()`? This may give some clues as to why the `$articleId` is wrong.

asmecher:
@geocode64, we use a pattern (not fully consistently) in OJS to ensure that data integrity is being checked in a request.

Consider a multi-journal setup where a user is an editor of one journal and a reader of another. If both are subscription journals, then subscription checks in the first journal will be bypassed because the user is an editor and is presumed to have access to all content.

However, if that user modifies the URL to the article so that it refers to the first journal (by journal path) and an article in the second journal (by numerical article ID), we'll need to ensure that we don't permit them access.

For cases like this we frequently supply the ID not just of the element we want to fetch (e.g. article), but also its "container" (e.g. journal). That way the integrity check can be done in the database. Because there's caching here as well, we also ensure that anything that's already been fetched from the database and is stored in memory gets an integrity check too.

asmecher:
Closing for lack of activity.

PR #809, opened 2016-04-01
pkp/pkp-lib#1517 include() Archive/Tar before checking if Archive/Tar is available.

ubermichael:
It is never available if it isn't included or required. Strangely, some journals report it as available even if it isn't included. Perhaps it's a PHP caching issue, or another plugin that requires it somewhere.

### DIFF
--- a/plugins/generic/pln/PLNGatewayPlugin.inc.php
+++ b/plugins/generic/pln/PLNGatewayPlugin.inc.php
@@ -21,6 +21,9 @@ import('lib.pkp.classes.core.ArrayItemIterator');
 
 define('PLN_PLUGIN_PING_ARTICLE_COUNT', 12);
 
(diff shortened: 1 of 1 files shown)


plugins/generic/pln/PLNGatewayPlugin.inc.php
```diff
@@ -21,6 +21,8 @@
 
 define('PLN_PLUGIN_PING_ARTICLE_COUNT', 12);
 
+include_once('Archive/Tar.php');
```
asmecher: For consistency, shouldn't we be suppressing errors in the `include_once`?

ubermichael:
Error message suppressed now.

ubermichael:
@asmecher can this one go into 2.4.8_1?

PR #816, opened 2016-04-11
pkp/pkp-lib#1263 port OMP native export changes to OJS

bozana:
S. d) and g) here: https://github.com/pkp/pkp-lib/issues/1263


bozana:
Ah, I'll check the error...

asmecher:
Merged, thanks! The OJS master tests are failing, and it's not caused by this change; I'll look at that once the OMP 1.2 release is out.

PR #818, opened 2016-04-12
pkp/pkp-lib#1347 adapt changes to ojs user export and import

bozana:
S. a), b) and c) here: https://github.com/pkp/pkp-lib/issues/1347


bozana:
What to do with the locale keys that are not used? Shall I remove them, every time I see them, or leave them for now and do a general clean once in the future?

asmecher:
Hmm, best to clean them up when you notice them. We'll need to to a thorough review as well.

PR #821, opened 2016-04-12
pkp/pkp-lib#1353 Navbar user home should stay in journal context for …

mfelczak:
…single journal installs


asmecher:
Thanks, @mfelczak!

PR #826, opened 2016-04-21
pkp/pkp-lib#1401: Add section parameter to getVar() call.

ctgraham:
Resolves https://github.com/pkp/pkp-lib/issues/1401


asmecher:
@ubermichael, FYI.

PR #832, opened 2016-04-28
pkp/pkp-lib#1433 Revert SPF changes and improve SPF compliance

jnugent:
This pull reverts the SPF changes and combined with this commit already in lib-pkp/ojs-dev-2_4 should fix a lot of the mail discrepancies:

https://github.com/pkp/pkp-lib/commit/7af3bf24a2e24f8e2f4afcde4750d7b16d9b80f7

Note that it also adjusts the From: header on Step 5 of the Submission form, in order to prevent the email to the author from being from the author.  The SPF change originally set the reply-to to null which got around that. 

I may add to this PR as I find other places where the From: header may still be incorrect.

### DIFF
--- a/plugins/generic/thesis/StudentThesisForm.inc.php
+++ b/plugins/generic/thesis/StudentThesisForm.inc.php
@@ -286,10 +286,10 @@ class StudentThesisForm extends Form {
 
 			import('classes.mail.MailTemplate');
 			$mail = new MailTemplate('THESIS_ABSTRACT_CONFIRM');
-			$mail->setReplyTo($thesisEmail, $thesisName);
+			$mail->setFrom($thesisEmail, "\"" . $thesisName . "\"");
 			$mail->assignParams($paramArray);
-			$mail->addRecipient($thesis->getSupervisorEmail(), $supervisorName);
-			$mail->addCc($thesis->getStudentEmail(), $studentName);
+			$mail->addRecipient($thesis->getSupervisorEmail(), "\"" . $supervisorName . "\"");
+			$mail->addCc($thesis->getStudentEmail(), "\"" . $studentName . "\"");
 			$mail->send();
 		}
 
--- a/plugins/generic/thesis/StudentThesisForm.inc.php
+++ b/plugins/generic/thesis/StudentThesisForm.inc.php
@@ -286,7 +286,7 @@ class StudentThesisForm extends Form {
 
 			import('classes.mail.MailTemplate');
 			$mail = new MailTemplate('THESIS_ABSTRACT_CONFIRM');
-			$mail->setFrom($thesisEmail, "\"" . $thesisName . "\"");
+			$mail->setFrom($thesisEmail, $thesisName);
 			$mail->assignParams($paramArray);
 			$mail->addRecipient($thesis->getSupervisorEmail(), "\"" . $supervisorName . "\"");
 			$mail->addCc($thesis->getStudentEmail(), "\"" . $studentName . "\"");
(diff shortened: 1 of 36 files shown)


jnugent:
@ctgraham in case you wanted to have a look at this for testing.

ctgraham:
The original PRs littered the code with changes of "From" to "ReplyTo", which I don't think is undone here.  Shouldn't these be reverted as well?
https://github.com/pkp/ojs/pull/337/files
https://github.com/pkp/pkp-lib/pull/180/files

jnugent:
It probably depends on the email which is why I am going through them.  My change to MailTemplate.inc.php forces a setFrom() to the journal or site contact if the user object is not present.  In those cases, leaving the original PR change where the setReplyTo might be fine because the setReplyTo will not override it.

jnugent:
I've pushed up a commit that rolls back everything from 2.4.6, and created a new branch in pkp-lib that does the same there, but maintains my return-path modification.  This should be what we discussed on the call. The other commit is here:

https://github.com/jnugent/pkp-lib/commit/32c16ec302d831481968c690e12590e19383bc6f

asmecher:
@jnugent, could you create an issue to go with this and link between it and the PRs?

ctgraham:
@jnugent, I'm currently testing these changes.  What tools did you use to validate SPF for your testing?

jnugent:
I'm partial to http://www.kitterman.com/spf/validate.html

jnugent:
Hey @ctgraham just bumping this.  We think it's probably tested as thoroughly as it's going to be and we would  like to merge.

Adding @jmacgreg so he can follow along.

plugins/generic/thesis/StudentThesisForm.inc.php
```diff
@@ -286,10 +286,10 @@ function execute() {
 
 			import('classes.mail.MailTemplate');
 			$mail = new MailTemplate('THESIS_ABSTRACT_CONFIRM');
-			$mail->setReplyTo($thesisEmail, $thesisName);
+			$mail->setFrom($thesisEmail, "\"" . $thesisName . "\"");
```
asmecher: Why are these quotes added?

asmecher:
Meanwhile, @jnugent, could you forward-port this to the master branch of OJS and OMP?

asmecher:
(And mind creating an issue for this? I need to make sure it's assigned to a release. Thanks!)

ctgraham:
This has been running on our Scholarly Exchange service for the past two weeks with no new complaints.  The issue is open as https://github.com/pkp/pkp-lib/issues/1433.

asmecher:
Great, looks like we're ready to go on this!

@jnugent, I'll merge, then could you...

- [ ] Cherry-pick to `ojs-stable-2_4_8` (for release with OJS 2.4.8-1)
- [ ] Forward-port to OJS and OMP master

jnugent:
@asmecher 

Two pull requests, against PKP-lib and OMP master:
https://github.com/pkp/omp/pull/299
https://github.com/pkp/pkp-lib/pull/1563

asmecher:
@jnugent, to keep the `master` branch parallel to `ojs-dev-2_4` WRT the SPF changes, I think you'd also need to remove the email header field in setup etc., no?

jnugent:
@asmecher If you're thinking of the emailHeader text field on the workflow template tab, the master branches will still have emails that originate from the primary contact (submission acks, review reminders, etc).  I think I have an idea on how to only include that header on mail messages sent from a non-user context.  Let me code a bit and update the PR.

PR #843, opened 2016-05-20
pkp/pkp-lib#1086 New additions for article metadata versioning

felixgruender:


felixgruender:
See https://github.com/pkp/pkp-lib/pull/1471

classes/article/ArticleDAO.inc.php
```diff
@@ -180,23 +180,26 @@ function updateObject($article) {
...
 				(int) $article->getHideAuthor(),
+				(int) $article->getHideSubmissionRevisions(),
 				(int) $article->getId()
 			)
 		);
 
 		$this->updateLocaleFields($article);
-
+		$contextId = Request::getContext()->getId();
+		$version = $this->getLatestRevisionId($article->getId(), $contextId);
```
bozana: Would it be possible to get the version/revisionId from article object (that is being updated) itself?

classes/article/ArticleGalley.inc.php
```diff
@@ -202,6 +202,58 @@ function getFirstGalleyFile($fileType = null, $allFiles = false) {
...
+	 * @return boolean
+	 */
+	function hasFilesPerSubmissionRevision($submissionRevision) {
+		$submissionFileDao = DAORegistry::getDAO('SubmissionFileDAO');
+		$submissionFiles = $submissionFileDao->getAllRevisionsByAssocId(ASSOC_TYPE_GALLEY, $this->getId());
+		
+		$hasFiles = false;
+		foreach($submissionFiles as $submissionFile) {
+			if ($submissionFile->getSubmissionSettingsRevision() == $submissionRevision) {
+				$hasFiles = true;
```
bozana: Maybe to immediately return true here?

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -242,11 +242,11 @@ function _fromRow($row) {
...
 		$galley->setGalleyType($row['galley_type']);
-
+		
 		$this->getDataObjectSettings('submission_galley_settings', 'galley_id', $row['galley_id'], $galley);
-
+		
 		HookRegistry::call('ArticleGalleyDAO::_fromRow', array(&$galley, &$row));
-
+		
 		return $galley;
```
bozana: Empty tabs inserted by the editor?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -537,17 +538,18 @@ function newDataObject() {
...
+		$publishedArticle = parent::_fromRow($row, $submissionRevision);
 		$publishedArticle->setPublishedArticleId($row['published_submission_id']);
 		$publishedArticle->setIssueId($row['issue_id']);
 		$publishedArticle->setSequence($row['seq']);
 		$publishedArticle->setAccessStatus($row['access_status']);
-
+		
 		$publishedArticle->setGalleys($this->galleyDao->getBySubmissionId($row['submission_id'])->toArray());
-
+
```
bozana: Empty tabs again :-)

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
```diff
@@ -300,7 +303,6 @@ function execute($request) {
 					$publishedArticle = $publishedArticleDao->newDataObject();
 					$publishedArticle->setId($submission->getId());
 					$publishedArticle->setIssueId($issueId);
-					$publishedArticle->setDatePublished(Core::getCurrentDate());
```
bozana: What about article that were not scheduled to an issue but are scheduled right now -- their date_published would be empty which is not good I believe -- the current date i.e. the date when an issue is assigned should be there per default, I believe. I can double check, how is it in the current master branch, but I am pretty sure it was like this...

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
```diff
@@ -342,6 +344,20 @@ function execute($request) {
 				$submission->setCopyrightHolder(null, null);
 				$submission->setLicenseURL(null);
 			}
+			
+			$submission->setDatePublished($this->getData('datePublished'));
+			
+			$submissionDao = Application::getSubmissionDAO();
+			if ($request->getUserVar('submissionRevision')) {
+				$revision = (int)$request->getUserVar('submissionRevision');
```
bozana: I am not sure if it would be OK/better to consider those new checkboxes as part of the form, like all the other form fields?

bozana: Ah, sorry, I actually cannot find this field at the glance...

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
```diff
@@ -342,6 +344,20 @@ function execute($request) {
...
+			} else {
+				$revision = $submissionDao->getLatestRevisionId($submission->getId(), $context->getId());
```
bozana: Also here, as for article update: would it maybe be better to have the information within the object itself? -- I cannot overlook all the situations for versioning... -- Maybe it is not better and not efficient... but just asking... :-)

pages/article/ArticleHandler.inc.php
```diff
@@ -49,20 +78,82 @@ function authorize($request, &$args, $roleAssignments) {
...
 		if ($this->journal->getSetting('enablePublicArticleId')) {
-			$publishedArticle = $publishedArticleDao->getPublishedArticleByBestArticleId((int) $this->journal->getId(), $articleId, true);
+			$publishedArticle = $publishedArticleDao->getPublishedArticleByBestArticleId((int) $this->journal->getId(), $this->articleId, true);
 		} else {
-			$publishedArticle = $publishedArticleDao->getPublishedArticleByArticleId((int) $articleId, (int) $this->journal->getId(), true);
+			$publishedArticle = $publishedArticleDao->getPublishedArticleByArticleId((int) $this->articleId, (int) $this->journal->getId(), true);
+		}
+		
+		if ($this->submissionRevision) {
+			$publishedArticle = $publishedArticleDao->getPublishedArticleByArticleId((int) $publishedArticle->getId(), (int) $this->journal->getId(), false, $this->submissionRevision);
```
bozana: If I enter a version that doesn't exist (...article/view/article/1/10) I get a strange web page...

pages/article/ArticleHandler.inc.php
```diff
@@ -182,9 +270,10 @@ function view($args, $request) {
 			$templateMgr->assign('galley', $galley);
 			$templateMgr->assign('section', $section);
 			$templateMgr->assign('journal', $journal);
-			$templateMgr->assign('fileId', $fileId);
 			$templateMgr->assign('defineTermsContextId', isset($defineTermsContextId)?$defineTermsContextId:null);
-
+
```
bozana: What about keywords in the line 265 -- will there be the right version of keywords?

pages/article/ArticleHandler.inc.php
```diff
@@ -195,7 +284,7 @@ function view($args, $request) {
 				$templateMgr->assign('licenseUrl', $article->getLicenseURL());
 				$templateMgr->assign('ccLicenseBadge', Application::getCCLicenseBadge($article->getLicenseURL()));
 			}
-
+
```
bozana: empty tabs...

pages/article/ArticleHandler.inc.php
```diff
@@ -17,18 +17,47 @@
...
-	/** galley associated with the request **/
+	/** @var galley object: galley associated with the request **/
 	var $galley;
-
+	
+	/** @var int: type of the requested object (article, galley or file) **/
+	var $type;
+	
+	/** @var int: article ID associated with the request **/
+	var $articleId;
```
bozana: Do we need both, article (above) and articleId? Also galley (above) and galleyId (below)?

plugins/oaiMetadataFormats/nlm/OAIMetadataFormat_NLM.inc.php
```diff
@@ -98,7 +98,7 @@ function toXml(&$record, $format = null) {
 			if ($locale == $primaryLocale) continue;
 			$response .= "\t\t\t\t<trans-title xml:lang=\"" . strtoupper(substr($locale, 0, 2)) . "\">" . htmlspecialchars(Core::cleanVar(strip_tags($title))) . "</trans-title>\n";
 		}
-
+
```
bozana: empty tabs...

templates/controllers/modals/submissionMetadata/issueEntryTabs.tpl
```diff
@@ -23,6 +23,21 @@
 	{rdelim});
 </script>
 {if not $hideHelp}<p class="pkp_help">{translate key="issue.manage.entryDescription"}</p>{/if}
+<form id='submissionSettingsRevisionForm' class='pkp_form'>
+	{if $submissionSettingsRevisions|@count > 1}
+		<div id='editPreviousRevision'>
```
bozana: Actually, you don't need divs around fbvForm elements -- it should be styled per default and thus always consistent.

templates/controllers/modals/submissionMetadata/issueEntryTabs.tpl
```diff
@@ -23,6 +23,21 @@
...
+			{fbvFormSection  for="submissionSettingsRevision"}
+				{fbvElement type="select" name="submissionSettingsRevision" id="submissionSettingsRevision" from=$submissionSettingsRevisions translate=false selected=$currentSubmissionSettingsRevision}
+			{/fbvFormSection}
+		</div>
+	{/if}
+	{if $isPublished}
+		<ul class="checkbox_and_radiobutton">
```
bozana: Here too. You should use "{fbvFormSection list=true}" for checkboxes.

templates/controllers/modals/submissionMetadata/issueEntryTabs.tpl
```diff
@@ -23,6 +23,21 @@
 	{rdelim});
 </script>
 {if not $hideHelp}<p class="pkp_help">{translate key="issue.manage.entryDescription"}</p>{/if}
+<form id='submissionSettingsRevisionForm' class='pkp_form'>
+	{if $submissionSettingsRevisions|@count > 1}
+		<div id='editPreviousRevision'>
+			{fbvFormSection  for="submissionSettingsRevision"}
+				{fbvElement type="select" name="submissionSettingsRevision" id="submissionSettingsRevision" from=$submissionSettingsRevisions translate=false selected=$currentSubmissionSettingsRevision}
```
bozana: I think you don't need name -- it will be automatically taken from the id.

templates/controllers/modals/submissionMetadata/issueEntryTabs.tpl
```diff
@@ -23,6 +23,21 @@
...
+			{/fbvFormSection}
+		</div>
+	{/if}
+	{if $isPublished}
+		<ul class="checkbox_and_radiobutton">
+			{fbvElement type="checkbox" id="saveAsRevision" name="saveAsRevision" label="submission.saveAsRevision"}
```
bozana: Here too: I think you don't need name -- it will be automatically taken from the id.

templates/frontend/pages/article.tpl
```diff
@@ -23,18 +31,51 @@
...
+				</ul>
+			</form>
+		{else}
+			{translate key="submission.noGalleyFiles"}
+		{/if}
 	{else}
 		{* Show article overview *}
 		{include file="frontend/objects/article_details.tpl"}
+		
+		{if $isPreviousRevision}
```
bozana: We will probably have to speak with Nate about the presentation in general, but maybe just a note here that only the link to the last article version is displayed and if on the last article version -- I do not see that there are older ones...

templates/controllers/modals/submissionMetadata/issueEntryTabs.tpl
bozana: Ah, I have just seen, that also else name is used, so you can leave it... Sorry... :-(

templates/frontend/pages/article.tpl
ojsde: 1) It was intentional to display just the link to the newest version when you are on an older revision.
2) You actually should see the links to all older versions when you are on the newest version (unless the "Hide links to previous versions of this submission" checkbox is activated)

felixgruender: (erroneously logged in as 'ojsde' in my last comment ;-))

pages/article/ArticleHandler.inc.php
felixgruender: I thought it was clearer to store article/articleId and galley/galleyId in class variables: so they are instantiated once and then used in several functions.

pages/article/ArticleHandler.inc.php
felixgruender: ...should work now.

templates/frontend/pages/article.tpl
```diff
@@ -23,18 +31,51 @@
...
+		<div class='version_info'>
+			<h2>{translate key="submission.linkToRecentRevision"}</h2>
+			{assign var=newVersionLink value=$article->getBestArticleId($currentJournal)}
+			<a href="{url op="view" path=$newVersionLink escape=false}">{$article->getLocalizedTitle()}</a>
+		</div>
+		{else}
+			{if $previousRevisions|@count > 0 && !$hideSubmissionRevisions}
```
bozana: Hmmm... For some reason, this if clause is not true, although I didn't check the "Hide links..." option :-P I will take a look why... I.e. I don't see the previous revisions... :-P ;-)

bozana: Hmmm... my $previousRevisions|@count is = 0 :-(

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
felixgruender: The main reason why I remove the setDatePublished function here was because 'datePublished' doesn't belong to the submissions table any more (it belongs now to submission_settings).
So I'm not quite sure where to initialize the date_published for unpublished submissions...

templates/frontend/pages/article.tpl
felixgruender: Strange - I can't reproduce this... 
$previousRevisions|@count has the right value in my installation...

templates/frontend/pages/article.tpl
bozana: @felixgruender, this doesn't work for me, because I use a custom public identifier for the article....

pages/article/ArticleHandler.inc.php
bozana: But then it has to be ensured at the appropriate places in the file that both are consistent. For example, article object is given as parameter to the hooks, for plugins to do something with it, so it should be consistent with the actual articleId. Also sometimes $this->article->getId() and sometimes $this->articleId is used.

templates/frontend/pages/article.tpl
bozana: This is maybe an example how important it is to use those variables consistently?

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
bozana: You should differentiate/consider the right case (if the submission has just been re/assigned to an issue) further down, in the line 348, when you set the published date for the submission ($submission->setDatePublished($this->getData('datePublished'));). In that case, instead of assigning it the value of the input text field, use the current date (Core::getCurrentDate()). Does that work?

plugins/reports/articles/ArticleReportDAO.inc.php
```diff
@@ -31,7 +31,8 @@ function getArticleReport($journalId) {
 				COALESCE(asl2.setting_value, aspl2.setting_value) AS abstract,
 				COALESCE(sl.setting_value, spl.setting_value) AS section_title,
 				a.status AS status,
-				a.language AS language
+				a.language AS language,
+				aspl1.submission_revision AS version
```
bozana: hmmm... this seems strange to me -- are you are looking only for the revision/version of that setting_name (title?) in the primary language?
if so, what if there is no such title? -- title doesn't have to exist in the primary locale, I believe...

templates/frontend/pages/article.tpl
bozana: Using a custom public identifier some other things are also not functioning -- for example, I see only the metadata (title, abstract) of the last version, when viewing the article.

bozana:
Dear Felix, in the most recent version (copy from your branch) I now as editor cannot read-only the previous submission version in the metadata modal. Also, I cannot add further authors to the current version :-(

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -242,7 +242,7 @@ function _fromRow($row) {
 		$galley->setRemoteURL($row['remote_url']);
 		$galley->setIsApproved($row['is_approved']);
 		$galley->setGalleyType($row['galley_type']);
-
+
```
bozana: empty tabs...

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -537,15 +538,16 @@ function newDataObject() {
...
-	function _fromRow($row, $callHooks = true) {
-		$publishedArticle = parent::_fromRow($row);
+	function _fromRow($row, $callHooks = true, $submissionRevision = null) {
+		$publishedArticle = parent::_fromRow($row, $submissionRevision);
 		$publishedArticle->setPublishedArticleId($row['published_submission_id']);
 		$publishedArticle->setIssueId($row['issue_id']);
 		$publishedArticle->setSequence($row['seq']);
 		$publishedArticle->setAccessStatus($row['access_status']);
-
+
```
bozana: empty tabs...

pages/article/ArticleHandler.inc.php
lilients: Hi @bozana. I took over from @felixgruender and removed the $articleId and uses $this->article->getId(). Thanks for the tip! The current version of the code can be viewed here: https://github.com/lilients/ojs and here https://github.com/lilients/pkp-lib. It is not yet final though.

templates/frontend/pages/article.tpl
lilients: Hi @bozana. I added the list of all versions to all article pages. But sure, Nate should have the last word.

lilients: Fixed custom ids: https://github.com/lilients/ojs/commit/b5d9dc96596a5bb466a79f22dc365a02c8a07f72

PR #851, opened 2016-05-27
Patch 2

Igor-In:
To make import with citation list


asmecher:
@Igor-In, I think this was probably made against the wrong branch, which is why it appears to contain so many commits. Double-check the branch and try again?

PR #855, opened 2016-05-30
pkp/pkp-lib#1074 Automatic database casts potentially cause havoc wit…

mfelczak:
…h public IDs


asmecher:
Merged, thanks! @bozana, this might need forward-porting along with your pub ID stuff into the master branch.

PR #861, opened 2016-06-12
pkp/pkp-lib#1514 fix re-map of file constants during upgrade

jnugent:
in fixing the original report in the issue:


jnugent:
I discovered a regression -- we removed supplement files at some point, so those are now mapped to attachments in this pull request.

The original bug is a result of the new PKP-lib constants swapping editor files and notes.  The original code converted editor files, but would have merged them into notes when the  second update ran. The update process now moves them both to unused constant values separately and then fixes them at the end.  I think this solves the problem.

asmecher:
Merged, though I haven't tested this. I'll be working on the upgrade process pretty heavily in the coming weeks so should give this a thorough going-over.

PR #868, opened 2016-06-21
Use strip_unsafe_html instead of escape to not break Issue title if s…

Commifreak:
…afe HTML is used

Just like @ /issue/view/ the archive should use `strip_unsafe_html` for filtering the issue title. We have some isue title with `<i>` and `escape` breaks it at the archive.

We use 2.4.8.0 currently.


asmecher:
@Commifreak, that field isn't intended to allow HTML elements; there will be other areas where this change would be needed -- but I'd argue that allowing HTML there will have unintended consequences. If a user legitimately wants to put a `<`, `>`, or `&` in their issue title, they will have to escape it using `&lt;`, `&gt;`, and `&amp;`.  I don't think many of our users would expect that.

In order to both support HTML and avoid confusing users, making that field TinyMCE-capable is an option -- but some investigation will need to be done to permit one-line fields to behave well with TinyMCE. https://pkp.sfu.ca/bugzilla/show_bug.cgi?id=6759

PR #870, opened 2016-06-21
pkp/pkp-lib#1537 consider public URL identifier in issue required policy

bozana:
S. https://github.com/pkp/pkp-lib/issues/1537

### DIFF
--- a/classes/security/authorization/OjsIssueRequiredPolicy.inc.php
+++ b/classes/security/authorization/OjsIssueRequiredPolicy.inc.php
@@ -74,7 +74,7 @@ class OjsIssueRequiredPolicy extends DataObjectRequiredPolicy {
 		$router = $this->_request->getRouter();
 		switch(true) {
 			case is_a($router, 'PKPPageRouter'):
-				if ( ctype_digit($this->_request->getUserVar($this->_parameterName)) ) {
+				if ( ctype_digit((string) $this->_request->getUserVar($this->_parameterName)) ) {
 					// We may expect a object id in the user vars
 					return (int) $this->_request->getUserVar($this->_parameterName);
 				} else if (isset($this->_args[0])) {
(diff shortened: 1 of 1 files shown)


classes/security/authorization/OjsIssueRequiredPolicy.inc.php
```diff
@@ -64,6 +64,31 @@ function dataObjectEffect() {
...
+	/**
+	 * @copydoc DataObjectRequiredPolicy::getDataObjectId()
+	 * Considers a not numeric public URL identifier
+	 */
+	function getDataObjectId() {
+		// Identify the data object id.
+		$router = $this->_request->getRouter();
+		switch(true) {
+			case is_a($router, 'PKPPageRouter'):
+				if ( is_numeric($this->_request->getUserVar($this->_parameterName)) ) {
```
asmecher: Might be better to use `ctype_digit` as in [this example from the stable codebase](https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/classes/issue/IssueDAO.inc.php#L183). `is_numeric` will return true with e.g. `0xa`, `2e3`, and other unexpected notations.

bozana:
@asmecher, I change it, s. the new commit. I will then change it also in the basis class in the pkp-lib and make a PR, ok?

classes/security/authorization/OjsIssueRequiredPolicy.inc.php
```diff
@@ -74,7 +74,7 @@ function getDataObjectId() {
 		$router = $this->_request->getRouter();
 		switch(true) {
 			case is_a($router, 'PKPPageRouter'):
-				if ( is_numeric($this->_request->getUserVar($this->_parameterName)) ) {
+				if ( ctype_digit($this->_request->getUserVar($this->_parameterName)) ) {
```
asmecher: Add a string cast -- `ctype_digit(123)` returns false, but `ctype_digit("123")` returns true.

bozana:
Like this?

asmecher:
Yes, that should do it!

bozana:
OK, I will then also make a PR for the basis class in pkp-lib... THANKS!!!

PR #871, opened 2016-06-22
Datacite DOI automatic registration

bozana:
S. https://github.com/pkp/pkp-lib/issues/1543

### DIFF
--- a/plugins/importexport/datacite/DataciteInfoSender.inc.php
+++ b/plugins/importexport/datacite/DataciteInfoSender.inc.php
@@ -14,7 +14,6 @@
  */
 
 import('lib.pkp.classes.scheduledTask.ScheduledTask');
-import('lib.pkp.classes.core.JSONManager');
 
 
 class DataciteInfoSender extends ScheduledTask {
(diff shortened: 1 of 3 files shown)


plugins/importexport/datacite/DataciteInfoSender.inc.php
```diff
@@ -137,8 +141,10 @@ function _getJournals() {
 			if ($doiPrefix) {
 				$journals[] =& $journal;
 			} else {
-				$this->notify(SCHEDULED_TASK_MESSAGE_TYPE_WARNING,
-					__('plugins.importexport.crossref.senderTask.warning.noDOIprefix', array('path' => $journal->getPath())));
+				$this->addExecutionLogEntry(
```
ctgraham: Shouldn't this depend on whether the user is executing this from the web interface, or whether this is called via a cron/acron job?

bozana: Is this not always executed via cron/acron job?

ctgraham: You're probably right.  I only looked at the function name and thought it might be shared by the UI.  The parent class makes that unlikely.

PR #873, opened 2016-06-22
Clean up temporary files after processing in the PLN plugin.

ubermichael:
tempnam() creates a temp file but doesn't automatically remove it, unlike tmpfile(). 

Fixes pkp/pkp-lib#1545


asmecher:
Merged and cherry-picked forward, thanks!

PR #876, opened 2016-06-23
fixed minor html error in locale.xml (all locales)

mtub:
* there's a html mismatch `<h4>...</div>` in locale.xml


asmecher:
Merged, thanks!

PR #884, opened 2016-06-25
pkp/pkp-lib#1204 no mime_type for remote galleys in crossref export

bozana:
S. https://github.com/pkp/pkp-lib/issues/1204

This was missing for the stable branch :-(


asmecher:
Thanks for catching this!

PR #865, opened 2016-06-18
Medra DOI automatic registration

marchitelli:
This is a tentative to port Automatic DOI registration to DataCite #594 to mEDRA registration plugin.

I have no more errors in my log but the articles are still unregistered. I tested the manual registration and it was ok, so the journal/doi/plugin configuration is ok.

On the journal the acron plugin is enabled and scheduled_task on config.inc.php are set to on
In the database the new setting is registered
"medraexportplugin";"''";30;"automaticRegistration";"1";"bool"

@bozana may you help me?

Thanks,
andrea

PR #888, opened 2016-06-27
Medra DOI automatic registration

marchitelli:
Automatic DOI registration to mEDRA based on DataCite plugin; supersedes the PR #865.


bozana:
@marchitelli, could you remove those files that should be a symlink to the files in the datacite folder?

bozana:
@marchitelli, I believe you have to remove changes to those files -- for them to be as in PKP github i.e. maybe to first reset everything back (e.g. with git reset HEAD~2) and then reset the changes to those files with git checkout -- file (e.g. git checkout -- plugins/importexport/medra/classes/form/DOIExportSettingsForm.inc.php). Then double check that just the necessary files are changed (with git status) and push again (eventually with push -f). Sorry :-(

marchitelli:
I can not revert the commit (I am a dummy git user!). I am restarting again :)

PR #889, opened 2016-06-27
pkp/pkp-lib#1564 Move status update outside of email send test

jnugent:
This pull request moves that status update outside of the $send test, so it happens every time.  Issue here:

https://github.com/pkp/pkp-lib/issues/1564


jnugent:
@jmacgreg this should fix ticket #2016060910000041

PR #896, opened 2016-06-28
Implement/test Crossref support for OJS 3.0

bozana:
Tanananaaaaaaaaaaaa: The Crossref export plugin for OJS 3.0 :-)


classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -0,0 +1,411 @@
...
+					$notificationManager->createTrivialNotification($user->getId(), NOTIFICATION_TYPE_SUCCESS);
+					return new JSONMessage();
+				} else {
+					return new JSONMessage(true, $form->fetch($request));
+				}
+			default:
+				$form->initData();
+				return new JSONMessage(true, $form->fetch($request));
+		}
+		return parent::manage($args, $request);
```
asmecher: I think this is dead code while there's a `default:` clause in place -- but generally it's good to call the parent method if possible. Maybe use something specific instead of `default:`?

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -0,0 +1,411 @@
...
+			DOI_EXPORT_ACTION_DEPOSIT => __('plugins.importexport.common.register'),
+			DOI_EXPORT_ACTION_CHECKSTATUS => __('plugins.importexport.common.checkStatus'),
+		);
+	}
+
+	/**
+	 * Get the plugin ID used as plugin settings prefix.
+	 * @return string
+	 */
+	abstract function getPluginId();
```
asmecher: `Id` suggests a numeric ID to me; maybe better to call it something else? e.g. `getPluginPrefix` or something?

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	/**
+	 * Saving object's DOI to the object's
+	 * "registeredDoi" setting.
+	 * We prefix the setting with the plugin's
+	 * id so that we do not get name clashes
+	 * when several DOI registration plug-ins
+	 * are active at the same time.
+	 * @parem $request Request
```
asmecher: `@parem` should be `@param` (twice)

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -0,0 +1,411 @@
...
+			null
+		);
+		// Cache issues.
+		$nullVar = null;
+		$cache = $this->getCache();
+		foreach ($issues as $issue) {
+			$cache->add($issue, $nullVar);
+			unset($issue);
+		}
+		return $issues->toArray();
```
asmecher: It looks like this code expects to iterate through the issue list twice, and I'm not sure that's possible (e.g. with a database result set). Better to convert to an array and then use the array in the foreach loop too.

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -0,0 +1,411 @@
...
+		$issueDao = DAORegistry::getDAO('IssueDAO'); /* @var $issueDao IssueDAO */
+		$issues = $issueDao->getByPubIdType(
+			$this->getPubIdType(),
+			$context?$context->getId():null,
+			$this->getPluginId(). '::' . DOI_EXPORT_REGISTERED_DOI,
+			null,
+			null
+		);
+		// Cache issues.
+		$nullVar = null;
```
asmecher: Now that we're not using references here, it's possible to pass `null` directly into `add` and remove the use of `$nullVar`.

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -0,0 +1,411 @@
...
+
+	/**
+	 * @copydoc PKPImportExportPlugin::usage
+	 */
+	function usage($scriptName) {
+		fatalError('Not implemented');
+	}
+
+	/**
+	 * @see PKPImportExportPlugin::executeCLI()
```
asmecher: If you use `@see`, then Doxygen will expect you to describe the parameters; if you use `@copydoc`, then the descriptions will be inherited.

controllers/grid/pubIds/PubIdExportIssuesListGridCellProvider.inc.php
```diff
@@ -0,0 +1,129 @@
...
+						'edit',
+						new AjaxModal(
+							$dispatcher->url($request, ROUTE_COMPONENT, null, 'grid.issues.BackIssueGridHandler', 'editIssue', null, array('issueId' => $publishedIssue->getId())),
+							__('plugins.importexport.common.settings.DOIPluginSettings')
+						),
+						$publishedIssue->getIssueIdentification(),
+						null
+					)
+				);
+				break;
```
asmecher: (`break;` after `return` is dead code)

controllers/grid/pubIds/PubIdExportIssuesListGridCellProvider.inc.php
```diff
@@ -0,0 +1,129 @@
...
+	 */
+	function getTemplateVarsFromRowColumn($row, $column) {
+		$publishedIssue = $row->getData();
+		$columnId = $column->getId();
+		assert(is_a($publishedIssue, 'Issue') && !empty($columnId));
+
+		switch ($columnId) {
+			case 'identification':
+				return array('label' => '');
+				break;
```
asmecher: (`break` after `return`; also below)

controllers/grid/pubIds/PubIdExportIssuesListGridHandler.inc.php
```diff
@@ -0,0 +1,228 @@
...
+			LOCALE_COMPONENT_APP_EDITOR
+		);
+
+		// Fetch the authorized roles and determine if the user is a manager.
+		$authorizedRoles = $this->getAuthorizedContextObject(ASSOC_TYPE_USER_ROLES);
+		$this->_isManager = in_array(ROLE_ID_MANAGER, $authorizedRoles);
+
+		$pluginCategory = $request->getUserVar('category');
+		$pluginPathName = $request->getUserVar('plugin');
+		$this->_plugin = PluginRegistry::loadPlugin($pluginCategory, $pluginPathName);
```
asmecher: Hmm, I'm nervous about these being passed in from the request without validation.

controllers/grid/pubIds/PubIdExportSubmissionsListGridHandler.inc.php
```diff
@@ -0,0 +1,299 @@
...
+
+	//
+	// Implement template methods from PKPHandler
+	//
+	/**
+	 * @copydoc PKPHandler::authorize()
+	 */
+	function authorize($request, &$args, $roleAssignments) {
+		import('lib.pkp.classes.security.authorization.ContextAccessPolicy');
+		$this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
```
asmecher: Isn't that an overly vague policy (e.g. permitting anyone with context access to view this)?

controllers/grid/pubIds/PubIdExportSubmissionsListGridHandler.inc.php
```diff
@@ -0,0 +1,299 @@
...
+		$authorizedRoles = $this->getAuthorizedContextObject(ASSOC_TYPE_USER_ROLES);
+		$this->_isManager = in_array(ROLE_ID_MANAGER, $authorizedRoles);
+
+		$pluginCategory = $request->getUserVar('category');
+		$pluginPathName = $request->getUserVar('plugin');
+		$this->_plugin = PluginRegistry::loadPlugin($pluginCategory, $pluginPathName);
+
+		// Grid columns.
+		$cellProvider = new PubIdExportSubmissionsListGridCellProvider($this->_plugin, $authorizedRoles);
+		$this->addColumn(
```
asmecher: We decided to remove issue IDs from the grid, IIRC.

lib/pkp
```diff
@@ -1 +1 @@
-Subproject commit d0e0deb561b459051083b4f0eac1d2562398d79d
+Subproject commit d071c7929e4dd54146b7953bae7666c93281cb04
```
asmecher: Is there a pkp-lib PR as well?

locale/en_US/locale.xml
```diff
@@ -1149,4 +1149,35 @@ Validation may fail on search URLs that require posted form data.]]></message>
 	<message key="search.cli.rebuildIndex.indexingByJournalNotSupported">This search implementation does not allow per-journ
 al re-indexing.</message>
 	<message key="search.cli.rebuildIndex.unknownJournal">The given journal path "{$journalPath}" could not be resolved to a journal.</message>
+
+	<!-- Common DOI export plugins (crossref, datacite, medra) keys -->
```
asmecher: Could these be put in `locale/en_US/manager.xml` instead?

plugins/importexport/crossref/CrossRefExportPlugin.inc.php
```diff
@@ -13,313 +13,513 @@
...
+	 * Hook callback that returns the deposit setting's names,
+	 * to consider them by article or issue update.
+	 *
+	 * @copydoc DOIPubIdExportPlugin::getAdditionalFieldNames()
+	 */
+	function getAdditionalFieldNames($hookName, $args) {
+		parent::getAdditionalFieldNames($hookName, $args);
+		assert(count($args) == 2);
+		$dao =& $args[0];
+		$returner =& $args[1];
```
asmecher: Maybe rename this, since it's not being returned :)

plugins/importexport/crossref/README
```diff
@@ -1,14 +1,16 @@
 ================================
 === OJS CrossRef Export Plugin
-=== Version: 1.1
-=== Release date: June 8, 2006
-=== Author: Juan Pablo Alperin <juan@geeky.net>
+=== Version: 2.1
+=== Author: Bozana Bokan <bozana.bokan@posteo.net>
+=== Author: Juan Pablo Alperin <juan@alperin.ca>
+=== Author: James MacGregor <jmacgreg@gmail.com>
+=== Last update: July 2016
```
asmecher: Better not to include "last update", since that info is available via github and we're likely not to update this consistently.

plugins/importexport/crossref/README
```diff
@@ -18,35 +20,24 @@ complete terms of this license.
...
-guidelines, however, it has not been explicitly tested as of this release.
+ - Tools > Import/Export > CrossRef XML Export Plugin
 
 Contact/Support
 ---------------
-Please email the author for support, bugfixes, or comments.
-Email: <juan@geeky.net>
+For support, bugfixes, or comments please use PKP-LIB GitHub issues: https://github.com/pkp/pkp-lib/issues
 
 Version History
```
asmecher: Likewise, I'd lean towards removing this. It's included in the OJS distro, so no need to version separately, and we're not likely to maintain this history consistently.

controllers/grid/pubIds/PubIdExportSubmissionsListGridHandler.inc.php
bozana: This is for submission listing

lib/pkp
bozana: it is just to update the reference -- you already merged the PR -- the display of xml validation messages (https://github.com/pkp/pkp-lib/pull/1569)

plugins/importexport/crossref/filter/ArticleCrossrefXmlFilter.inc.php
```diff
@@ -0,0 +1,278 @@
...
+	function appendCollectionNodes($doc, $doiDataNode, $submission, $submissionFiles) {
+		$deployment = $this->getDeployment();
+		$context = $deployment->getContext();
+		$request = Application::getRequest();
+
+		// start of the text-mining collection element
+		$textMiningCollectionNode = $doc->createElementNS($deployment->getNamespace(), 'collection');
+		$textMiningCollectionNode->setAttribute('property', 'text-mining');
+		foreach ($submissionFiles as $betGalleyId => $submissionFile) {
+			$resourceURL = $request->url($context->getPath(), 'article', 'download', array($submission->getBestArticleId(), $betGalleyId, $submissionFile->getFileId()));
```
asmecher: OTOH, I think it's better to call getBestArticleId with the context ID specified, but please check!

plugins/importexport/crossref/filter/filterConfig.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE filterConfig SYSTEM "../../../../lib/pkp/dtd/filterConfig.dtd">
+
+<!--
+  * filterConfig.xml
```
asmecher: Best to include the full path.

plugins/importexport/crossref/templates/index.tpl
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * Copyright (c) 2014-2016 Simon Fraser University Library
+ * Copyright (c) 2003-2016 John Willinsky
+ * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+ *
+ * List of operations this plugin can perform
+ *}
+{strip}
+{assign var="pageTitle" value="plugins.importexport.crossref.displayName"}
+{include file="common/header.tpl"}
```
asmecher: FYI, this whole stanza can be made a little cleaner/briefer:

    {include file="common/header.tpl" pageTitle="plugins.importexport.crossref.displayName"}

plugins/importexport/crossref/templates/index.tpl
```diff
@@ -0,0 +1,113 @@
...
+	{assign var="allowExport" value=false}
+{else}
+	{assign var="allowExport" value=true}
+{/if}
+
+<script type="text/javascript">
+	// Attach the JS file tab handler.
+	$(function() {ldelim}
+		$('#importExportTabs').pkpHandler('$.pkp.controllers.TabHandler');
+		//$('#importExportTabs').tabs('option', 'cache', true);
```
asmecher: (Commented code)

plugins/importexport/crossref/templates/index.tpl
```diff
@@ -0,0 +1,113 @@
...
+					{include file="controllers/notification/inPlaceNotificationContent.tpl" notificationId=crossrefConfigurationErrors notificationStyleClass=notifyWarning notificationTitle="plugins.importexport.common.missingRequirements"|translate notificationContents="plugins.importexport.crossref.error.noDOIContentObjects"|translate}
+				{/if}
+			</div>
+		{/if}
+
+		{url|assign:crossrefSettingsGridUrl router=$smarty.const.ROUTE_COMPONENT component="grid.settings.plugins.settingsPluginGridHandler" op="manage" plugin="CrossRefExportPlugin" category="importexport" escape=false}
+		{load_url_in_div id="crossrefSettingsGridContainer" url=$crossrefSettingsGridUrl}
+	</div>
+
+	{if $allowExport}
```
asmecher: Reduce indentation by one tab (since this is getting pretty deeply nested):

    {if $allowExport && exportArticles}

asmecher: Oops, never mind, you can't.

templates/controllers/grid/pubIds/pubIdExportSubmissionsGridFilter.tpl
```diff
@@ -0,0 +1,32 @@
+{**
+ * templates/controllers/grid/pubIds/pubIdExportSubmissionsGridFilter.tpl
+ *
+ * Copyright (c) 2014-2016 Simon Fraser University Library
+ * Copyright (c) 2000-2016 John Willinsky
+ * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+ *
+ * Filter template for submissions lists.
+ *}
+{assign var="formId" value="submissionsListFilter-"|concat:$filterData.gridId}
```
asmecher: Hmm, I wonder if there's not a cleaner way of exposing the right ID to the templates.

asmecher:
Thanks, this looks great! Reviewed.

controllers/grid/pubIds/PubIdExportSubmissionsListGridHandler.inc.php
asmecher: Oops! Yes, of course.

lib/pkp
asmecher: OK, thanks!

plugins/importexport/crossref/filter/ArticleCrossrefXmlFilter.inc.php
bozana: This is I believe for DB queries, but when you call it on the object itself, it gets its data...

classes/plugins/DOIPubIdExportPlugin.inc.php
bozana: could it be like this:
```
default:
	parent::manage($args, $request);
	$form->initData();
	return new JSONMessage(true, $form->fetch($request));
```

controllers/grid/pubIds/PubIdExportSubmissionsListGridHandler.inc.php
bozana: Ah, I totally forgot about the policies... :-\
Would this be OK, i.e. if the user is JM:
```
import('lib.pkp.classes.security.authorization.PolicySet');
$rolePolicy = new PolicySet(COMBINING_PERMIT_OVERRIDES);
import('lib.pkp.classes.security.authorization.RoleBasedHandlerOperationPolicy');
foreach($roleAssignments as $role => $operations) {
	$rolePolicy->addPolicy(new RoleBasedHandlerOperationPolicy($request, $role, $operations));
}
$this->addPolicy($rolePolicy);
return parent::authorize($request, $args, $roleAssignments);
```

controllers/grid/pubIds/PubIdExportIssuesListGridHandler.inc.php
bozana: :-)
Hmmm.. Because it is integration with load_url_in_div, maybe to implement a new CrossrefSubmissionListGridHandler that will extend this one and set the plugin variable?

templates/controllers/grid/pubIds/pubIdExportSubmissionsGridFilter.tpl
bozana: Hmmm... I believe I don't know how else would be better and work properly :-\

classes/plugins/DOIPubIdExportPlugin.inc.php
bozana: Actually, calling parent::manage does not make sense i.e is even wrong, because the parent has only this there: assert(false). Thus, I will remove it entirely, OK?

controllers/grid/pubIds/PubIdExportIssuesListGridHandler.inc.php
bozana: Hmmm... PageRouter cannot find my new CrossrefSubmissionListGridHandler in the crossref folder. I registered for the LoadComponentHandler hook, but it seems it is not passed through... :-(

classes/plugins/DOIPubIdExportPlugin.inc.php
asmecher: At some point we may want to e.g. add a hook to the parent implementation. I'd suggest using a specific value instead of `default:` and keeping the parent call for unhandled cases.

controllers/grid/pubIds/PubIdExportSubmissionsListGridHandler.inc.php
asmecher: OTOH that should be OK, but please give it a test!

controllers/grid/pubIds/PubIdExportIssuesListGridHandler.inc.php
asmecher: Hmm, perhaps the import/export plugin lifecycle loads to late for the component to be resolved. It's OK to get these from the request, but some checking for sanity is probably needed.

controllers/grid/pubIds/PubIdExportIssuesListGridHandler.inc.php
```diff
@@ -0,0 +1,228 @@
...
+	protected function getFilterForm() {
+		return 'controllers/grid/pubIds/pubIdExportIssuesGridFilter.tpl';
+	}
+
+	/**
+	 * @copydoc GridHandler::renderFilter()
+	 */
+	function renderFilter($request, $filterData = array()) {
+		$statusNames = $this->_plugin->getStatusNames();
+		$filterData = array(
```
asmecher: Just in case subclasses try to use the `$filterData` parameter, best to `array_merge` the data you want here with the stuff that's passed into the function.

templates/controllers/grid/pubIds/pubIdExportSubmissionsGridFilter.tpl
asmecher: Yeah, I didn't come up with anything worth pursuing. Go for it.

classes/plugins/DOIPubIdExportPlugin.inc.php
bozana: but the parent call makes an error i.e. assert(false)

asmecher: Yes, and I think that's appropriate. Just suggesting that the current `default:` operation should get a specific name (e.g. `case 'index':`); both `case` statements should return from the function once they've handled their operations. Only unhandled cases would go to the parent implementation.

bozana: Aaaaaaaaaaaa.... Well, my brain is not functioning so well... Now I got it... :-P

bozana: This should then be changed for each import/export plugin I believe...

bozana: Ah no, somewhere else...

bozana: Sorry, I found it :-\

controllers/grid/pubIds/PubIdExportSubmissionsListGridHandler.inc.php
bozana: Tested, as reviewer I cannot access it and also not as journal manager of another journal :-)

asmecher: OK, great!

controllers/grid/pubIds/PubIdExportIssuesListGridHandler.inc.php
bozana: @asmecher, maybe that's why also acron cannot be properly registered, but I am wondering how is this working in stable branch? Hmmm...
What do you think I could check? If plugin exists? If an importexport plugin? Hmmm...

bozana:
@asmecher, s. the new commit "code and ui review fixes". Just one code review comment is still left: to check the plugin request parameter. Also some CSS changes are needed for new buttons/link actions. But else, that's it, I think :-)

bozana:
@asmecher, I think it is your last commit to the pkp-lib ("Minor code cleanup") that is making troubles for tests -- when I run them without it, everything is fine. Could/would you take a look?

asmecher:
Sorry, I think I got it. Pushed up: https://github.com/pkp/pkp-lib/commit/d20b5eebff93bd2a784446cae5bca3ce53f2394c

asmecher:
Cherry-picked!

PR #898, opened 2016-06-29
revert archive send commit, adjust author submission status check

jnugent:
After discussion with Co Action, they'd prefer to not have the submission actually archived, but simply display a different submission status to the author.  This removes the previous fix to archive without sending, but also lets the _DECLINE decision fall through and return a _REVIEW status for the author instead.

### DIFF
--- a/classes/submission/author/AuthorSubmission.inc.php
+++ b/classes/submission/author/AuthorSubmission.inc.php
@@ -170,7 +170,7 @@ class AuthorSubmission extends Article {
 
 		$latestDecision = $this->getMostRecentDecision();
 		if ($latestDecision) {
-			if ($latestDecision == SUBMISSION_EDITOR_DECISION_ACCEPT || $latestDecision == SUBMISSION_EDITOR_DECISION_DECLINE) {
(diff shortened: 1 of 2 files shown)


jmacgreg:
Thanks Jason, that looks great to me!

jnugent:
Committed to ojs-2_4-dev and cherry picked to 2.4.8 stable.

asmecher:
I'm missing some context on this one. Was this a Coaction modification that made it into the core codebase sometime recently?

jnugent:
No but this reverts behavior that they decided they didn't want.  I don't think this fixes an earlier problem.  Just longstanding issue with how author submissions are shown on the author index page

asmecher:
OK, I'm not sure I follow but if you & James are behind it that's fine with me.

jnugent:
Alec, I didn't explain it very well.  The old behaviour was that if an Editor rejected a submission but did not send the notification email, it still showed as "In Editing" on the Author submission list.  CoAction would like that to say "In Review", but only for the Author, so by removing that decision test the case falls through.  When they do finally decide to send the email, the submission is properly archived.

asmecher:
OK, thanks!

PR #904, opened 2016-07-04
Added link id to editorialTeam page in about/index.tpl

marchitelli:
Added id for better styling if it is necessary


asmecher:
Merged, thanks! I'll also cherry-pick this for release in OJS 2.4.8-1, which will be our next stable release.

PR #905, opened 2016-07-04
OJS3 locale updates for all the locales

mtub:
This is updating all locales, see pkp/pkp-lib#1582

Resulting in a clear starting point for OJS3 translations. From now on, translations updates can target the master branch (when intended for OJS3).


asmecher:
MERGED! Thanks, @mtub.

PR #924, opened 2016-07-13
convert html entities to UTF-8

ubermichael:
The PLN staging server was tossing HTTP 500 errors for deposit XML that contained HTML entities like `&copy;` (which isn't well-formed XML anyway, so it's probably the right thing to do, although maybe it should be HTTP 400 since it isn't an internal error but a bad request).

This patch encodes all the metadata fields which might have entities as UTF-8, and also filters out an invalid UTF-8 sequences at the same time. 

Fixes pkp/lib-pkp#1607

### DIFF
--- a/plugins/generic/pln/classes/DepositPackage.inc.php
+++ b/plugins/generic/pln/classes/DepositPackage.inc.php
@@ -84,6 +84,27 @@ class DepositPackage {
 	 function getPackageFilePath() {
 		return $this->getDepositDir() . DIRECTORY_SEPARATOR . $this->_deposit->getUUID() . ".zip";
 	}
+    
+    /**
+     * Create a DOMElement in the $dom, and set the element name, namespace, and 
+     * content appropriately. Named HTML entities in the content are converted
+     * to UTF-8 characters. Any invalid UTF-8 characters will be dropped.
+     * 
+     * @param DOMDocument $dom
+     * @param string $elementName
+     * @param string $content
+     * @param string $namespace
+     * @return DOMElement
+     */
+	function _generateElement($dom, $elementName, $content, $namespace = null){
+		$encoded = html_entity_decode($content, ENT_NOQUOTES | ENT_HTML401, 'UTF-8');
+		$filtered = iconv('UTF-8', 'UTF-8//IGNORE', $encoded);
+		if($namespace !== null) {
+			return $dom->createElementNS($namespace, $elementName, $filtered);
+		} else {
+			return $dom->createElement($elementName, $filtered);
+		}
+	}
 
 	/**
 	 * Create an atom document for this deposit.
@@ -111,19 +132,19 @@ class DepositPackage {
 		$entry->setAttributeNS('http://www.w3.org/2000/xmlns/' ,'xmlns:dcterms', 'http://purl.org/dc/terms/');
 		$entry->setAttributeNS('http://www.w3.org/2000/xmlns/' ,'xmlns:pkp', 'http://pkp.sfu.ca/SWORD');
 		
-		$email = $atom->createElement('email', $journal->getSetting('contactEmail'));
+		$email = $this->_generateElement($atom, 'email', $journal->getSetting('contactEmail'));
 		$entry->appendChild($email);
 		
-		$title = $atom->createElement('title', $journal->getLocalizedTitle());
+		$title = $this->_generateElement($atom, 'title', $journal->getLocalizedTitle());
 		$entry->appendChild($title);
 		
-		$pkpJournalUrl = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'pkp:journal_url', $journal->getUrl());
+		$pkpJournalUrl = $this->_generateElement($atom, 'pkp:journal_url', $journal->getUrl(), 'http://pkp.sfu.ca/SWORD');
 		$entry->appendChild($pkpJournalUrl);
 
-		$pkpPublisher = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'pkp:publisherName', $journal->getSetting('publisherInstitution'));
+		$pkpPublisher = $this->_generateElement($atom, 'pkp:publisherName', $journal->getSetting('publisherInstitution'), 'http://pkp.sfu.ca/SWORD');
 		$entry->appendChild($pkpPublisher);
 
-		$pkpPublisherUrl = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'pkp:publisherUrl', $journal->getSetting('publisherUrl'));
+		$pkpPublisherUrl = $this->_generateElement($atom, 'pkp:publisherUrl', $journal->getSetting('publisherUrl'), 'http://pkp.sfu.ca/SWORD');
 		$entry->appendChild($pkpPublisherUrl);
 
 		$issn = '';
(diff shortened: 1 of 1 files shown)


plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -84,6 +84,27 @@ function getAtomDocumentPath() {
...
+     * 
+     * @param DOMDocument $dom
+     * @param string $elementName
+     * @param string $content
+     * @param string $namespace
+     * @return DOMElement
+     */
+	function _generateElement($dom, $elementName, $content, $namespace = null){
+		$encoded = html_entity_decode($content, ENT_NOQUOTES | ENT_HTML401, 'UTF-8');
+		$filtered = iconv('UTF-8', 'UTF-8//IGNORE', $encoded);
```
asmecher: `iconv` (historically at least) has not always been available; it might be safer to use `mb_convert_encoding` per e.g. http://stackoverflow.com/questions/8215050/replacing-invalid-utf-8-characters-by-question-marks-mbstring-substitute-charac

plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -84,6 +84,27 @@ function getAtomDocumentPath() {
 	 function getPackageFilePath() {
 		return $this->getDepositDir() . DIRECTORY_SEPARATOR . $this->_deposit->getUUID() . ".zip";
 	}
+    
+    /**
+     * Create a DOMElement in the $dom, and set the element name, namespace, and
```
asmecher: Looks like space-based indentation.

plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -111,19 +132,19 @@ function generateAtomDocument() {
 		$entry->setAttributeNS('http://www.w3.org/2000/xmlns/' ,'xmlns:dcterms', 'http://purl.org/dc/terms/');
 		$entry->setAttributeNS('http://www.w3.org/2000/xmlns/' ,'xmlns:pkp', 'http://pkp.sfu.ca/SWORD');
 		
-		$email = $atom->createElement('email', $journal->getSetting('contactEmail'));
+		$email = $this->_generateElement($atom, 'email', $journal->getSetting('contactEmail'));
```
asmecher: Some of these fields should definitely not contain entities -- fields like the contact email are displayed client-side using the Smarty `|escape` filter, which means that `&copy;` will be dumped out to the browser escaped. Generally entities would only come in via TinyMCE-enabled fields.

ctgraham:
Linking to pkp/pkp-lib#1607 .

plugins/generic/pln/classes/DepositPackage.inc.php
ctgraham: How about (untested):
```
if (html_entities($content) != $content) {
  $data = $dom->createCDATASection( $content );
} else {
   $data = $dom->createTextNode( $content );
}
if ($namespace !== null ) {
   $element = $dom->createElementNS($namespace, $elementName);
} else {
   $element = $dom->createElement($elementName);
}
return $element->appendChild($data);
```

ubermichael: CDATA could work, I guess. Double encoding will also create a compatibility problem - we have 3,000+ deposits now and some of them have properly encoded ampersands (Foo & Sons Publishers, for example). 

But you still need to filter out invalid UTF-8 characters in the data, even in the CDATA sections. I don’t have good internet access right now, otherwise I’d code it up.

ctgraham: So to clean the Unicode with `mb_convert_encoding()` we could add in:
```
$originalSubstitute = mb_substitute_character();
mb_substitute_character(mb_convert_encoding('&#xFFFD;', 'UTF-8', 'HTML-ENTITIES');)
$content = mb_convert_encoding($content, 'UTF-8', 'UTF-8');
mb_substitute_character($originalSubstitute);
```

plugins/generic/pln/classes/DepositPackage.inc.php
ubermichael: I figure it's possible that a journal could have bad byte sequences or enter the email as 
`"Bob Terwilliger" <bob@example.com>`

Best to be safe.

asmecher:
@ubermichael, I wonder if it's worth introducing some kind of revision number to the XML so that we can link the XML back to the code that produced it... If we have many twists and turns like this (which is not unlikely) then it would help. I'd consider either including a revision number that gets manually incremented when changes are introduced (prone to forgetfulness) or including the OJS version number in the deposit (automatic but not as immediate to determine whether there were changes). Or is the OJS version number already included in the deposit?

plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -84,6 +84,32 @@ function getAtomDocumentPath() {
...
+		// remove any invalid UTF-8.
+		$original = mb_substitute_character();
+		mb_substitute_character(0xFFFD);
+		$filtered = mb_convert_encoding($content, 'UTF-8', 'UTF-8');
+		mb_substitute_character($original);
+		
+		// put the filtered content in a CDATA, as it may contain markup that 
+		// isn't valid XML.
+		$node = $dom->createCDATASection($filtered);
+		$element = $dom->createElementNS($namespace, $elementName);
```
ctgraham: Does `createElementNS()` handle a null namespace?  I don't see that case in the PHP docs.

ubermichael:
The OJS version number isn't included, but may be useful information to have. But since people can manually update the plugin (even though it isn't a good idea), it would be useful to have that information as well.

Perhaps after https://github.com/ubermichael/ojs/blob/ojs-dev-2_4/plugins/generic/pln/classes/DepositPackage.inc.php#L138

```php
		$pluginVersionFile = $this->getPluginPath() . DIRECTORY_SEPARATOR . 'version.xml';
		$pluginVersion =& VersionCheck::parseVersionXml($pluginVersionFile);		
		$versionDao =& DAORegistry::getDAO('VersionDAO');
		$ojsVersion =& $versionDao->getCurrentVersion();
		$version = $atom->createElementNS('http://pkp.sfu.ca/SWORD', 'versions');
		$version->setAttribute('ojs', $ojsVersion);
		$version->setAttribute('plugin', $pluginVersion);
		$entry->appendChild($version);

```

plugins/generic/pln/classes/DepositPackage.inc.php
ubermichael: It does. If the namespace URI is null, it does the right thing.

plugins/generic/pln/classes/DepositPackage.inc.php
```diff
@@ -84,6 +84,32 @@ function getAtomDocumentPath() {
 	 function getPackageFilePath() {
 		return $this->getDepositDir() . DIRECTORY_SEPARATOR . $this->_deposit->getUUID() . ".zip";
 	}
+	
+	/**
+	 * Create a DOMElement in the $dom, and set the element name, namespace, and 
+	 * content appropriately. Named HTML entities in the content are converted
```
ctgraham: This docblock is no longer accurate.

ubermichael: Fixed.

asmecher:
Also cherry-picked to `ojs-stable-2_4_8`.

PR #927, opened 2016-07-14
Locales update: remove unnecessary lines and files

mtub:
This removes 

* obsolete keys from a bunch of locales (keys that  had been removed from the en_US locale some time ago)
* untranslated lines
* untranslated files (form the pl_PL locale)

I don't intend to cherry-pick this for `ojs-stable-2_4_8`.


locale/sv_SE/locale.xml
```diff
@@ -1026,135 +938,6 @@ Validering av URL:er kan misslyckas på sökningar som kräver inskickad data.]]
 	 * DEFAULT - Default settings for setup items.
 	 *
 	-->
-	<message key="default.journalSettings.checklist.notPreviouslyPublished">Manuskriptet har inte publicerats tidigare, och ligger inte för publicering i annan tidskrift (eller se bifogad förklaring under Kommentarer under Redaktör).</message>
```
asmecher: Why are these being removed?

locale/vi_VN/locale.xml
```diff
@@ -834,87 +821,6 @@ Việc xác thực có thể không thành công đối với các website tìm
 	 * DEFAULT - Default settings for setup items.
 	 *
 	-->
-	<message key="default.journalSettings.checklist.notPreviouslyPublished">Bài nộp chưa từng được xuất hoặc chưa từng được gửi cho tạp chí khác (hoặc được giải thích trong Lời bình gửi BTV).</message>
```
asmecher: Also all these?

locale/sv_SE/locale.xml
mtub: I think this is what happened: They are in the wrong file. The translator plugin identifies them as obsolete (does not check if a key has been moved/needs to be moved).

I have started reviewing everything (already hate it) and will repair the few keys that need to stay (will move them to the correct file, from locale.xml to default.xml), update this PR, and will then let you know. Thanks for spotting this, @asmecher .

mtub:
@asmecher , I've identified the lines that should not have been removed but where mis-classified by the translator plugin. I've moved the strings to the correct place. I think this is safe now.

asmecher:
Thanks as always, @mtub -- merging the three.

asmecher:
...make that two.

PR #939, opened 2016-07-21
pkp/pkp-lib#383 fix JS event target form submit

bozana:
the form to be submitted was not correctly determined: two forms (submissions and issues export form) have the same JS handler and both were submitted when clicking on an action link. now the form of the event target element is figured out/located and submitted.

### DIFF
--- a/js/controllers/form/DOIPubIdExportFormHandler.js
+++ b/js/controllers/form/DOIPubIdExportFormHandler.js
@@ -25,10 +25,34 @@
 			function($formElement, options) {
 		this.parent($formElement, options);
 
-		$('#exportSubmissionXmlForm a[id*="-button-"]').click(this.callbackWrapper(
-				this.submitAction_));
-		$('#exportIssueXmlForm a[id*="-button-"]').click(this.callbackWrapper(
-				this.submitAction_));
+		this.submissionsActions_ = [];
+		this.issuesActions_ = [];
+		this.actionRegExp_ = [];
+
+		var i;
+		if (options.submissionsActions) {
+			for (i = 0; i < options.submissionsActions.length; i++) {
+				/** @type {{_id:string}}*/
+				this.optionsSubmissionsAction = options.submissionsActions[i];
+				this.submissionsActions_.push(this.optionsSubmissionsAction._id);
+				$('#exportSubmissionXmlForm a[id^="' + this.optionsSubmissionsAction._id +
+					'-button-"]').click(this.callbackWrapper(this.submitAction_));
+			}
+			this.actionRegExp_.exportSubmissionXmlForm =
+					'(' + this.submissionsActions_.join('|') + ')';
+		}
+
+		if (options.issuesActions) {
+			for (i = 0; i < options.issuesActions.length; i++) {
+				/** @type {{_id:string}}*/
+				this.optionsIssuessAction = options.issuesActions[i];
+				this.issuesActions_.push(this.optionsIssuessAction._id);
+				$('#exportIssueXmlForm a[id^="' + this.optionsIssuessAction._id +
+					'-button-"]').click(this.callbackWrapper(this.submitAction_));
+			}
+			this.actionRegExp_.exportIssueXmlForm =
+					'(' + this.issuesActions_.join('|') + ')';
+		}
 
 	};
 	$.pkp.classes.Helper.inherits(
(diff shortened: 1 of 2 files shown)


js/controllers/form/DOIPubIdExportFormHandler.js
```diff
@@ -51,14 +51,12 @@
 	$.pkp.controllers.form.DOIPubIdExportFormHandler.
 			prototype.submitAction_ = function(submitButton, event) {
 
-		var $formElement = this.getHtmlElement(),
-				idPattern = new RegExp('(.*)-button-'),
-				button = event.target.id,
-				idPatternResult = idPattern.exec(button),
+		var idPattern = new RegExp('(.*)-button-'),
```
asmecher: I wonder if this will open us up to some kind of strange XSS attack. It might be safer to use `[a-z0-9]+` instead of `.*`

bozana:
@asmecher, would the second commit "fix JS form action" be a better solution: the JS will get the form actions buttons as parameter and only those will be also considered there?

js/controllers/form/DOIPubIdExportFormHandler.js
```diff
@@ -25,10 +25,34 @@
...
+			}
+			this.actionRegExp_.exportSubmissionXmlForm =
+					'(' + this.submissionsActions_.join('|') + ')';
+		}
+
+		if (options.issuesActions) {
+			for (i = 0; i < options.issuesActions.length; i++) {
+				/** @type {{_id:string}}*/
+				this.optionsIssuessAction = options.issuesActions[i];
+				this.issuesActions_.push(this.optionsIssuessAction._id);
```
asmecher: Sometimes you have `Issues` and sometimes `Issuess` -- needs fix & test.

js/controllers/form/DOIPubIdExportFormHandler.js
```diff
@@ -25,10 +25,34 @@
...
-				this.submitAction_));
+		this.submissionsActions_ = [];
+		this.issuesActions_ = [];
+		this.actionRegExp_ = [];
+
+		var i;
+		if (options.submissionsActions) {
+			for (i = 0; i < options.submissionsActions.length; i++) {
+				/** @type {{_id:string}}*/
+				this.optionsSubmissionsAction = options.submissionsActions[i];
```
asmecher: I think `optionsSubmissionsAction` needs to be defined in the class, no?

js/controllers/form/DOIPubIdExportFormHandler.js
```diff
@@ -25,10 +25,34 @@
...
+				this.submissionsActions_.push(this.optionsSubmissionsAction._id);
```
asmecher: Hmm, _id comes from the PHP object via `json_encode`, correct? I'd rather not access a private/protected property directly.

js/controllers/form/DOIPubIdExportFormHandler.js
bozana: Ah, sorry @asmecher, I solved it in my recent PR for medra, s. https://github.com/pkp/ojs/pull/949/commits/8e1a79ad2b46f132ee19b25928326b29282eca95#diff-7 I will close this PR and add the changes according to your comments heere there, OK?

asmecher:
Thanks, but that approach still feels confusing to me, and I'm hoping something is possible that's conceptually cleaner and involves less JS. What about considering the way the buttons are being built, e.g. moving them into the grids or something? (I know, I'm a broken record on that one, just a suggestion :)

js/controllers/form/DOIPubIdExportFormHandler.js
asmecher: OK, thanks!

bozana:
Hmmm... I think we cannot move those actions to grid easily, because the grid doesn't know about the selected checkboxes i.e. there is a form around it. Maybe then to use the normal submit buttons instead of action links + JS? -- I used the action links in order to style them as they are now, but... ;-)

asmecher:
Yes, that might be worth a try. Failing that, I'd prefer the earlier approach, just because it uses a lot less bespoke Javascript and both approaches use JS RegExps, which we're not using elsewhere.

asmecher:
BTW, it looks like Travis testing has started working again; their Twitter feed notes an outage/queuing problems on the 25th/26th. Looks like it's recovered now.

bozana:
Hmmm... Now I am not sure what solution would you suggest i.e. what earlier solution you prefer and what would be worth a try :-)

asmecher:
I mean that I preferred the approach I posted https://github.com/pkp/ojs/pull/939#discussion_r71938547 on to the later one with the extended JS -- but I like your suggestion of trying normal submit buttons, as I thought something felt wrong about both other approaches.

bozana:
OK, I will then try the good old normal submit buttons :-) I will now close this PR and do it with other changes here: https://github.com/pkp/ojs/pull/949/

PR #949, opened 2016-07-26
Implement mEDRA DOI export plugin

bozana:
S. https://github.com/pkp/pkp-lib/issues/1603


classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -451,6 +451,71 @@ function deleteAllPubIds($journalId, $pubIdType) {
...
+		$result = $this->retrieveRange(
+				'SELECT	g.*, sf.*
+			FROM	submission_galleys g
+				' . ($contextId?' JOIN submissions s ON (s.submission_id = g.submission_id)':'') . '
+				LEFT JOIN published_submissions ps ON (ps.submission_id = g.submission_id)
+				LEFT JOIN issues i ON (ps.issue_id = i.issue_id AND i.date_published IS NOT NULL)
+				LEFT JOIN submission_files sf ON (g.file_id = sf.file_id)
+				LEFT JOIN submission_files nsf ON (nsf.file_id = g.file_id AND nsf.revision > sf.revision AND nsf.file_id IS NULL )
+				LEFT JOIN submission_galley_settings gs ON (g.galley_id = gs.galley_id)
+				'. ($title?' LEFT JOIN submission_settings sst ON (s.submission_id = sst.submission_id)':'')
```
asmecher: I think this should search in the user's current locale only, no? Otherwise you might get multiple results for the same submission.

asmecher: Also, better to use strict comparisons when accepting optional string parameters. If someone tries to search for "false" as a substring, the parameter will be ignored.

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -100,16 +105,166 @@ function manage($args, $request) {
...
+				// Add link actions
+				$actions = $this->getExportActions($context);
+				$actionNames = array_intersect_key($this->getExportActionNames(), array_flip($actions));
+				import('lib.pkp.classes.linkAction.request.NullAction');
+				$linkActions = array();
+				foreach ($actionNames as $action => $actionName) {
+					$linkActions[] = new LinkAction($action, new NullAction(), $actionName);
+				}
+				$templateMgr = TemplateManager::getManager($request);
+				$templateMgr->assign('plugin', $this);
```
asmecher: It's a little tidier to do batch assigns using the array form of the function:

    $templateMgr->assign(array(
        'one' => 1,
        'two' => 2,
    ));

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -100,16 +105,166 @@ function manage($args, $request) {
...
+				$templateMgr->assign('exportRepresentations', $exportRepresentations);
+				$templateMgr->display($this->getTemplatePath() . 'index.tpl');
+				break;
+			case 'exportSubmissions':
+			case 'exportIssues':
+				$selectedSubmissions = (array) $request->getUserVar('selectedSubmissions');
+				$selectedIssues = (array) $request->getUserVar('selectedIssues');
+
+				if (empty($selectedSubmissions) && empty($selectedIssues)) {
+					echo __('plugins.importexport.common.error.noObjectsSelected');
```
asmecher: If this is dumping an error message to the screen, maybe better to use a template, or `fatalError` if it's not something that will frequently be presented to the user.

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -165,6 +340,61 @@ function getPubIdDisplayType() {
...
+		return $xml;
+	}
+
+	/**
+	 * Deposit XML i.e. publication objects DOIs.
+	 * This must be implemented in the subclasses, if the action is supported.
+	 * @param $objects array Array of published articles, issues or galleys
+	 * @param $context Context
+	 * @param $filename Export XML filename
+	 * @return boolean Weather the DOI has been registered/found
```
asmecher: *whether

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -269,13 +499,14 @@ function updateObject($object) {
 
 	/**
 	 * Identify the dao name and update method for the given object.
-	 * @param $object Issue|PublishedArticle
+	 * @param $object Issue|PublishedArticle|ArticleGalley
 	 * @return array (DAO, Method)
 	 */
 	function getDaoMethod($object) {
 		$configurations = array(
```
asmecher: Now that all the functions are named the same, can this be cleaned up/removed?

controllers/grid/pubIds/PubIdExportRepresentationsListGridHandler.inc.php
```diff
@@ -0,0 +1,314 @@
...
+
+
+	//
+	// Implemented methods from GridHandler.
+	//
+	/**
+	 * @copydoc GridHandler::getRowInstance()
+	 */
+	function getRowInstance() {
+		return new GridRow();
```
asmecher: I think you can remove this function.

plugins/importexport/crossref/CrossrefInfoSender.inc.php
```diff
@@ -63,102 +62,35 @@ function executeActions() {
...
 			// Notify journal managers if there is a new failed DOI status
 			if ($notify) {
 				$roleDao = DAORegistry::getDAO('RoleDAO');
-				$journalManagers = $roleDao->getUsersByRoleId(ROLE_ID_JOURNAL_MANAGER, $journal->getId());
+				$journalManagers = $roleDao->getUsersByRoleId(ROLE_ID_MANAGER, $journal->getId());
 				import('classes.notification.NotificationManager');
 				$notificationManager = new NotificationManager();
 				while ($journalManager = $journalManagers->next()) {
 					$notificationManager->createTrivialNotification($journalManager->getId(), NOTIFICATION_TYPE_ERROR, array('contents' => __('plugins.importexport.crossref.notification.failed')));
 					unset($journalManager);
```
asmecher: No need to `unset` unless you're doing a `foreach` by reference.

plugins/importexport/crossref/CrossrefInfoSender.inc.php
```diff
@@ -189,12 +121,77 @@ function _getJournals() {
...
+	 * @param $objects array
+	 * @param $filter string
+	 * @param $journal Journal
+	 */
+	function _registerObjects($objects, $filter, $journal) {
+		$plugin = $this->_plugin;
+		// export XML
+		$exportXml = $plugin->exportXML($objects, $filter, $journal);
+		// Write the XML to a file.
+		$exportFileName = $plugin->getExportPath() . date('Ymd-His') . '.xml';
```
asmecher: Does this need to be disambiguated with the journal name, or is the export path already journal-specific?

plugins/importexport/crossref/classes/form/CrossRefSettingsForm.inc.php
```diff
@@ -140,7 +141,7 @@ function getFormFields() {
 	 * @return boolean
 	 */
 	function isOptional($settingName) {
-		return in_array($settingName, array('username', 'password', 'automaticRegistration'));
+		return in_array($settingName, array('username', 'password', 'automaticRegistration', 'testMode'));
```
asmecher: Happy to see test mode moving to a setting rather than an obscure URL parameter :)

plugins/importexport/medra/MedraExportPlugin.inc.php
```diff
@@ -13,123 +13,147 @@
  * @brief mEDRA Onix for DOI (O4DOI) export/registration plugin.
  */
 
-if (!class_exists('DOIExportPlugin')) { // Bug #7848
-	import('plugins.importexport.medra.classes.DOIExportPlugin');
-}
+import('classes.plugins.DOIPubIdExportPlugin');
```
asmecher: Phew, glad to see that go.

plugins/importexport/medra/README
```diff
@@ -17,16 +17,23 @@ complete terms of this license.
...
+Contact/Support
+---------------
+For support, bugfixes, or comments please use PKP-LIB GitHub issues: https://github.com/pkp/pkp-lib/issues
 
 Note
 ---------
 In order to register DOIs with mEDRA from within OJS you will have to enter your username and password. 
 If you do not enter or have your own username and password you'll still be able to export into the 
 Onix for DOI (O4DOI) format, but you cannot register your DOIs from within OJS.
 Please note, that the passowrd will be saved as plain text, i.e. not encripted, due to mEDRA registration service requirements.
```
asmecher: *encrypted

plugins/importexport/medra/classes/DOIExportDom.inc.php
```diff
@@ -1 +0,0 @@
-../../datacite/classes/DOIExportDom.inc.php
```
asmecher: Woo!

plugins/importexport/medra/classes/form/MedraSettingsForm.inc.php
```diff
@@ -41,17 +84,29 @@ function MedraSettingsForm(&$plugin, $journalId) {
 		$this->addCheck(new FormValidatorInSet($this, 'publicationCountry', FORM_VALIDATOR_REQUIRED_VALUE, 'plugins.importexport.medra.settings.form.publicationCountry', array_keys($this->_getCountries())));
 		// The username is used in HTTP basic authentication and according to RFC2617 it therefore may not contain a colon.
 		$this->addCheck(new FormValidatorRegExp($this, 'username', FORM_VALIDATOR_OPTIONAL_VALUE, 'plugins.importexport.medra.settings.form.usernameRequired', '/^[^:]+$/'));
+		$this->addCheck(new FormValidatorPost($this));
```
asmecher: Once the CSRF stuff is merged in, we'll need to add the CSRF checks to the code in this PR too -- let's talk about that later.

asmecher:
(Incidentally, I'm not sure why Travis has stopped testing pull requests, but it happened for some of Nate's PRs, too -- I'll have to look tomorrow.)

classes/plugins/DOIPubIdExportPlugin.inc.php
bozana: What do you mean here?

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -165,6 +340,61 @@ function getPubIdDisplayType() {
...
+		}
+	}
+
+	/**
+	 * Update deposit status.
+	 * The function must be implemented in the subclasses, if the action is supported.
+	 * @param $context Context
+	 * @param $object The object getting deposited
+	 */
+	function updateDepositStatus($context, $object) { }
```
bozana: Hmmm... this is implemented only in Crossref plugin, but used in checkStatus function in this class. CheckStatus is also used only in Crossref plugin, thus I think it is best to remove it from here and put it in the CrossrefExportPlugin class.

plugins/importexport/crossref/CrossrefInfoSender.inc.php
bozana: Wow, good point! I wonder it worked without problems till now! Or maybe it didn't and it happened already that a journal registered the DOIs from another journal in the same installation, in case the both exports were done at the same time. But maybe this was never the case. Phew... :-)

bozana:
Good that you trust my judgement, although I myself am not sure :-DDD ;-)
I run the tests locally, they were OK, so this with Travis is not crutial for this PR for now i.e. it can wait till I submit the changes according to your comments...
THANKS A LOT!!!

plugins/importexport/crossref/CrossrefInfoSender.inc.php
bozana: @asmecher, ah, the file names contained the pub object name and id e.g. ...-article-1.xml. This, file name differentiation, should be enough/OK, right?

asmecher: All I see here is the date -- where does the other info come in?

plugins/importexport/medra/MedraExportPlugin.inc.php
```diff
@@ -13,123 +13,147 @@
...
+			DOI_EXPORT_ACTION_DEPOSIT => __('plugins.importexport.common.action.register'),
+			DOI_EXPORT_ACTION_EXPORT => __('plugins.importexport.common.action.export'),
+			DOI_EXPORT_ACTION_MARKREGISTERED => __('plugins.importexport.common.action.markRegistered'),
+		);
+	}
+
+	/**
+	 * @copydoc DOIPubIdExportPlugin::getPluginSettingsPreffix()
+	 */
+	function getPluginSettingsPreffix() {
```
asmecher: *Prefix

plugins/importexport/crossref/CrossrefInfoSender.inc.php
bozana: yes, yes, that is my failure here, but in stable is as I said above, so I would change it in the same way if OK :-)

asmecher: Ah, OK -- yes, that makes sense!

classes/plugins/DOIPubIdExportPlugin.inc.php
asmecher: Just thinking structurally that we can make some improvements as we standardize the function naming etc. For example, in the DataObject class, we could add

    function getDAO() {
        assert(false);
    }

...then start to implement this in specific subclasses. Then this function could be entirely removed, and the `DOIPubIdExportPlugin::updateObject` function could call `getDAO` on the object and then invoke `updateObject` on that directly.

classes/article/ArticleGalleyDAO.inc.php
bozana: Do you mean " != null"? :-)

bozana: You are right for the current locale -- I think it is ok to restrict it to the current locale, right?
I believe those sql locale checks for articles are wrong in general -- they look if there is current and primary locale, but the primary locale has no sense for an article, it doesn't necessarily have to be in the primary locale -- it has it's own submission locale.

classes/plugins/DOIPubIdExportPlugin.inc.php
bozana: This tidiness obsessed people... :-DDD ;-) Joking, I agree... :-)

bozana:
I considered all code review comments, s. commit "changes after the code review" above + this lib-pkp PR https://github.com/pkp/pkp-lib/pull/1657.

asmecher:
Merged, thanks! That's a lot of work.

PR #956, opened 2016-08-04
medra locale fixes

bozana:
small fixes: adding a missing EN key and removing a IT key that is not used.


asmecher:
Merged, thanks!

PR #972, opened 2016-08-12
SQL fixes to get only published objects with DOIs

bozana:
For some reason, in one query case it was important to have the 'i.published = 1' at the very beginning of WHERE clause... :-O
And a typo...

### DIFF
--- a/classes/article/ArticleGalleyDAO.inc.php
+++ b/classes/article/ArticleGalleyDAO.inc.php
@@ -493,7 +493,7 @@ class ArticleGalleyDAO extends RepresentationDAO implements PKPPubIdPluginDAO {
 			FROM	submission_galleys g
 				' . ($contextId != null?' JOIN submissions s ON (s.submission_id = g.submission_id)':'') . '
 				LEFT JOIN published_submissions ps ON (ps.submission_id = g.submission_id)
-				LEFT JOIN issues i ON (ps.issue_id = i.issue_id)
+				JOIN issues i ON (ps.issue_id = i.issue_id)
 				LEFT JOIN submission_files sf ON (g.file_id = sf.file_id)
 				LEFT JOIN submission_files nsf ON (nsf.file_id = g.file_id AND nsf.revision > sf.revision AND nsf.file_id IS NULL )
 				LEFT JOIN submission_galley_settings gs ON (g.galley_id = gs.galley_id)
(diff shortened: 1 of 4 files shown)


classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -493,15 +493,15 @@ function getByPubIdType($pubIdType, $contextId = null, $title = null, $author =
 			FROM	submission_galleys g
 				' . ($contextId != null?' JOIN submissions s ON (s.submission_id = g.submission_id)':'') . '
 				LEFT JOIN published_submissions ps ON (ps.submission_id = g.submission_id)
-				LEFT JOIN issues i ON (ps.issue_id = i.issue_id AND i.date_published IS NOT NULL)
+				LEFT JOIN issues i ON (ps.issue_id = i.issue_id)
```
asmecher: I think the problem is the `LEFT JOIN`, which would return results when an `issue` entry did not match. You could achieve the same more clearly, I think, by changing the `LEFT JOIN` to a `JOIN` and removing the `date_published` requirement entirely (since it's `NOT NULL` per the table definition).

bozana: @asmecher, I will try. Could you also take a look at the changes in the IssueDAO below? There I also had to move i.published at the beginning...

asmecher: Looking at `i.published` to check for a published issue is fine; `issues` entries can either be published (`published=1`) or unpublished (`published=0`). But watch out for the order of the parameter array -- I think moving `getCompletionConditions` up potentially puts them out of order.

bozana: that is ok, I believe -- `getCompletionConditions` does not have any parameters...

bozana:
@asmecher, I did as you suggested + exchanged the `getCompletionJoins` and `getCompletionConditions` with JOIN and `i.published = 1`. It works. OK so?Could you please take another look/review it?

classes/article/ArticleGalleyDAO.inc.php
asmecher: Ah, of course, and what you're moving has no parameters either. Never mind.

classes/issue/IssueDAO.inc.php
```diff
@@ -586,12 +586,12 @@ function getByPubIdType($pubIdType, $contextId = null, $pubIdSettingName = null,
 				LEFT JOIN issue_settings ist ON (i.issue_id = ist.issue_id)
 				'. ($pubIdSettingName != null?' LEFT JOIN issue_settings iss ON (i.issue_id = iss.issue_id AND iss.setting_name = ?)':'') .'
 			WHERE
-				ist.setting_name = ? AND ist.setting_value IS NOT NULL
+				i.published = 1 AND ist.setting_name = ? AND ist.setting_value IS NOT NULL
```
asmecher: I still can't figure out how moving `i.published` around in this query would have any effect. Are you sure it fixes something?

bozana: yes, I don't understand it either, but it is so. For example this query, searching for unregistered issues:
```
SELECT i.*
            FROM issues i
                LEFT JOIN custom_issue_orders o ON (o.issue_id = i.issue_id)
                LEFT JOIN issue_settings ist ON (i.issue_id = ist.issue_id)
               LEFT JOIN issue_settings iss ON (i.issue_id = iss.issue_id AND iss.setting_name = 'crossref::registeredDoi')
            WHERE
                i.published = 1 AND ist.setting_name = 'pub-id::doi' AND ist.setting_value IS NOT NULL
                AND i.journal_id = 1
                AND iss.setting_value IS NULL OR iss.setting_value = ''
                ORDER BY i.date_published DESC 
```

If you change it below, you will also get the unpublished issues...

asmecher: Oh! It looks like that OR condition needs to be bracketed.

bozana: Wow, yes, you are right, it makes sense! I will do that as well, but i will leave the `i.published = 1` above, ok?

asmecher: Sure, that change doesn't bother me.

bozana:
One more commit, thanks a lot @asmecher!!!

asmecher:
Merged, thanks!

bozana:
Thank you! :-)

PR #978, opened 2016-08-18
Use https on creative commons license image

samueloph:
This is needed to prevent OJS installations that uses https to become a mixed content page (ojs links a http image on the article's page).

I have made one commit to change this on master as well, but it looks like OJS 3 won't link the license image, so the commit just change a href that doesn't "downgrade" OJS to a mixed content page, although i believe its a nice change too.
I can open a PR for master too, here is the master commit: https://github.com/samueloph/ojs/commit/0d0a4ce884c902beff91d2ce955de4fab37f00bd.

PS: I really didn't like that much to mess with docs/dev/TechnicalReference.pdf too, but i can easily drop that change (or whoever commits this can do it).


samueloph:
I just realised that master does have this too, but it got moved to pkp-lib, i'm doing another PR there to address this.

locale/ar_IQ/locale.xml
```diff
@@ -55,12 +55,12 @@
 	<message key="submission.license.cc.by-nd4">CC Attribution-NoDerivatives 4.0</message>
 	<message key="submission.license.cc.by4">CC Attribution 4.0</message>
 	<message key="submission.license.cc.by-sa4">CC Attribution-ShareAlike 4.0</message>
-	<message key="submission.license.cc.by-nc-nd4.footer"><![CDATA[<a rel="license" href="http://creativecommons.org/licenses/by-nc-nd/4.0/"><img alt="Creative Commons License" style="border-width:0" src="http://i.creativecommons.org/l/by-nc-nd/4.0/88x31.png" /></a><br />هذا العمل مرخص حسب <a rel="license" href="http://creativecommons.org/licenses/by-nc-nd/4.0/">Creative Commons Attribution-NonCommercial-NoDerivatives 4.0 International License</a>.]]></message>
```
ctgraham: This change and similar are unnecessary.  The use of http in the `<a>` tag is fine and will not generate cross-protocol warnings.

samueloph: Agreed, but i did use sed to replace http with https, so this one went together and i thought it would be nice to use https on href too, what do you think?

ctgraham:
Please limit this change only to the use of the URL which would trigger a cross-protocol violation, such as in an `` or `<link href="">`

locale/ar_IQ/locale.xml
ctgraham: Currently creativecommons.org seems to be automatically redirecting http traffic to https, which I suppose works in favor of this broader change.  The change still feels too broad to me.  When you use the [Choose a License tool](https://creativecommons.org/choose/), they currently output code with the `<a href>` as http, and the `` as https.  I think we should mimic that model.

Commifreak:
Is there a chance, that the img-tags get updated in the next few updates? I dont see, that this change is applied, yet?

Also, if the img tags getting updated by you: Note the new image urls:

That changed from **http://i.creativecommons.org/l/by/4.0/88x31.png** to **https://licensebuttons.net/l/by/4.0/88x31.png**

As example.

asmecher:
@Commifreak, which of our apps are you using and what version?

Commifreak:
Oh, sorry. OJS 2.4.8-1

Gruß,
R. Kluth

asmecher:
@Commifreak, I've back-ported this to 2.x with the following commits:

- https://github.com/pkp/pkp-lib/commit/7f37a4d7c8ad290ec292662890fcf3ecad22c1f4 (`lib/pkp` submodule)
- https://github.com/pkp/ojs/commit/31945ecdf225b303a20c5675f3cbacc261b04135 (OJS application)

Let me know if these work how you expect.

Commifreak:
Why did you not updated the licensebuttons img link in this opportunity? :P 

Yes, it is working now. The green lock remains green in the PDF-view, now with secured img src tags ;)

asmecher:
Yep, I got so excited about the protocol-relative URLs that I missed the domain change:

- https://github.com/pkp/ojs/commit/03e1c03d5acdcf32061a895e9622533bc38eae58
- https://github.com/pkp/pkp-lib/commit/4259ddd044dba41355e5ae6457ddf6f867d04e0f

Commifreak:
Thnaks :) Working as expected.

PR #984, opened 2016-08-23
Updated German translation of OJS 3.0

mtub:
* updated de_DE locale (mostly: manager.xml and last minute updates to OJS 3)
* minor updated to en_US locale (errors detected while reviewing)

… pending another round of de_DE review and comparison with changes suggested earlier, but let's merge this since it's ready.


asmecher:
Looks good to the extent of my German -- which is to say, the XML validates :) Thanks, @mtub.

PR #989, opened 2016-08-24
Easier CSS application without pseudo classes

ojsbsb:
Recently, we have been applying complex CSS to an OJS about page. As most `li` items did not have ids, it was necessary to use pseudo classes such as nth-child. This resulted in very complex and fragile css, which was not working well in all browsers.

The attached commit contains those changes which allow for much easier and human-readable css files to be used for OJS themes. We have more of this kind of changes, but we'd like to see first whether you are interested in reusing it.

Also, this commit allows to style the searchField select field's label separately from the field, which was not possible on all browsers (as the select box was part of the label, styles or visibility settings were applied to both with no reliable way of limiting it only to the label wording).

### DIFF
--- a/templates/about/index.tpl
+++ b/templates/about/index.tpl
@@ -12,16 +12,16 @@
 {assign var="pageTitle" value="about.aboutTheJournal"}
 {include file="common/header.tpl"}
 {/strip}
-<div id="aboutPeople">
+<div id="people">
 <h3>{translate key="about.people"}</h3>
 <ul>
 	{if not (empty($journalSettings.mailingAddress) && empty($journalSettings.contactName) && empty($journalSettings.contactAffiliation) && empty($journalSettings.contactMailingAddress) && empty($journalSettings.contactPhone) && empty($journalSettings.contactFax) && empty($journalSettings.contactEmail) && empty($journalSettings.supportName) && empty($journalSettings.supportPhone) && empty($journalSettings.supportEmail))}
-		<li><a href="{url op="contact"}">{translate key="about.contact"}</a></li>
+		<li id="contact"><a href="{url op="contact"}">{translate key="about.contact"}</a></li>
 	{/if}
-	<li id="editorialTeamLink"><a href="{url op="editorialTeam"}">{translate key="about.editorialTeam"}</a></li>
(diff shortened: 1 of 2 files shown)


templates/about/index.tpl
```diff
@@ -12,16 +12,16 @@
...
 	{if not (empty($journalSettings.mailingAddress) && empty($journalSettings.contactName) && empty($journalSettings.contactAffiliation) && empty($journalSettings.contactMailingAddress) && empty($journalSettings.contactPhone) && empty($journalSettings.contactFax) && empty($journalSettings.contactEmail) && empty($journalSettings.supportName) && empty($journalSettings.supportPhone) && empty($journalSettings.supportEmail))}
-		<li><a href="{url op="contact"}">{translate key="about.contact"}</a></li>
+		<li id="contact"><a href="{url op="contact"}">{translate key="about.contact"}</a></li>
 	{/if}
-	<li id="editorialTeamLink"><a href="{url op="editorialTeam"}">{translate key="about.editorialTeam"}</a></li>
+	<li id="editorialTeam"><a href="{url op="editorialTeam"}">{translate key="about.editorialTeam"}</a></li>
 	{if $peopleGroups}
 		{iterate from=peopleGroups item=peopleGroup}
-			<li><a href="{url op="displayMembership" path=$peopleGroup->getId()}">{$peopleGroup->getLocalizedTitle()|escape}</a></li>
+			<li id="{$peopleGroup->getId()}"><a href="{url op="displayMembership" path=$peopleGroup->getId()}">{$peopleGroup->getLocalizedTitle()|escape}</a></li>
```
asmecher: I would suggest prefixing this with something, rather than just using the numeric ID.

templates/about/index.tpl
```diff
@@ -12,16 +12,16 @@
 {assign var="pageTitle" value="about.aboutTheJournal"}
 {include file="common/header.tpl"}
 {/strip}
-<div id="aboutPeople">
```
asmecher: People may already be using this for styling; better not to change it, I think.

asmecher:
This looks sensible -- thanks, @ojsbsb.

PR #990, opened 2016-08-25
Proxy support for DataVerse

Commifreak:
DataVerse plugin does not support proxy right now. Now it does.

Can someone review this and include this maybe? :)

### DIFF
--- a/plugins/generic/dataverse/DataversePlugin.inc.php
+++ b/plugins/generic/dataverse/DataversePlugin.inc.php
@@ -1716,6 +1716,13 @@ class DataversePlugin extends GenericPlugin {
 	 * @return SWORDAPPClient
 	 */
 	function _initSwordClient($options = array(CURLOPT_SSL_VERIFYPEER => FALSE)) {
+		if ($httpProxyHost = Config::getVar('proxy', 'http_host')) {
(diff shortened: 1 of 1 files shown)


axfelix:
Looks fine to me -- although is it hardcoded to an HTTP and not an HTTPS proxy?

Commifreak:
Dont know. I got this from an internal code from ojs/pkp. I dont know how the Config class is working. Is the third param a fallback value?

Please adapt it if needed :)

Gruß,
R. Kluth

Commifreak:
Yep - 3rd parameter is default value, so there is no hardcoded http-pnly thing. If you mean that.

http_proxy contains only the proxy IP address. Curl does the rest.

I think the code is ok 👍

axfelix:
OK -- haven't had the chance to test as I don't have a suitable proxy setup, but given that the implementation is identical to https://github.com/pkp/ojs/blob/1394d9d4db192bc85ea97fe9f4cea2afdfb71672/plugins/importexport/datacite/DataciteExportPlugin.inc.php and reported as working, I'm merging this. Thanks for the contribution!

Commifreak:
Exactly :)

Thanks ;)

Gruß,
R. Kluth

PR #1003, opened 2016-08-27
pkp/pkp-lib#1604 doaj plugin

bozana:
S. https://github.com/pkp/pkp-lib/issues/1604


classes/plugins/PubObjectsExportPlugin.inc.php
```diff
@@ -0,0 +1,523 @@
...
+		return $xml;
+	}
+
+	/**
+	 * Display XML validation errors.
+	 * @param $errors array
+	 * @param $xml string
+	 */
+	function displayXMLValidationErrors($errors, $xml) {
+		echo '<h2>Validation errors:</h2>';
```
asmecher: At some point these should be localized, but OK if that's not done for the 3.0 release.

classes/plugins/PubObjectsExportPlugin.inc.php
```diff
@@ -0,0 +1,523 @@
...
+	function getDepositStatusSettingName() {
+		return $this->getPluginSettingsPrefix().'::status';
+	}
+
+
+
+	/**
+	 * @copydoc PKPImportExportPlugin::usage
+	 */
+	function usage($scriptName) {
```
asmecher: Don't be afraid to use `abstract` anymore :)

classes/plugins/PubObjectsExportPlugin.inc.php
bozana: Ah I did localized them. Sorry, I will change that...

plugins/importexport/doaj/DOAJExportDom.inc.php
```diff
@@ -1,435 +0,0 @@
...
-			case "tr": return "tur"; break;
-			case "tw": return "twi"; break;
-			case "ug": return "uig"; break;
-			case "uk": return "ukr"; break;
-			case "ur": return "urd"; break;
-			case "uz": return "uzb"; break;
-			case "ve": return "ven"; break;
-			case "vi": return "vie"; break;
-			case "vo": return "vol"; break;
-			case "wa": return "wln"; break;
```
asmecher: What happened to all these country codes? Did you find something matching in e.g. CountryDAO?

asmecher:
Looks good, @bozana! I had planned not to work on this plugin until later, when we could dedicate some time to their CRUD API (https://github.com/pkp/pkp-lib/issues/850), but actually this is an important enough thing for us that I'm very happy to have something to include in 3.0.

classes/plugins/PubObjectsExportPlugin.inc.php
bozana: May I leave them for now, because the child classes don't have them implemented for now?

plugins/importexport/doaj/DOAJExportDom.inc.php
bozana: This is ISO639-2b, right? Because we do not use language metadata filed any more i.e. here not every language is possible any more, but just article locale i.e. a locale that we have listed in `registry/locales.xml`, I could use the function `PKPLocale::get3LetterIsoFromLocale` i.e. the code from `registry/locales.xml`. Is this correct?

classes/plugins/PubObjectsExportPlugin.inc.php
asmecher: Sure.

plugins/importexport/doaj/DOAJExportDom.inc.php
asmecher: I suspect the original DOAJ plugin was written before we had `$article->getLocale`, and just had `$article->getLanguage`, which was free-form and only had a suggestion to use the two-letter code. So yes, I think your approach is correct and better.

bozana:
@asmecher, I rebased and added one more commit -- to localize those xmlvalidation error messages...

asmecher:
Merged, thanks!

PR #1004, opened 2016-08-29
[crossref] Update pt-br and pt-pt translation

samueloph:
OJS 2.4.8-1 introduced some big changes on the crossref plugin strings and it looks like no translation were made at all, OJS should warns about this because it is seriously breaking the usability of 2.4.8-1 to every journal editor that doesn't uses en_US.

I made a patch that updates the translation for pt_BR and pt_PT.

@mtub Can you have a look at this?

Thanks!

### DIFF
--- a/plugins/importexport/datacite/locale/pt_PT/common.xml
+++ b/plugins/importexport/datacite/locale/pt_PT/common.xml
@@ -2,55 +2,67 @@
 <!DOCTYPE locale SYSTEM "../../../../../lib/pkp/dtd/locale.dtd">
 
 <!--
-  * @file plugins/importexport/medra/locale/pt_PT/common.xml
+  * plugins/importexport/datacite/locale/pt_PT/common.xml
   *
   * Copyright (c) 2013-2016 Simon Fraser University Library
   * Copyright (c) 2003-2016 John Willinsky
   * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
   *
-  * Shared localization strings
+  * Shared localization strings.
   -->
 
 <locale name="pt_PT" full_name="Português (Portugal)">
-	<message key="plugins.importexport.common.settings.form.password.description">Tenha em atenção que a senha será guardada como texto, não encriptada.</message>
-	<message key="plugins.importexport.common.markRegistered">Marcar como registado</message>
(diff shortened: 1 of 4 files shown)


samueloph:
Please note that the Travis CI is failing but its not because of this PR.

asmecher:
@samueloph, the Google Closure compiler we were using has been removed from its old location; we're aware of the issue (see https://github.com/pkp/pkp-lib/issues/1761).

samueloph:
@mtub Could you please have a look at this?

samueloph:
Hey folks,

Could this be taken more seriously?
Almost every brazilian user is having problems with this regression and i bet every non-english user too, since it doesn't look like other translations were updated. Users should be warned when upgrading to 2.4.8-1.

ctgraham:
+1

plugins/importexport/datacite/locale/pt_PT/common.xml
```diff
@@ -2,55 +2,67 @@
 <!DOCTYPE locale SYSTEM "../../../../../lib/pkp/dtd/locale.dtd">
 
 <!--
-  * @file plugins/importexport/medra/locale/pt_PT/common.xml
+  * plugins/importexport/datacite/locale/pt_PT/common.xml
```
mtub: this is the correct file location...I will make a note to change this in the other files of the datacite plugin

plugins/importexport/datacite/locale/pt_PT/common.xml
```diff
@@ -2,55 +2,67 @@
...
   *
   * Copyright (c) 2013-2016 Simon Fraser University Library
   * Copyright (c) 2003-2016 John Willinsky
   * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
   *
-  * Shared localization strings
+  * Shared localization strings.
```
mtub: I don't mind adding this to the comment, but I wonder why you introduced this and maybe we will overwrite this in the future when updating headers.

mtub:
I can cherry-pick this, but feel free to go ahead, this might be the faster road...

asmecher:
Cherry-picked to `ojs-stable-2_4_8`, thanks!

samueloph:
Thanks @mtub and @asmecher, no need to say sorry, you're the ones helping me :)

PR #1016, opened 2016-09-15
Implement CLI for import/export plugins

bozana:
S. https://github.com/pkp/pkp-lib/issues/1601


bozana:
@asmecher, do you maybe know why the tests are failing?

bozana:
Here the PR: https://github.com/pkp/ojs/pull/1016
@asmecher, could you take a look?
Then still the CLI implementation for users import/export plugin is left.

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -302,6 +302,54 @@ function getUnregisteredGalleys($context) {
...
+		$issueDao = DAORegistry::getDAO('IssueDAO');
+		foreach ($issueIds as $issueId) {
+			$publishedIssue = $issueDao->getById($issueId, $context->getId());
+			if ($publishedIssue && $publishedIssue->getStoredPubId('doi')) $publishedIssues[] = $publishedIssue;
+		}
+		return $publishedIssues;
+	}
+
+	/**
+	 * Get article galleys with a DOI asigned  from gallley IDs.
```
asmecher: *assigned from galley

classes/plugins/DOIPubIdExportPlugin.inc.php
```diff
@@ -302,6 +302,54 @@ function getUnregisteredGalleys($context) {
...
+	 * @param $galleyIds array
+	 * @param $context Context
+	 * @return array
+	 */
+	function getArticleGalleys($galleyIds, $context) {
```
asmecher: Maybe this should be renamed, since it's specifically getting galleys with DOIs?

asmecher:
Looks good -- but it needs a rebase. Please go ahead and merge/cherry-pick when you're ready, thanks!

classes/plugins/DOIPubIdExportPlugin.inc.php
bozana: @asmecher, this function overwrites a parent function. The parent function (in PubObjectsExportPlugin i.e. for export of published objects e.g. in DOAJ) returns all article galleys and this one (in DOIPubIdExportPlugin i.e. for DOI export plugins) returns only galleys with DOI. They are named the same because they are called at the same place in the same way, so that I don't have to differentiate there anything else. OK so?

PR #1020, opened 2016-09-20
pkp/pkp/lib#1709 allow articles import and consider default locales in native import

bozana:
S. c) and e) here https://github.com/pkp/pkp-lib/issues/1709


plugins/importexport/native/NativeImportExportPlugin.inc.php
```diff
@@ -204,11 +214,12 @@ function exportIssues($issueIds, $context, $user) {
 	 * @param $importXml string XML contents to import
 	 * @param $context Context
 	 * @param $user User
+	 * @param $filter string Filter to be used
```
asmecher: This is now a Filter object, not a string.

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
```diff
@@ -338,6 +347,82 @@ function _getDateIssueSetterMappings() {
...
+				} else if ($index > 0) {
+					// the current title matches, but the prev titles didn't
+					fatalError(__('plugins.importexport.native.import.error.sectionTitleMatch', array('sectionTitle' => $title, 'issueTitle' => $issue->getIssueIdentification())));
+				}
+				$foundSectionId = $sectionId;
+				$foundSectionTitle = $title;
+			} else {
+				if ($foundSectionId) {
+					// a prev title matched, but the current doesn't
+					fatalError(__('plugins.importexport.native.import.error.sectionTitleMatch', array('sectionTitle' => $foundSectionTitle, 'issueTitle' => $issue->getIssueIdentification())));
```
asmecher: These `fatalError` calls will need changing when we alter the import/export tools to clean up partially-imported data upon error, but let's change that later.

bozana: Yes, I've first started to work on the other kind of error messages that will be displayed to the user... then I will clean up the partially-imported data...

plugins/importexport/native/NativeImportExportPlugin.inc.php
bozana: Is this not something like "native-xml=>issue" i.e. a string?

asmecher: Ah, yes, you're right -- I misread the calling code as fetching the filter object and passing it in. Never mind!

PR #1021, opened 2016-09-23
pkp/pkp-lib#1709 implement issue_identification element for native im…

bozana:
…port/export

s. d) here: https://github.com/pkp/pkp-lib/issues/1709

### DIFF
--- a/classes/issue/IssueDAO.inc.php
+++ b/classes/issue/IssueDAO.inc.php
@@ -174,6 +174,51 @@ class IssueDAO extends DAO implements PKPPubIdPluginDAO {
 	}
 
 	/**
+	 * Retrieve Issues by identification
+	 * @param $journalId int
+	 * @param $volume int
+	 * @param $number string
+	 * @param $year int
+	 * @param $titles array
+	 * @return DAOResultFactory
+	 */
+	function getIssuesByIdentification($journalId, $volume = null, $number = null, $year = null, $titles = array()) {
+		$params = array((int) $journalId);
+
+		$sqlVol = $sqlNum = $sqlYear = '';
+		if ($volume !== null) {
+			$sqlVol = ' AND i.volume = ?';
+			$params[] = (int) $volume;
+		}
+		if ($number !== null) {
+			$sqlNum = ' AND i.number = ?';
+			$params[] = $number;
+		}
+		if ($year !== null) {
+			$sqlYear = ' AND i.year = ?';
+			$params[] = (int) $year;
+		}
+
+		$i = 1;
+		$sqlTitleJoin = $sqlTitleWhere = '';
+		foreach ($titles as $title) {
+			$sqlTitleJoin .= ' LEFT JOIN issue_settings iss' .$i .' ON (i.issue_id = iss' .$i .'.issue_id)';
+			$sqlTitleWhere .= ' AND iss' .$i .'.setting_name = \'title\' AND iss' .$i .'.setting_value = ?';
+			$params[] = $title;
+			$i++;
+		}
+
+		$sql = 'SELECT i.* FROM issues i'
+			.$sqlTitleJoin .
+			' WHERE i.journal_id = ?'
+			.$sqlVol .$sqlNum .$sqlYear
+			.$sqlTitleWhere;
+
+		$result = $this->retrieve($sql, $params);
+		return new DAOResultFactory($result, $this, '_returnIssueFromRow');
+	}
+
+	/**
 	 * Retrieve Issue by "best" issue id -- public ID if it exists,
 	 * falling back on the internal issue ID otherwise.
 	 * @param $issueId string
--- a/classes/issue/IssueDAO.inc.php
+++ b/classes/issue/IssueDAO.inc.php
@@ -183,38 +183,36 @@ class IssueDAO extends DAO implements PKPPubIdPluginDAO {
 	 * @return DAOResultFactory
 	 */
 	function getIssuesByIdentification($journalId, $volume = null, $number = null, $year = null, $titles = array()) {
-		$params = array((int) $journalId);
-
-		$sqlVol = $sqlNum = $sqlYear = '';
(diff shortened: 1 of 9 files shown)


classes/issue/IssueDAO.inc.php
```diff
@@ -174,6 +174,51 @@ function getPublishedIssuesByNumber($journalId, $volume = null, $number = null,
...
+		}
+		if ($year !== null) {
+			$sqlYear = ' AND i.year = ?';
+			$params[] = (int) $year;
+		}
+
+		$i = 1;
+		$sqlTitleJoin = $sqlTitleWhere = '';
+		foreach ($titles as $title) {
+			$sqlTitleJoin .= ' LEFT JOIN issue_settings iss' .$i .' ON (i.issue_id = iss' .$i .'.issue_id)';
```
asmecher: If you use a `JOIN` rather than a `LEFT JOIN`, you can put all the conditions into the `ON (...)` clause. `LEFT JOIN` is only needed when you want the match to be "optional".

classes/issue/IssueDAO.inc.php
```diff
@@ -174,6 +174,51 @@ function getPublishedIssuesByNumber($journalId, $volume = null, $number = null,
...
+			$sqlTitleWhere .= ' AND iss' .$i .'.setting_name = \'title\' AND iss' .$i .'.setting_value = ?';
+			$params[] = $title;
+			$i++;
+		}
+
+		$sql = 'SELECT i.* FROM issues i'
+			.$sqlTitleJoin .
+			' WHERE i.journal_id = ?'
+			.$sqlVol .$sqlNum .$sqlYear
```
asmecher: Purely stylistic, but I prefer to keep all the SQL together in one string concatenation -- see e.g. `UserGroupDAO::getContextUsersCount`. It makes the query in all its permutations more readable IMO (though it does mean you need to track the order of parameters as they're added to the array).

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
```diff
@@ -288,6 +281,39 @@ function addSections($doc, $issueNode, $issue) {
...
 		$issueNode->appendChild($sectionsNode);
 	}
+
+	/**
+	 * Create and return an issue identification node.
+	 * @param $doc DOMDocument
+	 * @param $issue Issue
+	 * @return DOMElement
+	 */
+	function createIssueIdentificationNode($doc, $issue) {
```
asmecher: It looks like this is duplicated in two places... Could it be reused?

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
```diff
@@ -321,6 +317,40 @@ function parseIssueCover($node, $issue) {
 		}
 	}
 
+	/**
+	 * Parse out the issue identification and store it in an issue.
+	 * @param $node DOMElement
+	 * @param $issue Issue
+	 */
+	function parseIssueIdentification($node, $issue) {
```
asmecher: Likewise, could this be reused?

asmecher:
Are any updates needed to `sample.xml` and `tests/functional/*.xml` to suit the XML changes?

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
bozana: Yes, do you maybe have an idea where to put this function then? It is used in two different filters, but logically it doesn't belong to their parent NativeImportExportFilter class, it is not so generic/common. Hmmm... :-)

plugins/importexport/native/filter/NativeXmlArticleFilter.inc.php
```diff
@@ -145,11 +148,55 @@ function parseArticleGalley($n, $submission) {
...
+						list($locale, $value) = $this->parseLocalizedContent($n);
+						if (empty($locale)) $locale = $context->getPrimaryLocale();
+						$titles[$locale] = $value;
+						break;
+				}
+			}
+		}
+		$issueDao = DAORegistry::getDAO('IssueDAO');
+		$issuesByIdentification = $issueDao->getIssuesByIdentification($context->getId(), $vol, $num, $year, $titles);
+		assert($issuesByIdentification->getCount() == 1);
```
bozana: Ah, yes, you are right -- I started to mix assertions and fatalErrors :-P

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
bozana: These two are different, one assigns the values directly to an issue, the other uses them to search after the issue in the DB. But I could eventually just extract the part that parse the xml and then return an array of all the values (volume, number, titles) if this would be better/cleaner solution...

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
asmecher: Hmm, I'm thinking a Helper class is probably the least worst way.

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
asmecher: Ah, OK, don't worry!

bozana:
@asmecher, I did the changes according to your comments, s. the last commit above...

asmecher:
I've restarted both failed tests -- I'd like at least one of the two to finish before I merge. Looks good, though!

asmecher:
@bozana, the Lens test failed:
https://travis-ci.org/pkp/ojs/jobs/163440974
...but I'm presuming that's because the submodule hash hasn't been updated since your latest PR was merged. I'm going to go ahead and merge this to master, and update submodule hashes as needed.

asmecher:
Tests passed, thanks!

PR #1031, opened 2016-09-28
Resurrect the OJS CLI tool for merging users

ptrckb:
The script 'mergeUsers.php' was relying on a removed function from pkp-lib (pkp/pkp-lib@3c09b12)
and its description was implying a limitation to OJS v2.


asmecher:
Thanks, @ptrckb, I'll also cherry-pick this over to other branches as needed!

PR #1033, opened 2016-09-28
pkp/pkp-lib#1840 metrics migration

bozana:
S. https://github.com/pkp/pkp-lib/issues/1840

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -480,9 +480,9 @@ class Upgrade extends Installer {
 		$metricsDao->purgeLoadBatch($loadId);
 
 		$fileTypeCounts = array(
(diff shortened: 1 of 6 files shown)


asmecher:
Phew, that looks a lot cleaner.

I think there's still a problem with undefined constants in `migrateCounterPluginUsageStatistics`. For example, `USAGE_STATS_REPORT_PLUGIN_FILE_TYPE_HTML` -- this is used in `Upgrade` but never defined in the 3.0 codebase.

bozana:
@asmecher, I think I fixed those undefined constants, could you please take a look? It would be good to find an older installation that could test this upgrade...

asmecher:
That looks good, thanks -- can you port over to ojs-stable-3_0_0 too? I'll work with @jmacgreg on upgrade testing.

PR #1034, opened 2016-09-30
pkp/pkp-lib#1838 Automatic emails for open access/subscription trigge…

mfelczak:
…red post-install


asmecher:
Thanks, @mfelczak -- this will also need the queries added to the DB upgrade script.

asmecher:
Thanks, @mfelczak -- one more addition :) The queries should also go into `dbscripts/xml/upgrade/3.0.0_update.xml` for the sake of people using `ojs-stable-3_0_0`. When I merge I'll cherry-pick forward to master as well.

mfelczak:
Thanks @asmecher, I've added `dbscripts/xml/upgrade/3.0.0_update.xml` to the PR

mfelczak:
@asmecher, also removed the Acron crontab setting from the DB which includes the extra open access + subscription tasks. The crontab setting will then be automatically generated with correct tasks post install/upgrade.

mfelczak:
Nevermind, I see that the crontab setting is always regenerated post install/upgrade, no need to remove the setting from the DB.

PR #1039, opened 2016-10-03
Adding reading tools for OJS3 Arabic locale

vormia:
Retrieving reading tools for Arabic language after OJS 3.0.0.0 release.


mtub:
Thank you again, @vormia

PR #1040, opened 2016-10-04
Retrieving plugins translation in Arabic

vormia:
These are missing Arabic translation files for all plugins in OJS 3.0.0.0


mtub:
@vormia , can you have a quick look at the two issues I mentioned? If you don't want to change them, that's fine, too, just let me know.

plugins/generic/backup/locale/ar_IQ/locale.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE locale SYSTEM "../../../../../lib/pkp/dtd/locale.dtd">
+
+<!--
+  * plugins/generic/backup/locale.xml
```
mtub: No worries, I was just pointing out that the path indicated in the comment is wrong (was wrong in the English version, too). No problem.

PR #1041, opened 2016-10-05
pkp/pkp-lib#1709 implement issue_identification element for native import/export

bozana:
This is the cherry-pick of the issue_identification changes for native import/export to ojs-stable-3_0_0. S. d) here: https://github.com/pkp/pkp-lib/issues/1709
@asmecher, could you please just double check if the reference to the submodule lensGalley is correct for this branch?


asmecher:
Merged and submodules updated, thanks!

PR #1051, opened 2016-10-11
Ojs stable 2 4 6

pitercito:
new Upgrade


asmecher:
Hi @pitercito! It looks like this pull request might've gone against the wrong branch -- it contains lots of things that don't look like they're supposed to be there. Can you double-check and try again?

PR #1053, opened 2016-10-12
changing application name from ojs2 to ojs3 #1839

kaschioudi:
cf, https://github.com/pkp/pkp-lib/issues/1839

### DIFF
--- a/classes/core/Application.inc.php
+++ b/classes/core/Application.inc.php
@@ -62,7 +62,7 @@ class Application extends PKPApplication {
 	 * @return string
 	 */
 	function getName() {
-		return 'ojs2';
+		return 'ojs3';
 	}
 
 	/**


asmecher:
@kaschioudi, have you checked to see what the impact of this change is on already-existing installations of OJS (which will refer to `ojs2` in the database)?

kaschioudi:
you're right. it actually brakes existing ojs3 installs. I get the error below.  
`Fatal error:  Call to a member function getVersionString() on a non-object in /www/pkp/ojs3/lib/pkp/classes/template/PKPTemplateManager.inc.php on line 230`

asmecher:
I'm not sure it'll be possible to update this in the upgrade scripts, because those will want to check what version you're using. It might be best just to leave this as is -- it's a back-end issue and not displayed to the user anywhere.

kaschioudi:
Alright. Please disregard my pull request.

asmecher:
OK! It'd be nice to update this, but not a big deal. I'll see if I can think of something.

PR #1054, opened 2016-10-12
pkp/pkp-lib#1709 consider article pages in native import/export

bozana:
S. j) here: https://github.com/pkp/pkp-lib/issues/1709


asmecher:
This looks fine, thanks -- does it need back-porting to `ojs-stable-3_0_0`? I suppose so.

PR #1055, opened 2016-10-12
pkp/pkp-lib#1709 provide export XML for download

bozana:
S. i) here: https://github.com/pkp/pkp-lib/issues/1709
(This PR considers also other export plugins)

### DIFF
--- a/classes/plugins/PubObjectsExportPlugin.inc.php
+++ b/classes/plugins/PubObjectsExportPlugin.inc.php
@@ -163,13 +163,12 @@ abstract class PubObjectsExportPlugin extends ImportExportPlugin {
 			assert($filter != null);
 			// Get the XML
 			$exportXml = $this->exportXML($objects, $filter, $context);
-			$exportFileName = $this->getExportFileName($objectsFileNamePart, $context);
-			file_put_contents($exportFileName, $exportXml);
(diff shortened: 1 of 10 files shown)


classes/plugins/PubObjectsExportPlugin.inc.php
```diff
@@ -163,8 +163,13 @@ function executeExportAction($request, $objects, $filter, $tab, $objectsFileName
 			assert($filter != null);
 			// Get the XML
 			$exportXml = $this->exportXML($objects, $filter, $context);
-			header('Content-type: application/xml');
-			echo $exportXml;
+			$exportFileName = $this->getExportFileName($objectsFileNamePart, $context);
+			file_put_contents($exportFileName, $exportXml);
```
asmecher: Some of this is starting to reinvent the TemporaryFileManager; I haven't worked with that class for a while, but I think it should be pretty quick to make use of it instead for management of file creation/deletion/etc.

bozana:
@asmecher, I did the change, to use TemporaryFileManager everywhere. Could you please take a look again (the last commit)? THANKS!!!

bozana:
@asmecher, I now changed the commit above to use just the normal FileManager...

asmecher:
Merged, thanks!

PR #1057, opened 2016-10-13
ability to reorder galley display order #1796

kaschioudi:
https://github.com/pkp/pkp-lib/issues/1796


bozana:
Thanks @kaschioudi, I can than cherry-pick this to master and see if this is also necessary for OMP...

PR #1064, opened 2016-10-16
Adding a reference to Arabic in registry

vormia:
I think this is the last addition to enable Arabic language in OJS 3.0.0.0


mtub:
superseded by #1065

PR #1066, opened 2016-10-18
pkp/pkp-lib#1894: Define specific Dataverse error message to prevent hijacking of NOTIFICATION_TYPE_ERROR

ctgraham:
Resolves pkp/pkp-lib#1894, but the use of the new constant is a TODO,

### DIFF
--- a/plugins/generic/dataverse/DataversePlugin.inc.php
+++ b/plugins/generic/dataverse/DataversePlugin.inc.php
@@ -41,6 +41,7 @@ define('NOTIFICATION_TYPE_DATAVERSE_FILE_DELETED',   NOTIFICATION_TYPE_PLUGIN_BA
 define('NOTIFICATION_TYPE_DATAVERSE_STUDY_DELETED',  NOTIFICATION_TYPE_PLUGIN_BASE + 0x0001005);
 define('NOTIFICATION_TYPE_DATAVERSE_STUDY_RELEASED', NOTIFICATION_TYPE_PLUGIN_BASE + 0x0001006);
 define('NOTIFICATION_TYPE_DATAVERSE_UNRELEASED',     NOTIFICATION_TYPE_PLUGIN_BASE + 0x0001007);
+define('NOTIFICATION_TYPE_DATAVERSE_ERROR',          NOTIFICATION_TYPE_PLUGIN_BASE + 0x0001008);
 
 class DataversePlugin extends GenericPlugin {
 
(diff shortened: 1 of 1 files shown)


bozana:
@ctgraham, shouldn't all the calls of NOTIFICATION_TYPE_ERROR here (when notifications are created) be replaced by the new NOTIFICATION_TYPE_DATAVERSE_ERROR? I see you inserted a TODO comment. Would you be able to check/evaluate it?

ctgraham:
I didn't evaluate these myself because I wasn't confident in knowing for sure when the Dataverse-specific message was appropriate.  I'll happily try a first run-through if you want to check my guesses.

The current message is "An error occured updating this submission's data in Dataverse."

Yes, should be replaced:
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L604
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L604
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L1592
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L1628

Maybe? Not really Dataverse API specific, but is Dataverse specific?
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L859
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L916
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L975
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L1033

No, this should probably be a different (more specific) message:
https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/plugins/generic/dataverse/DataversePlugin.inc.php#L1560

bozana:
Thanks a lot @ctgraham. But earlier the same message was used for all those cases, right? -- Thus, we could at least do the same, right? I will take a look how it could be improved i.e. better differentiated, but I do not use Dataverse plugin i.e. I am not so familiar with it and I cannot test it...

ctgraham:
Same limitation here.  Since we don't use it and I'm not familiar with it, I'm just guessing based on the code context.

But, yes, to retain the prior function, we could just change all of those references.

PR #1069, opened 2016-10-20
pkp/pkp-lib#1905: Add check in acron to (re)validate scheduled task immediately before run

ctgraham:
Comes closer to resolving pkp/pkp-lib#1905 by giving the semaphore only one method's worth of execution time to be out of sync.  It is still possible to have a duplicate run, but it needs to be outside of the same second (or on a database which doesn't support the "affected rows" query) and must have both processes executing somewhere between the new `$this->_getTasksToRun()` call and the `$taskDao->updateLastRunTime($className, time())` simultaneously.

### DIFF
--- a/plugins/generic/acron/AcronPlugin.inc.php
+++ b/plugins/generic/acron/AcronPlugin.inc.php
@@ -244,7 +244,11 @@ class AcronPlugin extends GenericPlugin {
 
 			// By updating the last run time as soon as feasible, we can minimize
 			// the race window. See bug #8737.
-			$updateResult = $taskDao->updateLastRunTime($className, time());
+			$tasksToRun = $this->_getTasksToRun();
+			$updateResult = 0;
+			if (in_array($task, $tasksToRun, true)) {
+				$updateResult = $taskDao->updateLastRunTime($className, time());
+			}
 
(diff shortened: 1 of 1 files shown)


plugins/generic/acron/AcronPlugin.inc.php
```diff
@@ -244,7 +244,11 @@ function shutdownFunction() {
 
 			// By updating the last run time as soon as feasible, we can minimize
 			// the race window. See bug #8737.
-			$updateResult = $taskDao->updateLastRunTime($className, time());
+			$tasksToRun = $this->_getTasksToRun();
```
bozana: Just to double check: it is not important here to also actualize ` $this->_tasksToRun` because it is shutdown i.e. the last function?

bozana: I mean should there be something like `$this->_tasksToRun = $tasksToRun` after this loop? -- because `$tasksToRun` might be different than `$this->_tasksToRun` at this point. I actually don't think so, but... wanted to ask however... :-)

ctgraham: I don't see that as needed, either.

PR #1070, opened 2016-10-21
pkp/pkp-lib#1910: use Crossref test API for status checks when appropriate

crism:
Well, my test submission still failed, but I think this fixes the status query, at least.


bozana:
Thanks a lot @crism!

PR #1115, opened 2016-11-24
port external feed plugin to ojs3 #2027

kaschioudi:
https://github.com/pkp/pkp-lib/issues/2027

@asmecher


plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -17,10 +17,12 @@
 
 class ExternalFeedBlockPlugin extends BlockPlugin {
 	/** @var string Name of parent plugin */
-	var $parentPluginName;
+	protected $parentPluginName;
+	protected $parentPlugin = null;
```
asmecher: Each member variable should have a `@var` description like the one above.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,367 +1,322 @@
...
+			$templateMgr =& TemplateManager::getManager();
+			$templateMgr->addStyleSheet('externalFeed', '/' . $this->getStyleSheetFile());
+
+			// Journal home page display
+			HookRegistry::register('TemplateManager::display', array($this, 'displayHomepage'));
+		}
+		return $success;
+	}
+	
+	public function getDisplayName() {
```
asmecher: Missing docs

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,367 +1,322 @@
...
+							return new JSONMessage(false);
+						}
+					}
+					
+				}
+				else {
+					$form->initData();
+				}
+				return new JSONMessage(true, $form->fetch($request));
+				break;
```
asmecher: `break;` after `return` is dead code. No big deal, of course!

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,367 +1,322 @@
...
+						$recentItems = 0;
+					}
+					
+					$entries[] = array(
+						'feedTitle' => $currentFeed->getLocalizedTitle(),
+						'items' => $feed->get_items(0, $recentItems),
+					);
+				}
+				
+				$templateManager =& $args[0];
```
asmecher: Since we're now requiring PHP5 (phew!) we're reducing dependency on references (to make the code more readable, and to reduce the chances of weird, hard-to-track reference-related bugs). In PHP5, objects are automatically treated by reference, so cases like this can be simplified to remove the reference `&`.

asmecher:
Is SimplePie available as a composer dependency or something else? We're trying to reduce the amount of 3rd-party code that's directly included in the repo.

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -17,10 +17,12 @@
...
 
-	function __construct($parentPluginName) {
+	public function __construct($parentPluginName, $plugin) {
```
asmecher: I think the `$parentPluginName` is only used to fetch the plugin, so if we're passing `$plugin` directly instead, it should be possible to get rid of `$parentPluginName`, I think?

plugins/generic/externalFeed/ExternalFeedSettingsForm.inc.php
```diff
@@ -1,130 +1,91 @@
...
 	 * Assign form data to user-submitted data.
 	 */
-	function readInputData() {
+	public function readInputData() {
 		$this->readUserVars(array('externalFeedStyleSheet'));
 	}
-
+	
 	/**
 	 * Display the form.
```
asmecher: Needs docs updated

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridCellProvider.inc.php
```diff
@@ -0,0 +1,63 @@
...
+	//
+	// Template methods from GridCellProvider
+	//
+	/**
+	* Get cell actions associated with this row/column combination
+	* @param $row GridRow
+	* @param $column GridColumn
+	* @return array an array of LinkAction instances
+	*/
+	function getCellActions($request, $row, $column, $position = GRID_ACTION_POSITION_DEFAULT) {
```
asmecher: I think you can remove this entire implementation, since it just falls back on the parent implementation.

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * @brief Handle external feeds grid requests.
+ */
+import('lib.pkp.classes.controllers.grid.GridHandler');
+import('plugins.generic.externalFeed.controllers.grid.ExternalFeedGridRow');
+import('plugins.generic.externalFeed.controllers.grid.ExternalFeedGridCellProvider');
+
+class ExternalFeedGridHandler extends GridHandler {
+	/** @var ExternalFeedPlugin The external feeds plugin */
+	static $plugin;
```
asmecher: Why is this static?

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
```diff
@@ -0,0 +1,240 @@
...
+	 * Display the grid's containing page.
+	 * @param $args array
+	 * @param $request PKPRequest
+	 */
+	function index($args, $request) {
+		$context = $request->getContext();
+		import('lib.pkp.classes.form.Form');
+		$form = new Form(self::$plugin->getTemplatePath() . '/externalFeeds.tpl');
+		$json = new JSONMessage(true, $form->fetch($request));
+		return $json->getString();
```
asmecher: OTOH it's now possible to return a JSONMessage directly from a handler function -- no need to convert to string.

plugins/generic/externalFeed/controllers/grid/form/ExternalFeedForm.inc.php
```diff
@@ -0,0 +1,147 @@
...
+ */
+
+import('lib.pkp.classes.form.Form');
+
+class ExternalFeedForm extends Form {
+	/** @var int Context (press / journal) ID */
+	protected $contextId;
+	
+	/** @var string Static page name */
+	protected $feedId;
```
asmecher: Is this a name/string, or a numeric ID?

plugins/generic/externalFeed/controllers/grid/form/ExternalFeedForm.inc.php
```diff
@@ -0,0 +1,147 @@
...
+		if ($this->getData('limitItems')) {
+			$this->addCheck(new FormValidator($this, 'recentItems', 'required', 'plugins.generic.externalFeed.settings.recentItemsRequired'));
+		}
+	}
+	
+	/**
+	 * @see Form::fetch
+	 */
+	function fetch($request) {
+		$templateMgr = TemplateManager::getManager();
```
asmecher: Best to provide `$request` as a parameter to `getManager`, when available.

plugins/generic/externalFeed/controllers/grid/form/ExternalFeedForm.inc.php
```diff
@@ -0,0 +1,147 @@
...
+		if ((int) $this->getData('recentItems') <= 0) $this->setData('recentItems', '');
+		
+		// If recent items is selected, check that we have a value
+		if ($this->getData('limitItems')) {
+			$this->addCheck(new FormValidator($this, 'recentItems', 'required', 'plugins.generic.externalFeed.settings.recentItemsRequired'));
+		}
+	}
+	
+	/**
+	 * @see Form::fetch
```
asmecher: Doxygen doesn't "import" function parameter descriptions when you use `@see`, only with `@copydoc`. So you could either change this to `@copydoc` (if appropriate), or document all parameters and return type (if any) and stick with `@see`.

plugins/generic/externalFeed/templates/editExternalFeedForm.tpl
```diff
@@ -0,0 +1,88 @@
...
+	{if $feedId}
+		<input type="hidden" name="feedId" value="{$feedId|escape}" />
+	{/if}
+	
+	{include file="controllers/notification/inPlaceNotification.tpl" notificationId="ExternalFeedFormNotification"}
+	
+	{fbvFormSection}
+		{fbvFormSection for="title " title="plugins.generic.externalFeed.form.title" description="plugins.generic.externalFeed.form.titleDescription"}
+			{assign var="formLocale" value=$formLocale|escape}
+			{fbvElement type="text" id="title" name=title[$formLocale] value=$title|escape}
```
asmecher: No need to `escape` the value when using `fbvElement` -- this could lead to double-escaping when special characters are included in the form element value.

plugins/generic/externalFeed/templates/editExternalFeedForm.tpl
```diff
@@ -0,0 +1,88 @@
...
+	{csrf}
+	
+	{if $feedId}
+		<input type="hidden" name="feedId" value="{$feedId|escape}" />
+	{/if}
+	
+	{include file="controllers/notification/inPlaceNotification.tpl" notificationId="ExternalFeedFormNotification"}
+	
+	{fbvFormSection}
+		{fbvFormSection for="title " title="plugins.generic.externalFeed.form.title" description="plugins.generic.externalFeed.form.titleDescription"}
```
asmecher: `"title "` should be `"title"` (space)

plugins/generic/externalFeed/templates/editExternalFeedForm.tpl
```diff
@@ -0,0 +1,88 @@
...
+		{/fbvFormSection}
+		
+		{fbvFormSection for="feedUrl " title="plugins.generic.externalFeed.form.feedUrl" description="plugins.generic.externalFeed.form.feedUrlDescription"}
+			{fbvElement type="text" id="feedUrl" value=$feedUrl|escape}
+		{/fbvFormSection}
+	{/fbvFormSection}
+	
+	{fbvFormSection list=true description="plugins.generic.externalFeed.form.display"}
+		
+		{if $displayHomepage}
```
asmecher: You can use the `compare` modifier to do this a little more tidily (see e.g. `templates/controllers/tab/settings/access/form/accessForm.tpl`).

plugins/generic/externalFeed/templates/editExternalFeedForm.tpl
```diff
@@ -0,0 +1,88 @@
...
+	
+	
+	{fbvFormSection}
+		{if $limitItems}
+			{assign var="limitItemsChecked" value=true}
+		{else}
+			{assign var="limitItemsChecked" value=false}
+		{/if}
+		<input type="checkbox" name="limitItems" id="limitItems" value="1" {if $limitItemsChecked} checked="checked" {/if}>
+		Limit to
```
asmecher: Hard-coded English text.

plugins/generic/externalFeed/templates/homepage.tpl
```diff
@@ -0,0 +1,50 @@
...
+* Copyright (c) 2003-2013 John Willinsky
+* Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+*
+* Template for external feed content on journal homepage
+*}
+
+<div id="externalFeedsHome">
+
+	{foreach from=$entries item=entry}
+		<h3>{$entry.feedTitle}</h3>
```
asmecher: I'm not sure whether we want to dump these out unescaped -- will these be in spec if they include HTML? You could either pass through `|escape` (if it's not allowed to contain HTML) or `|strip_unsafe_html` (if we want to permit limited HTML).

plugins/generic/externalFeed/templates/homepage.tpl
```diff
@@ -0,0 +1,50 @@
...
+				<td colspan="2" class="headseparator">&nbsp;</td>
+			</tr>
+			
+			{assign var="separator" value=''}
+			
+			{foreach from=$entry.items item=item}
+				{$separator}
+			 	<tr class="title">
+					<td colspan="2" class="title">
+						<h4>{$item->get_title()}</h4>
```
asmecher: Likewise, we need to consider escaping.

asmecher: (and elsewhere in this template)

plugins/generic/externalFeed/templates/homepage.tpl
```diff
@@ -0,0 +1,50 @@
...
+		
+		<table class="externalFeeds">
+			<tr>
+				<td colspan="2" class="headseparator">&nbsp;</td>
+			</tr>
+			
+			{assign var="separator" value=''}
+			
+			{foreach from=$entry.items item=item}
+				{$separator}
```
asmecher: You could get rid of `$separator` and instead use the Smarty `first` property:
http://www.smarty.net/docsv2/en/language.function.foreach.tpl#foreach.property.first

asmecher:
Sorry to give you so many little comments -- still trying to describe our coding style etc., and all the things we've changed since that code was last maintained! Let me know when you've had a chance to review them.

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
kaschioudi: this is static because the LoadComponentHandler callback sets the reference to ExternalFeedPlugin through a static function call. (cf: https://github.com/kaschioudi/ojs/blob/2027-fix/plugins/generic/externalFeed/ExternalFeedPlugin.inc.php#L244)

This implementation was based on staticPages plugin
(cf: https://github.com/pkp/staticPages/blob/59aa1106de6a31210f789d3f3dda0907395a0563/StaticPagesPlugin.inc.php#L140)

plugins/generic/externalFeed/templates/homepage.tpl
kaschioudi: Simple pie handles sanitization: (cf https://github.com/kaschioudi/ojs/blob/00b283cd82345e10514bb072443c00da4597fb4d/plugins/generic/externalFeed/simplepie/SimplePie.inc.php#L14820). 

Here's the default list of tags which are filtered out (https://github.com/kaschioudi/ojs/blob/00b283cd82345e10514bb072443c00da4597fb4d/plugins/generic/externalFeed/simplepie/SimplePie.inc.php#L711) 

However It's possible to specify additional tags in case that list is not complete.

asmecher: Ah, thanks!

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
asmecher: Ah, I see. This could probably use some cleanup but better to do it consistently with other plugins -- in other words, don't worry about it for now :)

plugins/generic/externalFeed/ExternalFeedSettingsForm.inc.php
```diff
@@ -1,50 +1,48 @@
...
 	 */
-	function __construct(&$plugin, $journalId) {
+	public function __construct(&$plugin, $journalId) {
 		$this->journalId = $journalId;
-		$this->plugin =& $plugin;
-
-		parent::__construct($plugin->getTemplatePath() . 'settingsForm.tpl');
-
+		$this->plugin = $plugin;
+
```
asmecher: This is just me being picky :) Remove whitespace on otherwise blank lines.

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
```diff
@@ -0,0 +1,239 @@
...
+		$feedId = $request->getUserVar('feedId');
+		$context = $request->getContext();
+		$this->setupTemplate($request);
+		// Create and present the edit form
+		import('plugins.generic.externalFeed.controllers.grid.form.ExternalFeedForm');
+		$externalFeedPlugin = self::$plugin;
+		$externalFeedForm = new ExternalFeedForm(self::$plugin, $context->getId(), $feedId);
+		$externalFeedForm->initData();
+		$json = new JSONMessage(true, $externalFeedForm->fetch($request));
+		return $json->getString();
```
asmecher: As elsewhere -- you can just return the `JSONMessage` directly.

plugins/generic/externalFeed/controllers/grid/form/ExternalFeedForm.inc.php
```diff
@@ -0,0 +1,148 @@
...
+			array(
+				'feedUrl',
+				'displayHomepage',
+				'displayBlock',
+				'limitItems',
+				'recentItems'
+			)
+		);
+		
+		$this->setData('title', Request::getUserVar('title')[AppLocale::getLocale()]);
```
asmecher: Better when possible to use an instance of `$request` rather than static calls.

plugins/generic/externalFeed/controllers/grid/form/ExternalFeedForm.inc.php
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+	
+	/**
+	 * Fetch the form.
+	 * @copydoc Form::fetch()
+	 */
+	function fetch($request) {
+		$templateMgr = TemplateManager::getManager($request);
+		$templateMgr->assign('feedId', $this->feedId);
```
asmecher: Just FYI, there's an array syntax for `TemplateManager::assign` that's a little tidier IMO:

    $templateMgr->assign(array(
        'a' => $a,
        ...
    ));

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -60,8 +60,7 @@ function getDescription() {
 	 * @return object
 	 */
 	function &getExternalFeedPlugin() {
```
asmecher: No need for this reference.

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -73,6 +72,14 @@ function getPluginPath() {
 		return $plugin->getPluginPath();
 	}
 
+	/**
+	 * Override the builtin to get the correct template path.
+	 * @return string
+	 */
+	function getTemplatePath() {
```
asmecher: @NateWr has recently added some changes that might require work here as part of https://github.com/pkp/pkp-lib/issues/2872.

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -89,7 +96,7 @@ function getContents(&$templateMgr, $request = null) {
 
 		$requestedPage = $request->getRequestedPage();
 		$externalFeedDao = DAORegistry::getDAO('ExternalFeedDAO');
-		$plugin->import('simplepie.SimplePie');
+		require_once('lib/pkp/lib/vendor/simplepie/simplepie/autoloader.php');
```
asmecher: Now that we've got autoloading in place, I don't think this is necessary any more.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,366 +1,328 @@
-<?php
```
asmecher: Did the line endings change format from UNIX to DOS or something here? I'm not sure why the entire file is rewritten.

plugins/generic/externalFeed/ExternalFeedSettingsForm.inc.php
```diff
@@ -1,49 +1,47 @@
...
- * Copyright (c) 2003-2016 John Willinsky
- * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
- *
- * @class ExternalFeedSettingsForm
- * @ingroup plugins_generic_externalFeed
- *
- * @brief Form for journal managers to modify External Feed plugin settings
- */
-
+* @file plugins/generic/externalFeed/ExternalFeedSettingsForm.inc.php
```
asmecher: Accidental spacing changes

plugins/generic/externalFeed/classes/ExternalFeed.inc.php
```diff
@@ -1,14 +1,14 @@
 <?php
 
 /**
- * @file plugins/generic/externalFeed/ExternalFeed.inc.php
+ * @file plugins/generic/externalFeed/classes/ExternalFeed.inc.php
  *
  * Copyright (c) 2014-2016 Simon Fraser University Library
```
asmecher: All copyright dates need to be updated to 2018.

asmecher: All "Simon Fraser University Library" copyrights need to be changed to "Simon Fraser University". I believe this change has already been made in the `master` branch, so a rebase ought to correct it.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -22,19 +22,21 @@ class ExternalFeedDAO extends DAO {
...
 	}
 
 	/**
 	 * Retrieve an ExternalFeed by ID.
 	 * @param $feedId int
+	 * @param $journalId int 
 	 * @return ExternalFeed
 	 */
-	function &getExternalFeed($feedId) {
+	function &getExternalFeed($feedId, $journalId) {
```
asmecher: References aren't necessary. Also, DAO getter conventions for fetching a single object should be `getById`, and the second parameter (`$journalId`) should be optional.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -65,7 +67,7 @@ function getExternalFeedJournalId($feedId) {
 	 */
 	function &_returnExternalFeedFromRow($row) {
```
asmecher: This reference isn't necessary.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -65,7 +67,7 @@ function getExternalFeedJournalId($feedId) {
...
 		$externalFeedPlugin =& PluginRegistry::getPlugin('generic', $this->parentPluginName);
```
asmecher: If it's possible to pass the plugin object around, rather than its name, that would be preferred. It's possible that there's a reason it's written this way -- I haven't looked in detail.

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridCellProvider.inc.php
```diff
@@ -0,0 +1,47 @@
+<?php
+/**
+ * @file controllers/grid/ExternalFeedGridCellProvider.inc.php
```
asmecher: Path needs correction.

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridCellProvider.inc.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import('lib.pkp.classes.controllers.grid.GridCellProvider');
+import('lib.pkp.classes.linkAction.request.RedirectAction');
+
+class ExternalFeedGridCellProvider extends GridCellProvider {
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
```
asmecher: Empty constructors can be removed.

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
```diff
@@ -0,0 +1,244 @@
...
+	function getDataElementSequence($row) {
+		return $row->getSequence();
+	}
+
+	/**
+	 * @copydoc GridHandler::setDataElementSequence()
+	 */
+	function setDataElementSequence($request, $rowId, $gridDataElement, $newSequence) {
+		$externalFeedId = $rowId;
+		$journal = $request->getJournal();
```
asmecher: I suspect this could be made OMP-compatible by changing a bit of OJS-specific terminology (e.g. `$journal = $request->getJournal`) into app-agnostic terminology (e.g. `$context = $request->getContext()`).

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridRow.inc.php
```diff
@@ -0,0 +1,69 @@
+<?php
+/**
+ * @file controllers/grid/ExternalFeedGridRow.inc.php
```
asmecher: Path needs correction (probably elsewhere too)

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridRow.inc.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @brief Handle custom blocks grid row requests.
+ */
+import('lib.pkp.classes.controllers.grid.GridRow');
+
+class ExternalFeedGridRow extends GridRow {
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
```
asmecher: Remove dead constructor (probably elsewhere too)

plugins/generic/externalFeed/controllers/grid/form/ExternalFeedForm.inc.php
```diff
@@ -0,0 +1,151 @@
...
+			array(
+				'feedUrl',
+				'displayHomepage',
+				'displayBlock',
+				'limitItems',
+				'recentItems'
+			)
+		);
+		
+		$request = PKPApplication::getRequest();
```
asmecher: If possible, avoid getting the request object this way, in favour of having it passed in from the calling code.

plugins/generic/externalFeed/controllers/grid/form/ExternalFeedForm.inc.php
```diff
@@ -0,0 +1,151 @@
...
+	
+		$externalFeedDao = DAORegistry::getDAO('ExternalFeedDAO');
+		$plugin->import('classes.ExternalFeed');
+	
+		if (isset($this->feedId)) {
+			$feed = $externalFeedDao->getExternalFeed($this->feedId, $this->contextId);
+		}
+	
+		if (!isset($feed)) {
+			$feed = new ExternalFeed();
```
asmecher: Use `$externalFeedDao->newDataObject` to construct `ExternalFeed` objects.

plugins/generic/externalFeed/simplepie/SimplePie.inc.php
```diff
@@ -1,15009 +0,0 @@
-<?php
-/**
- * SimplePie
```
asmecher: I forget if we discussed this elsewhere -- but is SimplePie available as a git submodule or (better yet) Composer dependency? We should avoid adding third-party code with straight-up `git add`.

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridCellProvider.inc.php
kaschioudi: What would be the correct path?

asmecher: It's fine -- I read it wrong.

kaschioudi:
@asmecher : all the tests seem to have failed due to an error while running `composer update`. any idea what happened?

asmecher:
It looks like you have at least 2 commits including `lib/pkp` submodule updates. There should be only one, and it needs to be the last commit. https://pkp.sfu.ca/wiki/index.php?title=Github_Documentation_for_PKP_Contributors#PKP_library_submodule_changes

kaschioudi:
@asmecher this is ready for review

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -36,50 +36,57 @@ function getHideManagement() {
...
+		return $this->parentPlugin;
 	}
 
 	/**
 	 * Override the builtin to get the correct plugin path.
 	 * @return string
 	 */
-	function getPluginPath() {
+	public function getPluginPath() {
 		$plugin =& $this->getExternalFeedPlugin();
```
asmecher: `getExternalFeedPlugin` no longer returns a reference, so you'll need to remove the reference here too.

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -36,50 +36,57 @@ function getHideManagement() {
...
+	}
+
 	/**
 	 * Get the HTML contents for this block.
 	 * @param $templateMgr object
 	 * @param $request PKPRequest
 	 * @return $string
 	 */
-	function getContents(&$templateMgr, $request = null) {
+	public function getContents(&$templateMgr, $request = null) {
```
asmecher: `BlockPlugin::getContents` doesn't use a reference on `$templateMgr`.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,375 +1,329 @@
...
-	}
-}
-
-?>
+<?php
+
+/**
+ * @file plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
+ *
+ * Copyright (c) 2013-2018 Simon Fraser University Library
```
asmecher: There are still some copyrights referring to Simon Fraser University Library, not Simon Fraser University.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,375 +1,329 @@
...
+		$success = parent::register($category, $path);
+
+		if (!Config::getVar('general', 'installed') || defined('RUNNING_UPGRADE')) return true;
+		if ($success && $this->getEnabled()) {
+			$this->import('classes.ExternalFeedDAO');
+			
+			HookRegistry::register('PluginRegistry::loadCategory', array($this, 'callbackLoadCategory'));
+			
+			$externalFeedDao = new ExternalFeedDAO($this->getName());
+			$returner =& DAORegistry::registerDAO('ExternalFeedDAO', $externalFeedDao);
```
asmecher: It doesn't look like `$returner` is used anywhere.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,375 +1,329 @@
...
+			
+			HookRegistry::register('Templates::Management::Settings::website', array($this, 'callbackShowWebsiteSettingsTabs'));
+			HookRegistry::register('LoadComponentHandler', array($this, 'setupGridHandler'));
+			
+			$templateMgr =& TemplateManager::getManager();
```
asmecher: References to objects are almost never required anymore, just FYI. Best to remove them unless they're required.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,375 +1,329 @@
...
+	/**
+	 * Get the display name of this plugin
+	 * 
+	 * @return string
+	 */
+	public function getDisplayName() {
+		return __('plugins.generic.externalFeed.displayName');
+	}
+
+	public function getDescription() {
```
asmecher: All functions should have code headers.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,375 +1,329 @@
...
+	 * Override the builtin to get the correct template path.
+	 * @return string
+	 */
+	public function getTemplatePath($inCore = false)
+	{
+		return $this->getTemplateResourceName() . ':templates/';
+	}
+
+	/**
+	 * @see Plugin::getActions()
```
asmecher: Better to use `@copydoc` instead of `@see`, unless you specify all the parameters/return type. `@see` just refers to the other function as a link; `@copydoc` brings in its documentation directly, incluidng parameters/return type.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,375 +1,329 @@
...
+	 */
+	function getActions($request, $verb) {
+		import('lib.pkp.classes.linkAction.request.AjaxModal');
+		import('lib.pkp.classes.linkAction.request.RedirectAction');
+
+		$router = $request->getRouter();
+		$dispatcher = $request->getDispatcher();
+
+		return array_merge(
```
asmecher: Not a big deal, as we're not totally consistent on this, but our PHP-land convention is to indent with single tabs (even where Crockford prefers double-tabs in Javascript).

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
```diff
@@ -1,375 +1,329 @@
...
+				parent::getActions($request, $verb)
+		);
+	}
+
+	/**
+	 * @see Plugin::manage()
+	 */
+	function manage($args, $request) {
+
+		$journal = $request->getJournal();
```
asmecher: Still needs review for OJS-specific language; see https://github.com/pkp/ojs/pull/1115#discussion_r161036124

plugins/generic/externalFeed/ExternalFeedSettingsForm.inc.php
```diff
@@ -17,33 +16,33 @@
...
+	/** @var $plugin object */
+	protected $plugin;
 
 	/**
 	 * Constructor
 	 * @param $plugin object
 	 * @param $journalId int
 	 */
-	function __construct(&$plugin, $journalId) {
+	public function __construct(&$plugin, $journalId) {
```
asmecher: Unnecessary reference.

plugins/generic/externalFeed/block.tpl
```diff
@@ -1,8 +1,8 @@
 {**
- * plugins/generic/externalFeed/block.tpl
+ * plugins/generic/externalFeed/templates/block.tpl
```
asmecher: It doesn't look like this file actually got moved into `templates`...?

plugins/generic/externalFeed/classes/ExternalFeed.inc.php
```diff
@@ -20,6 +20,10 @@
 
 class ExternalFeed extends DataObject {
 
+	public function __construct() {
```
asmecher: Is this empty constructor necessary? (https://github.com/pkp/ojs/pull/1115#discussion_r161035708)

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -22,19 +22,33 @@ class ExternalFeedDAO extends DAO {
...
 		$this->parentPluginName = $parentPluginName;
 		parent::__construct();
 	}
 
+	/**
+	 * Instantiate a new data object.
+	 * @return ExternalFeed
+	 */
+	function newDataObject() {
+		import('plugins.generic.externalFeed.classes.ExternalFeed');
```
asmecher: It's likely that we're not doing this consistently, but it's better to use `$plugin->import` rather than hard-coding the canonical path to the plugin.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -22,19 +22,33 @@ class ExternalFeedDAO extends DAO {
...
+		return new ExternalFeed();
+	}
+
 	/**
 	 * Retrieve an ExternalFeed by ID.
 	 * @param $feedId int
+	 * @param $journalId int
```
asmecher: Wrong variable name.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -92,7 +103,7 @@ function &_returnExternalFeedFromRow($row) {
 	 * @param $externalFeed ExternalFeed
 	 * @return int 
 	 */
-	function insertExternalFeed(&$externalFeed) {
+	public function insertExternalFeed(&$externalFeed) {
```
asmecher: Our CRUD function name conventions have changed. Recent code uses `insertObject`, `getById`, `updateObject`, and `deleteObject`. This reference is also unnecessary.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -199,7 +210,7 @@ function deleteExternalFeedById($feedId) {
 	 * Delete external_feed by journal ID.
 	 * @param $journalId int
 	 */
-	function deleteExternalFeedsByJournalId($journalId) {
+	public function deleteExternalFeedsByJournalId($journalId) {
```
asmecher: Under the modern (and app-agnostic) conventions, functions like this would be `deleteByContextId`.

asmecher:
Thanks, @kaschioudi -- this still needs some review and clean-up; I've posted some new comments but also some of the old ones need attention (e.g. https://github.com/pkp/ojs/pull/1115/files#r161036360).

plugins/generic/externalFeed/block.tpl
kaschioudi: this was moved to `templates` then moved back to root folder (where `getBlockTemplateFilename` expects it) but i forgot to modify the path again.

kaschioudi:
> but also some of the old ones need attention (e.g. https://github.com/pkp/ojs/pull/1115/files#r161036360).

@asmecher : I am a little bit confused. Regarding path correction, I thought you said it everything was fine (https://github.com/pkp/ojs/pull/1115/files#r161852272)

kaschioudi:
@asmecher these builds are failing due to some unrelated changes. any idea?

asmecher:
I fixed an XML typo in the OJS repo -- testing now, but I expect it'll pass.

kaschioudi:
alright. will a rebase trigger my tests again?

asmecher:
Yes, a rebase and force-push should do it.

asmecher:
>> but also some of the old ones need attention (e.g. https://github.com/pkp/ojs/pull/1115/files#r161036360).

> @asmecher : I am a little bit confused. Regarding path correction, I thought you said it everything was fine (https://github.com/pkp/ojs/pull/1115/files#r161852272)

The first review identified some paths that needed correction, and there were some left in the second review as well. I often don't note all the areas where clean-up is needed if I'd just be repeating myself (e.g. paths, indents, dead constructors, etc), so I'd suggest checking more generally in a PR for that kind of clean-up rather than just the lines mentioned in a review.

kaschioudi:
@asmecher : all tests have passed now.

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -16,18 +16,21 @@
 import('lib.pkp.classes.plugins.BlockPlugin');
 
 class ExternalFeedBlockPlugin extends BlockPlugin {
-	/** @var string Name of parent plugin */
-	var $parentPluginName;
+	/** @var ExternalFeedPlugin reference to external feed plugin */
+	protected $parentPlugin = null;
```
asmecher: Protected members should have a `_` prefix on their names.

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -16,18 +16,21 @@
...
 
-	function __construct($parentPluginName) {
-		$this->parentPluginName = $parentPluginName;
+	/**
+	 * Constructor
```
asmecher: Needs `@var` self-documentation

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -36,62 +39,67 @@ function getHideManagement() {
 	 * its category.
 	 * @return String name of plugin
 	 */
-	function getName() {
+	public function getName() {
```
asmecher: I don't think the `getName` function is required anymore (the parent classes should provide a good enough implementation).

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -36,62 +39,67 @@ function getHideManagement() {
...
 	 * Get the display name of this plugin.
 	 * @return String
 	 */
-	function getDisplayName() {
+	public function getDisplayName() {
 		return __('plugins.generic.externalFeed.block.displayName');
 	}
 
 	/**
 	 * Get a description of the plugin.
```
asmecher: Needs `@return` (or `@copydoc` to bring in parent description)

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
```diff
@@ -36,62 +39,67 @@ function getHideManagement() {
...
+		if (!$context) return '';
 
-		$journalId = $journal->getId();
-		$plugin =& $this->getExternalFeedPlugin();
+		$plugin = $this->getExternalFeedPlugin();
 		if (!$plugin->getEnabled()) return '';
 
 		$requestedPage = $request->getRequestedPage();
 		$externalFeedDao = DAORegistry::getDAO('ExternalFeedDAO');
-		$plugin->import('simplepie.SimplePie');
```
asmecher: I don't think anything else in our apps uses SimplePie, does it? If not, then it's best to leave the SimplePie dependency inside the plugin, by adding a `composer.json` dependency in the plugin's installation directory.

How we fetch the dependencies depends on where the plugin lives. There are two ways this is currently done:
1. For plugins built into the main codebase, there are lines in `tools/buildpkg.sh` to run composer as part of the build process.
2. For plugins that live in their own repos, it's up to the plugin builder to fetch composer dependencies when building a release package to upload to github.

Currently this plugin lives in the `ojs` repo, but especially since it can be compatible with OMP with app-specific lingo removed, it would be better to move it to its own repo instead. I'd follow the second process.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
asmecher: Did you look into this?

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -1,45 +1,59 @@
...
 	 * @return ExternalFeed
 	 */
-	function &getExternalFeed($feedId) {
+	public function getById($feedId, $contextId = null) {
+		$params = array((int) $feedId);
+		if ($contextId) $params[] = (int) $contextId;
+
 		$result = $this->retrieve(
-			'SELECT * FROM external_feeds WHERE feed_id = ?', $feedId
+			'SELECT * FROM external_feeds WHERE feed_id = ? ' . ($contextId?' AND journal_id = ?':''),
```
asmecher: Again, this still has some OJS-specific language when I think it could easily be made to work verbatim with OMP as well. Search and replace throughout (https://github.com/pkp/ojs/pull/1115#discussion_r162739683).

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -63,11 +77,8 @@ function getExternalFeedJournalId($feedId) {
 	 * @param $row array
 	 * @return ExternalFeed
 	 */
-	function &_returnExternalFeedFromRow($row) {
-		$externalFeedPlugin =& PluginRegistry::getPlugin('generic', $this->parentPluginName);
-		$externalFeedPlugin->import('ExternalFeed');
-
-		$externalFeed = new ExternalFeed();
+	public function _returnExternalFeedFromRow($row) {
```
asmecher: Our newer conventions use a standard `_fromRow` rather than a different-for-every-class `_returnXYZFromRow`.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -125,15 +136,15 @@ function insertExternalFeed(&$externalFeed) {
...
+	public function getLocaleFieldNames() {
 		return array('title');
 	}
 
 	/**
 	 * Update the localized fields for this object.
 	 * @param $externalFeed
 	 */
-	function updateLocaleFields(&$externalFeed) {
+	public function updateLocaleFields(&$externalFeed) {
```
asmecher: This still has an unnecessary reference. (Elsewhere too.)

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -176,7 +187,7 @@ function updateExternalFeed(&$externalFeed) {
 	 * @param $externalFeed ExternalFeed 
 	 * @return boolean
 	 */
-	function deleteExternalFeed($externalFeed) {
+	public function deleteExternalFeed($externalFeed) {
```
asmecher: This still uses the old conventions (https://github.com/pkp/ojs/pull/1115#discussion_r162740795).

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
```diff
@@ -226,7 +237,7 @@ function getExternalFeedsByJournalId($journalId, $rangeInfo = null) {
 	/**
 	 * Sequentially renumber external feeds in their sequence order.
```
asmecher: Needs self-documentation (probably also elsewhere)

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
```diff
@@ -0,0 +1,244 @@
...
+	/**
+	 * @copydoc GridHandler::setDataElementSequence()
+	 */
+	function setDataElementSequence($request, $rowId, $gridDataElement, $newSequence) {
+		$externalFeedId = $rowId;
+		$context = $request->getContext();
+		$externalFeedDao = DAORegistry::getDAO('ExternalFeedDAO');
+		
+		// Ensure externalFeed is valid and for this journal
+		if (($externalFeedId != null && $externalFeedDao->getExternalFeedJournalId($externalFeedId) == $context->getId())) {
```
asmecher: The check for the context ID matching the feed context ID is already done in ExternalFeedDAO::getById, so this is duplicated code.

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
```diff
@@ -0,0 +1,244 @@
...
+	 * @param $request PKPRequest
+	 * @return string Serialized JSON object
+	 */
+	function delete($args, $request) {
+		$feedId = $request->getUserVar('feedId');
+		$context = $request->getContext();
+		
+		$externalFeedDao = DAORegistry::getDAO('ExternalFeedDAO');
+		if ($externalFeedDao->getExternalFeedJournalId($feedId) == $context->getId()) {
+			$externalFeedDao->deleteExternalFeedById($feedId);
```
asmecher: By adding a `contextId` parameter to this function, we can make integrity checks inherent rather than checking that the context ID matches first and then deleting conditionally. There should never be a `delete` without a context check so it's good to build it in. With that done, and my other comment about code calling `deleteExternalFeedById` (which as I've also noted should be renamed to `deleteById`), it should be possible to remove `getExternalFeedJournalId` entirely.

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
kaschioudi: yes. but to be honest, I am not sure what's happening here. any hint?

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
kaschioudi: Meaning I add `composer.json` to the plugin folder and once the plugin is moved to its own repo I upload a release package (zip file) containing the vendor folder. right?

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
kaschioudi: So the table column should be `context_id` instead of `journal_id`.  right?

plugins/generic/externalFeed/ExternalFeedBlockPlugin.inc.php
kaschioudi: Regarding `simplepie` library import. Okay to use `require_once(dirname(__FILE__) . '/vendor/autoload.php');` just like 
https://github.com/pkp/ojs/blob/master/plugins/paymethod/paypal/PaypalPaymentPlugin.inc.php#L17?

asmecher: Yes, I think that's the best way.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
asmecher: Yes, that's right. We don't have a good toolset for upgrade scripts within plugins, so it's something to consider before the plugin is released -- maybe file this in the new repo for some more specific consideration?

plugins/generic/externalFeed/ExternalFeedPlugin.inc.php
asmecher: It looks like your editor converted the file from UNIX-like line endings to DOS line endings. You can use `dos2unix` to convert it back, but also make sure your editor is set up for UNIX endings so it doesn't happen again.

plugins/generic/externalFeed/classes/ExternalFeedDAO.inc.php
kaschioudi: https://github.com/kaschioudi/externalFeed/issues/1

plugins/generic/externalFeed/controllers/grid/ExternalFeedGridHandler.inc.php
kaschioudi: @asmecher : if my understanding is correct, `contextId` should be added to `deleteById` signature so we would have `function deleteById($feedId, $contextId)`, right? 

How about `ExternalFeedDAO::deleteObject` which relies on `deleteById`.  Should `deleteObject` also receive `contextId` as parameter? I am asking because I looked around and none of DAO::deleteObject seem to do so.

asmecher: That's right (with the addition of the optional second parameter) -- it should be

    function deleteById($feedId, $contextId = null)

The `deleteObject` function is OK not to specify the second parameter to `deleteById`, as these objects can only be instantiated safely, i.e. in ways where the context ID will be the appropriate value.

kaschioudi: Okay. thanks.

kaschioudi:
@asmecher : tests don't seem to be failing due to external feed plugin code. am i correct?

bozana:
@kaschioudi, I fixed that... could you maybe do a rebase?

kaschioudi:
thanks @bozana : I just did a rebase..

PR #1125, opened 2016-12-06
pkp/pkp-lib#2068: Change DOIExportPlugin assertion to !isempty() instead of passing "register" value directly.

ctgraham:
Resolves pkp/pkp-lib#2068.

### DIFF
--- a/plugins/importexport/datacite/classes/DOIExportPlugin.inc.php
+++ b/plugins/importexport/datacite/classes/DOIExportPlugin.inc.php
@@ -239,7 +239,7 @@ class DOIExportPlugin extends ImportExportPlugin {
 					);
 					break;
 				} else { // Register selected objects.
-					assert($request->getUserVar('register'));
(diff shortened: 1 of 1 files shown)


plugins/importexport/datacite/classes/DOIExportPlugin.inc.php
```diff
@@ -239,7 +239,7 @@ function process(&$request, &$journal) {
 					);
 					break;
 				} else { // Register selected objects.
-					assert($request->getUserVar('register'));
+					assert(!empty($request->getUserVar('register')));
```
asmecher: See http://php.net/manual/en/function.empty.php -- before PHP 5.5.0, empty() could only handle variables, not expressions.

ctgraham:
New push uses `!= false` instead of `!empty()`.

asmecher:
Merged and cherry-picked over to ojs-stable-2_4_8.

PR #1132, opened 2016-12-09
Fix error in RefWorks export (misplaced {if})

mittinatten:
My log was clogging up with smarty errors from the RefWorks citation plugin. Only real change was on line 31, the rest was just indentation to be able to locate the error.

### DIFF
--- a/plugins/citationFormats/refWorks/citation.tpl
+++ b/plugins/citationFormats/refWorks/citation.tpl
@@ -14,42 +14,17 @@
 	{url|assign:"articleUrl" page="article" op="view" path=$article->getBestArticleId()}
 {/if}
 <form class="refworks_citation_form" action="http://www.refworks.com/express/expressimport.asp?vendor=Public%20Knowledge%20Project&filter=BibTeX&encoding=65001" method="post" target="RefWorksMain">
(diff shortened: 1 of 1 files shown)


asmecher:
@mittinatten, I think this will change the output indentation considerably -- part of the reason the code was so entangled was to preserve decent-looking indentation in the output. I'm not sure that's important for RefWorks, however, as I don't have it and can't test.

mittinatten:
I don't have it either, the output I got for a few tests looked sane though (and was all on one line?). I wonder if it worked at all before, since there were syntax errors. I can revert back to the old indentation, now that I know where the error came from, and resubmit the pull request. But I won't be able to test if the output is compatible with RefWorks.

asmecher:
Thanks, @mittinatten -- I've asked around locally to see if someone has what we need to test it.

mittinatten:
Reverted back to the old indentation in the last commit

asmecher:
Thanks, @mittinatten -- I've merged and cherry-picked over to `master`. I've also got someone over here who can test with RefWorks, so will verify!

mittinatten:
Great!

PR #1146, opened 2016-12-15
Partial fix for pkp/pkp-lib#1924 

jprk:
New function `glob_ci_pattern` that provides us with a case-insensitive globbing pattern so that the upgrade does not complain about not finding "123-456-7-??.PDF" when the file is in fact "123-456-7-??.pdf". Tested on Linux only, I am afraid that it might not work on Windows machines, depends on `glob()` implementation on Windows ([this post](http://php.net/manual/en/function.glob.php#108440) suggests that even Windows PHP supports character classes, but I have not way to test it).

I do not know if this is the best way to deal with the error, but it seems to: When I have a manuscript submitted with `.pdf` extension, our OJS 2.4.5 will save it as `<aid>-<fid>-<ver>-SM.pdf`, while a submission ending with `.PDF` will be saved as `<aid>-<fid>-<ver>-SM.PDF`. The database update will change the numbering of the files, and it seems to force the file extensions to lowercase. Nevertheless, the possibility of uppercase or mixed case file extension is ignored by the upgrade script which assumes that all extensions are lowercase.

Sorry for the number of reverts ...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,9 @@ temp
 .buildpath
 .settings
 .DS_Store
+# Local changes
+/.idea*
+*.bak
+*.live
+*.log
+GeoLiteCity.dat
--- a/.gitignore
+++ b/.gitignore
@@ -7,9 +7,3 @@ temp
 .buildpath
 .settings
 .DS_Store
-# Local changes
-/.idea*
-*.bak
-*.live
-*.log
-GeoLiteCity.dat
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -879,11 +879,13 @@ class Upgrade extends Installer {
 				foreach ($submissionFiles as $submissionFile) {
 					$generatedFilename = $submissionFile->getServerFileName();
 					$basePath = $submissionFileManager->getBasePath() . '/';
+					// Some filesystems are case sensitive. Make sure that we look for all possible variants
+					// matching given extension, i.e. not only for 'pdf', but also for 'PDF', 'Pdf' etc.
 					$globPattern = $submissionFile->getSubmissionId() . '-' .
 						$submissionFile->getFileId() . '-' .
 						$submissionFile->getRevision() . '-' .
 						'??' .
-						'.' . strtolower_codesafe($submissionFile->getExtension());
+						'.' . glob_ci_pattern($submissionFile->getExtension());
 
 					$matchedResults = array_merge(
 						glob($basePath . '*/*/' . $globPattern),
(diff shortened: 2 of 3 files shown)


asmecher:
@bozana, would you mind taking a look at this?

bozana:
I haven't tested it, which I would like, but maybe first just to ask about another possibility to solve it i.e. what do you think about it:
Could we maybe first get all the files in an array with `glob`, before the foreach loop, something like this:
`$files= array_merge(glob($basePath . '*/*/'), glob($basePath . '*/'));`
and then in the foreach loop use `preg_grep` with option "i" (case insensitive), something like this:
`$matchedResults = preg_grep('/' . $globPattern . '/i', $files);`
I am not sure if this would be better -- I don't know about its performance -- but it would mean that we do not need those two new functions `strtoupper_codesafe` and `glob_ci_pattern` ?

What do you think @jprk and @asmecher ?

Also maybe a comment about another possibility:
If the pattern would be something like this:
`$globPattern = $submissionFile->getSubmissionId() . '-' . $submissionFile->getFileId() . '-' . $submissionFile->getRevision() . '-' . '??' . '.{' . strtolower_codesafe($submissionFile->getExtension()) . ',' . strtoupper_codesafe($submissionFile->getExtension()) . '}' ;`
we could match something like .pdf and .PDF but not something like .pDf (which maybe is never the case anyway?), thus it would save the new function `glob_ci_pattern` ?

asmecher:
@jprk, do you know how it is that your filename cases are getting changed? I'm not sure how this would happen on a case-sensitive system.

jprk:
I have no idea. The original system is a slightly modified OJS 2.4.6. I
have not tested the file upload code, but my assumption is that the file
uploader strips the filename from the original extension, replaces it with
the "encoded" OJS filename, and glues the original extension back. This
way, when uploading `AaAaAa.PdF` you will get `XX-YY-NNN.PdF`.

I am pretty sure that some of our authors would generate and upload a PDF
with uppercase filenames...

Jan

jprk:
@bozana: Using `preg_grep` will require compiling a new regular expression for every file. This could potentially slow things down significantly. Ayway, if you opt for it, I would suggest modifying the code to be benevolent with respect to upper/lowercase letters in filename extension. I have seen `.pdf`, `.PDF` and `.Pdf` quite frequently and as OJS seem to not sanitize the extension to lowercase, you would ran into problems sooner or later when someone uploads file with mixed-case extension.

jprk:
@asmecher, @bozana: Just out of curiosity, why did you choose to force the extension to lowercase in the update code when (in my case) you have the correct OJS file name stored in `article_files.file_name`? Is there any reason not to trust the database?

bozana:
Hmmm.... Yes, I think we could use `$fileManager->parseFileExtension($this->getOriginalFileName()` to get the extension. As far as I can see, the files in OJS 2.4.8 are saved with the same extension as the one from the original file, which can be everything, e.d. pDF. @asmecher, would you agree?

asmecher:
Yes, @bozana, I think that's a good solution.

bozana:
I believe this was then solved/merged witch another PR... Thus closing...

PR #1172, opened 2017-01-06
fix migration problem with multilingual review forms #1782

ajnyga:
Migrating from OJS2 to OJS3 causes multilingual review forms to break and show "Array" as values. 
Mentioned here: https://github.com/pkp/pkp-lib/issues/1782

Although the discussion there suggests that the original code should be working, I could not get it to work without this change.

This code worked both with a clean install of 2.4.8.1 and our production server with 45 journals.


ajnyga:
and sorry about the second commit, you can of course ignore that

asmecher:
With apologies for the delay -- yes, this is exactly the fix. Thanks for spotting it. Rather than merging this PR with the second commit, I'll cherry-pick in the first commit manually.

PR #1181, opened 2017-01-09
pkp/pkp-lib#2134 usage events for html and embedded files

bozana:
s. https://github.com/pkp/pkp-lib/issues/2134#issuecomment-271299975

### DIFF
--- a/plugins/generic/htmlArticleGalley/HtmlArticleGalleyPlugin.inc.php
+++ b/plugins/generic/htmlArticleGalley/HtmlArticleGalleyPlugin.inc.php
@@ -100,6 +100,8 @@ class HtmlArticleGalleyPlugin extends GenericPlugin {
 
 		if ($galley && $galley->getFileType() == 'text/html' && $galley->getFileId() == $fileId) {
 			echo $this->_getHtmlContents($request, $galley);
+			$returner = true;
+			HookRegistry::call('HtmlArticleGalleyPlugin::articleDownloadFinished', array(&$returner));
 			return true;
 		}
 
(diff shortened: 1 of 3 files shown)


plugins/generic/htmlArticleGalley/HtmlArticleGalleyPlugin.inc.php
```diff
@@ -100,6 +100,8 @@ function articleDownloadCallback($hookName, $args) {
 
 		if ($galley && $galley->getFileType() == 'text/html' && $galley->getFileId() == $fileId) {
 			echo $this->_getHtmlContents($request, $galley);
+			$returner = true;
+			HookRegistry::call('FileManager::downloadFileFinished', array(&$returner));
```
asmecher: Rather than calling the File Manager's hook here, I'd suggest calling a different (new) hook, and registering against it in UsageEventPlugin.

bozana: Ah, yes, sure! Why hasn't this crossed my mind at all!? Although it seems so logical... :-\

bozana:
@asmecher, I did the change -- I had to make it then more generic and also change something in pkp-lib. Could you take a look, also at the pkp-lib changes: https://github.com/pkp/pkp-lib/pull/2170
If OK so, I will cherry-pick it to master and also consider OMP...

asmecher:
Hmm, yes, it did get a little messier -- but I think it's probably necessary. Go ahead, thanks!

PR #1188, opened 2017-01-12
pkp/pkp-lib#2159 migrate missing stage id in edit_decisions

bozana:


dbscripts/xml/upgrade/3.0.0_update.xml
```diff
@@ -208,4 +208,7 @@
 		<query>DELETE FROM scheduled_tasks WHERE class_name = 'classes.tasks.OpenAccessNotification'</query>
 		<query>DELETE FROM scheduled_tasks WHERE class_name = 'classes.tasks.SubscriptionExpiryReminder'</query>
 	</sql>
+	<sql>
+		<query>UPDATE edit_decisions SET stage_id=3</query><!--(WORKFLOW_STAGE_ID_EXTERNAL_REVIEW -->
```
asmecher: We'll also have to consider upgrades from 3.0.0 to 3.0.1. Will `stage_id` be `0` in cases where OJS2 submissions got upgraded without that data? We could add a condition for that and update across the board.

PR #1192, opened 2017-01-13
Replace page counting with ranges

ctgraham:
Resolves pkp/pkp-lib#2076.

Deprecates `Article::getStartPage()` and `Article::getEndPage()` which were introduced for loose page counting.

Depends on https://github.com/pkp/pkp-lib/issues/2196

### DIFF
--- a/classes/article/Article.inc.php
+++ b/classes/article/Article.inc.php
@@ -228,19 +228,25 @@ class Article extends Submission {
 	}
 
 	/**
+	 * Deprecated - this was used for loosely counting pages.  Either use ranges as entered, or write an actual page counting function.
 	 * Get starting page of an article.
+	 * @see Submission::getPageArray()
 	 * @return int
 	 */
 	function getStartingPage() {
+		if (Config::getVar('debug', 'deprecation_warnings')) trigger_error('Deprecated call to Article::getStartingPage().');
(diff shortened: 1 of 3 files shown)


bozana:
@ctgraham, you could change the submodule link to the pkp/master because I already integrated your changes there... + temporarily fix the tests, if you do not want to wait...

ctgraham:
Re-pushed without overriding the pkp-lib submodule, with passing checks.

bozana:
I took a look at this changes and they look good to me. I am just not sure, if now is the right time to declare those functions as deprecated, because, if I see it correctly, they are still in use in a few other places, right?

ctgraham:
@bozana, can you point me at any places the `Article::getStartingPage()` and `Article::getEndingPage()` are in use within core?  I thought I caught them all.

bozana:
If I see it correctly (with grep):
-- plugins/citationFormats/bibtex/citation.tpl
-- plugins/generic/webFeed/templates/rss.tpl
-- plugins/generic/googleScholar/GoogleScholarPlugin.inc.php
in OJS master :-)

ctgraham:
Ah, apparently I already had my eye on some of this with https://github.com/pkp/pkp-lib/issues/2075.

I'll resolve the outliers there.

PR #1202, opened 2017-01-20
DOAJ CRUD API

bozana:
s. https://github.com/pkp/pkp-lib/issues/850

### DIFF
--- a/classes/plugins/PubObjectsExportPlugin.inc.php
+++ b/classes/plugins/PubObjectsExportPlugin.inc.php
@@ -26,6 +26,9 @@ define('EXPORT_ACTION_EXPORT', 'export');
 define('EXPORT_ACTION_MARKREGISTERED', 'markRegistered');
 define('EXPORT_ACTION_DEPOSIT', 'deposit');
 
+// Configuration errors.
+define('EXPORT_CONFIG_ERROR_SETTINGS', 0x02);
+
 
 abstract class PubObjectsExportPlugin extends ImportExportPlugin {
 	/** @var PubObjectCache */
@@ -104,6 +107,19 @@ abstract class PubObjectsExportPlugin extends ImportExportPlugin {
 		switch (array_shift($args)) {
 			case 'index':
 			case '':
+				// Check for configuration errors:
+				$configurationErrors = array();
+				// missing plugin settings
+				$form = $this->_instantiateSettingsForm($context);
+				foreach($form->getFormFields() as $fieldName => $fieldType) {
+					if ($form->isOptional($fieldName)) continue;
+					$pluginSetting = $this->getSetting($context->getId(), $fieldName);
+					if (empty($pluginSetting)) {
+						$configurationErrors[] = EXPORT_CONFIG_ERROR_SETTINGS;
+						break;
+					}
+				}
+
 				// Add link actions
 				$actions = $this->getExportActions($context);
 				$actionNames = array_intersect_key($this->getExportActionNames(), array_flip($actions));
@@ -116,6 +132,7 @@ abstract class PubObjectsExportPlugin extends ImportExportPlugin {
 				$templateMgr->assign(array(
 					'plugin' => $this,
 					'actionNames' => $actionNames,
+					'configurationErrors' => $configurationErrors,
 				));
 				break;
 			case 'exportSubmissions':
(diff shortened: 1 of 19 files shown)


classes/plugins/PubObjectsExportPlugin.inc.php
```diff
@@ -338,6 +409,35 @@ function getAdditionalFieldNames($hookName, $args) {
...
+		$taskFilesPath[] = $this->getPluginPath() . DIRECTORY_SEPARATOR . 'scheduledTasks.xml';
+		return false;
+	}
+
+	/**
+	 * Retrieve all unregistered articles.
+	 * @param $context Context
+	 * @return array
+	 */
+	function getUnregisteredArticles($context) {
```
asmecher: Generally speaking, it's better (I think) to use iterator objects rather than pulling everything out into an array. Since some of these lists are going to involve a lot of articles, using the iterator will instantiate them one at a time, where converting to an array will need them all to exist in memory at once.

plugins/importexport/doaj/DOAJInfoSender.inc.php
```diff
@@ -0,0 +1,122 @@
...
+		$result = $plugin->depositXML($objects, $journal, $exportJson);
+		if ($result !== true) {
+			$this->_addLogEntry($result);
+		}
+
+	}
+
+	/**
+	 * Add execution log entry
+	 * @param $result array
```
asmecher: `$errors` would be a better parameter name, I think

plugins/importexport/doaj/filter/DOAJJsonFilter.inc.php
```diff
@@ -0,0 +1,178 @@
...
+import('lib.pkp.plugins.importexport.native.filter.NativeImportExportFilter');
+
+
+class DOAJJsonFilter extends NativeImportExportFilter {
+	/**
+	 * Constructor
+	 * @param $filterGroup FilterGroup
+	 */
+	function __construct($filterGroup) {
+		$this->setDisplayName('DOAJ JSON export');
```
asmecher: Hard-coded English

plugins/importexport/doaj/filter/DOAJJsonFilter.inc.php
```diff
@@ -0,0 +1,178 @@
...
+			$yearMonth = explode('-', $publicationDate);
+			$article['bibjson']['year'] = $yearMonth[0];
+			$article['bibjson']['month'] = $yearMonth[1];
+			/** --- FirstPage / LastPage (from PubMed plugin)---
+			 * there is some ambiguity for online journals as to what
+			 * "page numbers" are; for example, some journals (eg. JMIR)
+			 * use the "e-location ID" as the "page numbers" in PubMed
+			 */
+			$pages = $pubObject->getPages();
+			if (preg_match("/([0-9]+)\s*-\s*([0-9]+)/i", $pages, $matches)) {
```
asmecher: Watch out for @ctgraham's recent changes to page number handling!

plugins/importexport/doaj/filter/DOAJJsonFilter.inc.php
```diff
@@ -0,0 +1,178 @@
...
+			$publisherRecordId = $pubObject->getId();
+			// documentType
+			$type = $pubObject->getType($pubObject->getLocale());
+			*/
+
+			$articles[] = $article;;
+
+		}
+
+		$json = json_encode($articles);
```
asmecher: Just a comment on style -- I would typically replace

    $a = some_thing();
    return $a;

....with the one-liner

    return some_thing();

plugins/importexport/doaj/filter/DOAJJsonFilter.inc.php
bozana: For some reason all Filters have hard coded names that are then saved in the DB. It seems to me that is OK so i.e. I am not sure if it makes sense to translate them i.e. to have them in the locale files? Else, maybe we should not store them in the DB? Where else is the name used?

asmecher: Hmm, I remember that now, thanks. I haven't looked into it in detail. Let's keep it consistent for now.

plugins/importexport/doaj/filter/DOAJJsonFilter.inc.php
bozana: Hmmm... usually yes, but the function here returns a reference, thus there would be this PHP Notice: Only variable references should be returned by reference ...

asmecher: Argh, yeah, the filter framework is one of the places we're still using references. It could use a review and fix for that, but that's out of scope here.

PR #1209, opened 2017-01-24
pkp/pkp-lib#850 remove the current quickSubmit plugin

bozana:
The new quickSubmit plugin is here: https://github.com/pkp/quickSubmit...


bozana:
@asmecher, this is the deletion of the quickSubmit folder in OJS. So OK?

asmecher:
Yes, thanks!

PR #1219, opened 2017-01-24
Add head elements from HTML Galley file

Vitaliy-1:
This line will allow to add html head elements from html galley files that
are uploaded by user on production stage. Actually this line was present in
OJS 3.0.0.
Suppose I am doing pull request right :)


asmecher:
@Vitaliy-1, but that variable isn't used anywhere -- is this part of a bigger change e.g. to embed HTML content rather than using an iframe to present it?

Vitaliy-1:
@asmecher , I forgot that this variable was removed in OJS3.0.1. We have different display.tpl files in htmlArticleGalley now.  `{$htmlGalleyContents} ` in my case inserted before my custom lines, that code menu in html galley. and then I append it with javascript. It can be done easier but I want to get html content first. 
Suppose you want to stay with an iframe?

asmecher:
@Vitaliy-1, we'll probably stay with an iframe for the moment, but will hopefully will add other options.

PR #1220, opened 2017-01-24
locale names checked and repaired 

mtub:
a little housekeeping, correcting and unifying the locale names in the actual locale files


asmecher:
Merged manually due to conflicts. Thanks!

PR #1222, opened 2017-01-25
Initial commit for Finnish locale

ajnyga:
This is the initial commit for a Finnish locale for OJS3. Includes the OJS translations and some plugins.

Tagging @mtub 

edit: pkp-lib in a while


mtub:
Thank you, @ajnyga - this is a very clean update. Glad to be able to include a new language in OJS 3.

PR #1228, opened 2017-01-26
Add migration script for comments to editor

ajnyga:
Initial commit for migrating comments to editor field to queries.

Still to consider: removing comments to editor field from the submissions table

Tagging @bozana 

See:
https://github.com/pkp/pkp-lib/issues/1873
https://github.com/pkp/pkp-lib/pull/2200

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -1559,6 +1559,52 @@ class Upgrade extends Installer {
 		$noteDao->update('DELETE FROM submission_comments WHERE comment_type=2'); // COMMENT_TYPE_EDITOR_DECISION
 		return true;
 	}
+	
+	/**
+	 * Convert comments to editors to queries.
+	 * @return boolean True indicates success.
+	 */
+	function convertCommentsToEditor() {
+		$commentsResult = $submissionDao->retrieve('SELECT article_id, user_id, comments_to_ed, date_submitted FROM articles_migration WHERE comments_to_ed IS NOT NULL AND comments_to_ed != ""');
+		
+		$queryDao = DAORegistry::getDAO('QueryDAO');
+		$noteDao = DAORegistry::getDAO('NoteDAO');
+		
+		while (!$commentsResult->EOF) {
+			$row = $commentsResult->getRowAssoc(false);
+			$commentsResult->MoveNext();
+
+			$comments_to_ed = PKPString::stripUnsafeHtml($row['comments_to_ed']);
+			
+			if ($comments_to_ed != ""){
+				$query = $queryDao->newDataObject();
+				$query->setAssocType(ASSOC_TYPE_SUBMISSION);
+				$query->setAssocId($row['article_id']);
+				$query->setStageId(WORKFLOW_STAGE_ID_SUBMISSION);
+				$query->setSequence(REALLY_BIG_NUMBER);
+				
+				$queryDao->insertObject($query);
+				$queryDao->resequence(ASSOC_TYPE_SUBMISSION, $row['article_id']);
+				$queryDao->insertParticipant($query->getId(), $row['user_id']);
+				
+				$queryId = $query->getId();
+
+				$note = $noteDao->newDataObject();
+				$note->setUserId($row['user_id']);
+				$note->setAssocType(ASSOC_TYPE_QUERY);
+				$note->setTitle('Comments to Editor');
+				$note->setContents($comments_to_ed);
+				$note->setDateCreated(strtotime($row['date_submitted']);
+				$note->setDateModified(strtotime($row['date_submitted']);			
+				$note->setAssocId($queryId);
+				$noteDao->insertObject($note);
+			}
+		}
+		$commentsResult->Close();
+
+		return true;
+	}	
+	
 
 	/**
 	 * Localize issue cover images.
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -1568,11 +1568,13 @@ class Upgrade extends Installer {
 		$submissionDao = DAORegistry::getDAO('SubmissionDAO');
 		$queryDao = DAORegistry::getDAO('QueryDAO');
 		$noteDao = DAORegistry::getDAO('NoteDAO');
+		$userGroupDao = DAORegistry::getDAO('UserGroupDAO');
+		
+		$managerUserGroup = $userGroupDao->getDefaultByRoleId($journal->getId(), ROLE_ID_MANAGER);
+		$managerUsers = $userGroupDao->getUsersById($managerUserGroup->getId(), $journal->getId());
+		$creatorUserId = $managerUsers->next()->getId();		
 
-		$commentsResult = $submissionDao->retrieve('SELECT submissions.submission_id, submissions.comments_to_ed, stage_assignments.user_id FROM submissions 
-LEFT JOIN stage_assignments ON stage_assignments.submission_id = submissions.submission_id
-LEFT JOIN user_groups ON user_groups.user_group_id = stage_assignments.user_group_id
-WHERE comments_to_ed IS NOT NULL AND comments_to_ed != \'\' AND user_groups.role_id = \'65536\' ORDER BY submissions.submission_id');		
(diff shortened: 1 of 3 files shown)


ajnyga:
There is a typo, I will do another commit and test this with a simple installation.

classes/install/Upgrade.inc.php
```diff
@@ -1577,6 +1582,10 @@ function convertCommentsToEditor() {
 			$comments_to_ed = PKPString::stripUnsafeHtml($row['comments_to_ed']);
 			
 			if ($comments_to_ed != ""){
+				
+				$user_id = $row['user_id'];
+				if (!$user_id) $user_id = 1;
```
bozana: Maybe to use here the same approach as for the file uploader migration:
```
$managerUserGroup = $userGroupDao->getDefaultByRoleId($journal->getId(), ROLE_ID_MANAGER);
$managerUsers = $userGroupDao->getUsersById($managerUserGroup->getId(), $journal->getId());
$creatorUserId = $managerUsers->next()->getId();
```
?

classes/install/Upgrade.inc.php
```diff
@@ -1585,17 +1594,17 @@ function convertCommentsToEditor() {
 				
 				$queryDao->insertObject($query);
 				$queryDao->resequence(ASSOC_TYPE_SUBMISSION, $row['article_id']);
```
bozana: should it now be "submission_id"?

classes/install/Upgrade.inc.php
```diff
@@ -1577,6 +1582,10 @@ function convertCommentsToEditor() {
 			$comments_to_ed = PKPString::stripUnsafeHtml($row['comments_to_ed']);
 			
 			if ($comments_to_ed != ""){
+				
+				$user_id = $row['user_id'];
+				if (!$user_id) $user_id = 1;
+				
 				$query = $queryDao->newDataObject();
 				$query->setAssocType(ASSOC_TYPE_SUBMISSION);
 				$query->setAssocId($row['article_id']);
```
bozana: should it now be "submission_id"?

classes/install/Upgrade.inc.php
ajnyga: ah, it should. That was one of the typos I was meaning to fix, but apparently did not succeed. I will commit in a while...

classes/install/Upgrade.inc.php
ajnyga: yes, of course

classes/install/Upgrade.inc.php
```diff
@@ -1585,17 +1594,17 @@ function convertCommentsToEditor() {
...
-				$queryDao->insertParticipant($query->getId(), $row['user_id']);
+				$queryDao->insertParticipant($query->getId(), $user_id);
 				
 				$queryId = $query->getId();
 
 				$note = $noteDao->newDataObject();
-				$note->setUserId($row['user_id']);
+				$note->setUserId($user_id);
 				$note->setAssocType(ASSOC_TYPE_QUERY);
 				$note->setTitle('Comments to Editor');
```
bozana: maybe to use here the translated key "submission.submit.coverNote" either in the submission locale or journal primary locale (I am not sure what would be better -- maybe submission locale?)?

ajnyga: Sure, I just saw that convertqueries had that hard coded as well, so I thought that there is some logic behind it.

classes/install/Upgrade.inc.php
```diff
@@ -1565,10 +1565,15 @@ function convertEditorDecisionNotes() {
...
-		
+		$submissionDao = DAORegistry::getDAO('SubmissionDAO');
 		$queryDao = DAORegistry::getDAO('QueryDAO');
 		$noteDao = DAORegistry::getDAO('NoteDAO');
+
+		$commentsResult = $submissionDao->retrieve('SELECT submissions.submission_id, submissions.comments_to_ed, stage_assignments.user_id FROM submissions 
+LEFT JOIN stage_assignments ON stage_assignments.submission_id = submissions.submission_id
+LEFT JOIN user_groups ON user_groups.user_group_id = stage_assignments.user_group_id
+WHERE comments_to_ed IS NOT NULL AND comments_to_ed != \'\' AND user_groups.role_id = \'65536\' ORDER BY submissions.submission_id');		
+
```
bozana: The SQL looks good :-)
You could just use the acronyms, like:
```
SELECT s.submission_id, s.comments_to_ed, sa.user_id FROM submissions s
LEFT JOIN stage_assignments sa ON sa.submission_id = s.submission_id
LEFT JOIN user_groups ug ON ug.user_group_id = sa.user_group_id
WHERE s.comments_to_ed IS NOT NULL AND s.comments_to_ed != '' AND ug.role_id = 65536 ORDER BY s.submission_id
```
\

bozana:
@ajnyga, thanks a lot! Just a few comments here... else everything fine as well :-)

classes/install/Upgrade.inc.php
ajnyga: actually to set the user_id to 1 or the manager is just a fallback. At least in our case we can get the correct user_id every time.

classes/install/Upgrade.inc.php
bozana: Hmmm... Maybe there is... :-\ @asmecher, any comment? :-)

classes/install/Upgrade.inc.php
bozana: Yes, I think this should also always be the case -- this stage_assignment is migrated correctly... Thus maybe you can also totally remove it...

ajnyga:
fixed the above, still waiting an answer for the localization issue. Also, check if I understood the part regarding this "if (!$user_id) $user_id = 1" correctly.

bozana:
@ajnyga, now I see that that part depends on the journal i.e. the journal ID. It makes it a little bit complicated -- you would have to move that code (that searches after a default journal manager) where the `if (!$user_id)` is. But, I think you can remove that part completely and assume that user_id is always there -- The user_id was always necessary in OJS 2.4.x and it was correctly migrated to the stage_assignments table. And in OJS 3.x the stage_assignments is correctly filled. Sorry :-(

ajnyga:
Ok, I will remove that check and assume that the sql always returns a user_id

ajnyga:
@bozana this has to be run in some other point I guess, because the stage_assignments table does not exist yet at that point?

So, I do not know, maybe leave the comments_to_ed field in the submissions table and just let the script just empty that field now. And maybe remove the field in a later upgrade? This way we could run this after the schema and we would have the stage_assignments table?

(and there is a type in the sql and the SubmissionDAO is not recognized, the other similar scripts use ArticleDAO, but these are easily fixed after the above)

bozana:
@ajnyga, oh, good catch!
Hmmm... Yes, I believe that would be the best way. Maybe you can then also leave the functions `getCommentsToEditor` and `setCommentsToEditor` in the Submission class.
For SubmissionDAO you can use `Application::getSubmissionDAO()`.
Today I have some time and could overtake this issue, if you want?

bozana:
@ajnyga, ah I've just realized that we cannot do it like that, because in the next release it will be the same situation for those upgrading from OJS 2.4.x or 3.0.0 or 3.0.1 Thus, I can take a look at this upgrade, ok?

bozana:
Ah, I know, we can copy submission_id and comments_to_ed to another temporary table, as it is always done, e.g. here: https://github.com/pkp/ojs/blob/master/dbscripts/xml/upgrade/3.0.0_preupdate.xml#L33 and then work with that table... I will try...

ajnyga:
are you doing the last changes here?

bozana:
yes...

ajnyga:
ok great :)

ajnyga:
@bozana, the quotes in the sql need to be escaped the correct Dao used, just that you get that in as well.

bozana:
@ajnyga, here are my new PR:
https://github.com/pkp/pkp-lib/pull/2252
https://github.com/pkp/ojs/pull/1237
There you can see the additional commits for further changes. 
I would like Alec to take a look at that changes too, because I removed commets_to_editor from import/export and I am not sure if this is correct.

bozana:
New PRs:
https://github.com/pkp/pkp-lib/pull/2252
https://github.com/pkp/ojs/pull/1237

bozana:
@asmecher -- the PRs above migrate the comments for the editor in OJS. Could you take a look? I will adopt it now for OMP as well...

asmecher:
I'm OK with removing that from the import/export tools, because we'll probably eventually add a more comprehensive tool for importing/exporting conversations.

PR #1234, opened 2017-01-30
Disable issuepubnotifications if pubmodenone

ajnyga:
The issue is explained here:
https://github.com/pkp/pkp-lib/issues/2250


ajnyga:
whoops, ignore this, I will do another one without the two other commits. I accidentally created a new branch befor switching to master...

PR #1237, opened 2017-01-31
migrate comments for the editor

bozana:
s. https://github.com/pkp/pkp-lib/issues/1873


dbscripts/xml/upgrade.xml
```diff
@@ -38,9 +38,13 @@
 		<data file="dbscripts/xml/upgrade/2.4.0_postUsageStatsMigration2.xml" condition="return $installer->tableExists('issue_galleys_stats_migration');" />
 		<data file="dbscripts/xml/upgrade/2.4.0_postCounterMigration.xml" condition="return $installer->tableExists('counter_monthly_log');" />
 		<data file="dbscripts/xml/upgrade/2.4.0_postTimedViewsMigration.xml" condition="return $installer->tableExists('timed_views_log');" />
+		<data file="dbscripts/xml/upgrade/3.0.0_preupdate_commentsToEditor.xml" condition="return $installer->columnExists('submissions', 'comments_to_ed');" />
```
asmecher: This will execute after the DB schema is sync'd, I think, so the column will already be removed by now, no?

asmecher: Ah, no, I see it's pre DB schema.

PR #1238, opened 2017-02-01
type: localization - Add missing strings to pt-pt translation.

nmiguelmoura:
Add several key / values to pt-pt locale.xml file, which didn't displayed correctly.	

modified:   locale/pt_PT/locale.xml


mtub:
Thank you!

PR #1242, opened 2017-02-01
pkp/pkp-lib#1873 comments_to_ed migration if no author is assigned

bozana:
s. https://github.com/pkp/pkp-lib/issues/1873
This PR considers the situation when no author is assigned to a submission with a comment for the editors -- which is possible in OJS 3.x -- in that situation the first JM found will be used.

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -1566,35 +1566,51 @@ class Upgrade extends Installer {
 	 */
 	function convertCommentsToEditor() {
 		$submissionDao = Application::getSubmissionDAO();
+		$stageAssignmetDao = DAORegistry::getDAO('StageAssignmentDAO');
 		$queryDao = DAORegistry::getDAO('QueryDAO');
 		$noteDao = DAORegistry::getDAO('NoteDAO');
+		$userGroupDao = DAORegistry::getDAO('UserGroupDAO');
+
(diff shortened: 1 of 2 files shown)


classes/install/Upgrade.inc.php
```diff
@@ -1566,35 +1566,51 @@ function convertEditorDecisionNotes() {
...
+		$stageAssignmetDao = DAORegistry::getDAO('StageAssignmentDAO');
 		$queryDao = DAORegistry::getDAO('QueryDAO');
 		$noteDao = DAORegistry::getDAO('NoteDAO');
+		$userGroupDao = DAORegistry::getDAO('UserGroupDAO');
+
+		import('lib.pkp.classes.security.Role'); // ROLE_ID_...
 
 		$commentsResult = $submissionDao->retrieve(
-			'SELECT s.submission_id, s.comments_to_ed, s.date_submitted, sa.user_id
+			'SELECT s.submission_id, s.comments_to_ed, s.date_submitted
```
asmecher: I think if you moved the `context_id` into this query as well, then you wouldn't need to instantiate the `Submission` object for these entries. That'll keep the number of DB queries in DAO functions required for this upgrade step to a minimum, which will mean we're less likely to hit an upgrade tangle later if we change schemas (since this upgrade step runs before the schema sync).

PR #1248, opened 2017-02-06
pkp/pkp-lib#1955 no XML validation option for the export plugins

bozana:
s. https://github.com/pkp/pkp-lib/issues/1955
this PR introduces the option not to validate XML when exporting... 
The native exports are not considered -- they have local schema and they are good to be validated in any case.
Also CLI exports are not considered for now, because it would mean change of the usage parameters and thus locales files.

### DIFF
--- a/classes/plugins/PubObjectsExportPlugin.inc.php
+++ b/classes/plugins/PubObjectsExportPlugin.inc.php
@@ -142,6 +142,7 @@ abstract class PubObjectsExportPlugin extends ImportExportPlugin {
 				$selectedIssues = (array) $request->getUserVar('selectedIssues');
 				$selectedRepresentations = (array) $request->getUserVar('selectedRepresentations');
 				$tab = (string) $request->getUserVar('tab');
+				$noValidation = $request->getUserVar('validation') ? false : true;
 
 				if (empty($selectedSubmissions) && empty($selectedIssues) && empty($selectedRepresentations)) {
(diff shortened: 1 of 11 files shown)


classes/plugins/PubObjectsExportPlugin.inc.php
```diff
@@ -335,14 +336,15 @@ function getExportActionNames() {
 	 * @param $context Context
 	 * @return string XML document.
 	 */
-	function exportXML($objects, $filter, $context) {
+	function exportXML($objects, $filter, $context, $noValidation = null) {
```
asmecher: New parameter needs documentation in code header.

classes/plugins/PubObjectsExportPlugin.inc.php
```diff
@@ -173,13 +174,13 @@ function display($args, $request) {
 	 * @param $tab string Tab to return to
 	 * @param $objectsFileNamePart string Export file name part for this kind of objects
 	 */
-	function executeExportAction($request, $objects, $filter, $tab, $objectsFileNamePart) {
+	function executeExportAction($request, $objects, $filter, $tab, $objectsFileNamePart, $noValidation = null) {
```
asmecher: Code documentation for new parameter (maybe elsewhere too)

asmecher:
This looks good aside from a couple minor comments -- please go ahead!

PR #1276, opened 2017-02-21
Updates in Arabic locale after OJS 3.0.2

vormia:
Some minor updates and additions to Arabic language files for OJS 3.0.2


mtub:
Thanks, @vormia !

PR #1277, opened 2017-02-22
pkp/pkp-lib#2072 Versioning for published articles

lilients:
See https://github.com/pkp/pkp-lib/issues/2072.


classes/article/ArticleDAO.inc.php
```diff
@@ -182,11 +183,14 @@ function updateObject($article) {
 		);
 
 		$this->updateLocaleFields($article);
+		$contextId = Request::getContext()->getId();
```
bozana: It is always better to try not to use Request statically -- whenever possible. So maybe to get the context id from the article object here?

classes/article/ArticleDAO.inc.php
```diff
@@ -182,11 +183,14 @@ function updateObject($article) {
...
+		$version = $article->getCurrentVersionId($contextId);
 
 		// update authors for this article
-		$authors = $article->getAuthors();
+		$authors = $article->getAuthors(false, $version);
 		for ($i=0, $count=count($authors); $i < $count; $i++) {
 			if ($authors[$i]->getId() > 0) {
+				$authors[$i]->setVersion($version);
```
bozana: Is it necessary to set the version here? -- because we search after the authors by the version (getAuthors (false, $version)?

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -121,20 +129,23 @@ function getGalleysBySetting($settingName, $settingValue, $articleId = null, $jo
...
 				' . ($contextId?'INNER JOIN submissions s ON (g.submission_id = s.submission_id) ':'') . '
-				LEFT JOIN submission_files sf ON (g.file_id = sf.file_id)
-				LEFT JOIN submission_files nsf ON (nsf.file_id = g.file_id AND nsf.revision > sf.revision)
+				LEFT JOIN submission_galley_files sfg ON (sfg.galley_id = g.galley_id)
+				LEFT JOIN submission_files sf ON (sfg.file_id = sf.file_id)
+				LEFT JOIN submission_files nsf ON (nsf.file_id = sfg.file_id AND nsf.revision > sf.revision)
 				WHERE g.submission_id = ?
 					AND nsf.file_id IS NULL
 					' . ($contextId?' AND s.context_id = ? ':'') . '
+					' . ($submissionRevision?' AND sfg.version = ? ':'') . '
```
bozana: I think the `. '` in the line before this line, as well as `'` in this line are not necessary?

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -235,27 +248,49 @@ function _fromRow($row) {
...
 				$galley->getLabel(),
 				$galley->getLocale(),
 				$galley->getSequence() == null ? $this->getNextGalleySequence($galley->getSubmissionId()) : $galley->getSequence(),
 				$galley->getRemoteURL(),
-				$galley->getFileId(),
 			)
 		);
 		$galley->setId($this->getInsertId());
 		$this->updateLocaleFields($galley);
+		$this->addFile($galley->getId(), $galley->getFileId(), $galley->getCurrentVersionId());
```
bozana: Should here `getCurrentVersionId` be used or the current galley (that we are inserting) version number?

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -235,27 +248,49 @@ function _fromRow($row) {
...
 
 		HookRegistry::call('ArticleGalleyDAO::insertNewGalley', array(&$galley, $galley->getId()));
 
 		return $galley->getId();
 	}
 
 	/**
+	* Connect galley version with file in submission_galley_files
+	* @param $galleyId ArticleGalleyId
```
bozana: The syntax for parameter is @param variable_name type explanation, thus here and below:
@param $galleyId integer ...
This ways it looks like ArticleGalleyId, FileId and SubmissionRevision are objects/classes

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -235,27 +248,49 @@ function _fromRow($row) {
...
+	* @param $fileId FileId
+	* @param $version SubmissionRevision
+	*/
+	function addFile($galleyId, $fileId=0, $version=1){
+		$this->update(
+			'INSERT INTO submission_galley_files
+				(galley_id, file_id, version)
+				VALUES
+				(?, ?, ?)
+			ON DUPLICATE KEY UPDATE
```
bozana: It seems that not all DB engines act in the same way on this command, e.g. see MySql documentation: "The effects are not identical for an InnoDB table where a is an auto-increment column. With an auto-increment column, an INSERT statement increases the auto-increment value but UPDATE does not.". Also, PostgreSQL seem to have different command. Thus it wold be better to differentiate it into insert and update, I believe...

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -307,7 +342,9 @@ function deleteById($galleyId, $articleId = null) {
 			$params
 		);
 		if ($this->getAffectedRows()) {
-			$this->update('DELETE FROM submission_galley_settings WHERE galley_id = ?', array((int) $galleyId));
+			$this->update('
+				DELETE FROM submission_galley_settings
+				WHERE galley_id = ? AND version = ?', array((int) $galleyId, $submissionRevision));
```
bozana: (int) cast for submission revision would be good too

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -493,23 +530,26 @@ function getExportable($contextId, $pubIdType = null, $title = null, $author = n
...
+				LEFT JOIN submission_galley_files sfg ON (sfg.galley_id = g.galley_id)
+				LEFT JOIN submission_files sf ON (sfg.file_id = sf.file_id)
+				LEFT JOIN submission_files nsf ON (nsf.file_id = sfg.file_id AND nsf.revision > sf.revision AND nsf.file_id IS NULL )
 				' . ($pubIdType != null?' LEFT JOIN submission_galley_settings gs ON (g.galley_id = gs.galley_id)':'')
 				. ($title != null?' LEFT JOIN submission_settings sst ON (s.submission_id = sst.submission_id)':'')
 				. ($author != null?' LEFT JOIN authors au ON (s.submission_id = au.submission_id)':'')
 				. ($pubIdSettingName != null?' LEFT JOIN submission_galley_settings gss ON (g.galley_id = gss.galley_id AND gss.setting_name = ?)':'') .'
 			WHERE
 				i.published = 1 AND s.context_id = ?
+				AND ss.setting_name = "datePublished" AND ss.submission_revision = 1
```
bozana: Always when possible single quotes `'` should be used

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -154,12 +154,15 @@ function getPublishedArticlesByJournalId($journalId = null, $rangeInfo = null, $
...
 			FROM	published_submissions ps
 				LEFT JOIN submissions s ON ps.submission_id = s.submission_id
+				LEFT JOIN submission_settings ss ON ss.submission_id = ps.submission_id
 				LEFT JOIN issues i ON ps.issue_id = i.issue_id
 				' . $this->getFetchJoins() . '
 			WHERE 	i.published = 1
 				' . ($journalId?'AND s.context_id = ?':'') . '
 				AND s.status <> ' . STATUS_DECLINED . '
-			ORDER BY ps.date_published '. ($reverse?'DESC':'ASC'),
+				AND ss.setting_name = "datePublished"
```
bozana: single quotes...

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -273,15 +279,18 @@ function getPublishedArticlesBySectionId($sectionId, $issueId) {
 	 */
 	function getPublishedArticleById($publishedArticleId) {
 		$result = $this->retrieve(
-			'SELECT * FROM published_submissions WHERE published_submission_id = ?', (int) $publishedArticleId
+			'SELECT ps.*, ss.setting_value FROM published_submissions ps
+				JOIN 	submissions_settings ss ON (ss.submission_id = ps.submission_id) 					WHERE 	ss.setting_name = "datePublished" AND
```
bozana: maybe to put the WHERE clause in the next line?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -221,6 +224,9 @@ function getPublishedArticlesInSections($issueId, $useCache = false) {
 					$publishedArticles[$currSectionId]['title'] = $publishedArticle->getSectionTitle();
 				}
 			}
+			// get metadata of the current version
+			$publishedArticle = $this->getPublishedArticleByArticleId($publishedArticle->getId(), null, false, $publishedArticle->getCurrentVersionId());
```
bozana: Hmmm... It looks like everywhere else in this class the submission revision 1 is used to get the published articles, using _fromRow without the parameters. For example in getPublishedArticlesByJournalId. Why then here only the current article version?
It seems that the concept is to get the first and then later to get the current version, or? If so maybe to do this here as well?
Maybe to take a look how those functions are used elsewhere? -- e.g. in the generic plugin WebFeedGatewayPlugin both functions getPublishedArticlesByJournalId or getPublishedArticlesInSections are used, but they would return different article versions, I suppose?
It would be even better if we could always immediately provide the latest article versions here, or?

classes/install/Upgrade.inc.php
```diff
@@ -1903,6 +1903,53 @@ function moveReviewerFiles() {
...
+		$result = $submissionDao->retrieve('SELECT submission_id, date_published FROM published_submissions');
+
+		while (!$result->EOF) {
+			$row = $result->GetRowAssoc(false);
+			$submissionId = $row['submission_id'];
+			$datePublished = $submissionDao->datetimeFromDB($row['date_published']);
+
+			$submission = $submissionDao->getById($submissionId);
+			$submission->setDatePublished($datePublished);
+			$submissionDao->updateLocaleFields($submission);
```
bozana: Maybe just to use the SQL insert statement instead of getting the submission objects and update them -- the upgrade scripts work with SQL statements, so it would be OK here to do so, I suppose... Maybe just only one SQL statement for everything (for the whole function) would also work? -- In that case we could put it in an XML file in the upgrade folder...

classes/install/Upgrade.inc.php
```diff
@@ -1903,6 +1903,53 @@ function moveReviewerFiles() {
...
+
+		while (!$result->EOF) {
+			$row = $result->GetRowAssoc(false);
+			$galleyId = $row['galley_id'];
+			$fileId = $row['file_id'];
+
+			if(isset($fileId) && $fileId != 0){
+				$galley = $galleyDao->getById($galleyId);
+				$galley->setFileId($fileId);
+				$galleyDao->updateObject($galley);
```
bozana: Here the same as above -- Maybe only one SQL statement would do the job...

classes/search/ArticleSearchDAO.inc.php
```diff
@@ -79,17 +79,21 @@ function getPhraseResults($journal, $phrase, $publishedFrom = null, $publishedTo
...
 				published_submissions ps,
+				submission_settings ss,
 				issues i,
 				submission_search_objects o NATURAL JOIN ' . $sqlFrom . '
 			WHERE
 				s.submission_id = o.submission_id AND
 				s.status = ' . STATUS_PUBLISHED . ' AND
 				ps.submission_id = s.submission_id AND
+				ss.submission_id = s.submission_id AND
+				ss.setting_name = "datePublished" AND
```
bozana: single quotes...

classes/article/ArticleGalleyDAO.inc.php
lilients: In this case using single quotes would break the sql statement because datePublished needs to be a string.

classes/article/PublishedArticleDAO.inc.php
lilients: In this case using single quotes would break the sql statement because datePublished needs to be a string.

classes/article/PublishedArticleDAO.inc.php
bozana: You can use the escape, I think like this: `\'`

classes/article/PublishedArticleDAO.inc.php
bozana: Witch escape `\'` ?

dbscripts/xml/upgrade.xml
```diff
@@ -131,6 +133,12 @@
 		<note file="docs/release-notes/README-3.0.1" />
 	</upgrade>
 
+	<upgrade minversion="3.0.0.0" maxversion="3.1.0.0">
```
bozana: I think the maxversion is 3.0.9.9 (before 3.1)

dbscripts/xml/upgrade.xml
```diff
@@ -116,6 +116,8 @@
 	<upgrade minversion="3.0.0.0" maxversion="3.0.0.9">
 		<code function="setFileUploader" />
 		<code function="setFileName" />
+		<code function="migrateDatesPublished" />
+		<data file="dbscripts/xml/upgrade/3.0.0_versioning.xml" />
```
bozana: I think you can name it 3.1.0_...

dbscripts/xml/upgrade.xml
```diff
@@ -131,6 +133,12 @@
 		<note file="docs/release-notes/README-3.0.1" />
 	</upgrade>
 
+	<upgrade minversion="3.0.0.0" maxversion="3.1.0.0">
+		<code function="migrateDatesPublished" />
+		<code function="migrateGalleyFiles" />
+		<data file="dbscripts/xml/upgrade/3.0.0_versioning.xml" />
```
bozana: Also here 3.1.0_...

dbscripts/xml/upgrade/3.0.0_versioning.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE data SYSTEM "../../../lib/pkp/dtd/xmlData.dtd">
+
+<!--
+  * 3.0.0_versioning.xml
+  *
+  * Copyright (c) 2014-2016 Simon Fraser University Library
+  * Copyright (c) 2003-2016 John Willinsky
```
bozana: year 2017 :-)

dbscripts/xml/upgrade/3.0.0_versioning.xml
```diff
@@ -0,0 +1,27 @@
...
+  * 3.0.0.0 versioning updates.
+  -->
+<data>
+	<sql>
+		<!-- submission_settings -->
+		<query>UPDATE submission_settings SET submission_revision = 1 WHERE submission_revision IS NULL OR submission_revision = 0</query>
+		<!-- authors -->
+		<query>UPDATE authors SET version = 1 WHERE version IS NULL OR version = 0</query>
+		<query>UPDATE author_settings SET version = 1 WHERE version IS NULL OR version = 0</query>
+		<query>ALTER TABLE authors DROP PRIMARY KEY, ADD PRIMARY KEY (author_id, version)</query>
```
bozana: Does this not happen automatically, when the keys are defined in the DB schema xml?

dbscripts/xml/upgrade/3.0.0_versioning.xml
```diff
@@ -0,0 +1,27 @@
...
+		<query>UPDATE submission_galley_settings SET version = 1 WHERE version IS NULL OR version = 0</query>
+		<query>UPDATE submission_galley_files SET version = 1 WHERE version IS NULL OR version = 0</query>
+		<query>ALTER TABLE submission_galleys DROP COLUMN file_id</query>
```
bozana: This should also happen automatically, when the appropriate xml DB schema is run...

classes/install/Upgrade.inc.php
bozana: If you need the information before the DB XML schemas are executed and thus the columns removed, you can create a temporary table in a preupdate script, s. for example commentsToEditor migration and dbscripts/xml/upgrade/3.0.0_preupdate_commentsToEditor.xml.

dbscripts/xml/upgrade/3.0.0_versioning.xml
bozana: S. also comment above how to solve the issue when you need information before the DB XML schemas are executed, so that you would not need this code...

pages/article/ArticleHandler.inc.php
```diff
@@ -17,18 +17,30 @@
 import('classes.handler.Handler');
 
 class ArticleHandler extends Handler {
-	/** journal associated with the request **/
+	/** @var journal object: journal associated with the request **/
```
bozana: instead "object" you can write the Class of the object i.e. "Journal" here

pages/article/ArticleHandler.inc.php
```diff
@@ -17,18 +17,30 @@
...
 
-	/** article associated with the request **/
+	/** @var article object: article associated with the request **/
 	var $article;
 
-	/** galley associated with the request **/
+	/** @var galley object: galley associated with the request **/
 	var $galley;
 
+	/** @var int: submission(article) revision ID associated with the request **/
```
bozana: variable name is missing

pages/article/ArticleHandler.inc.php
```diff
@@ -17,18 +17,30 @@
 import('classes.handler.Handler');
 
 class ArticleHandler extends Handler {
-	/** journal associated with the request **/
+	/** @var journal object: journal associated with the request **/
 	var $journal;
```
bozana: This variable is not used much. Can all the other journal variables then be replaced with this one? -- it seems to be a little bit inconsistent...

pages/article/ArticleHandler.inc.php
```diff
@@ -57,11 +69,24 @@ function initialize($request, $args) {
...
 
 		$issueDao = DAORegistry::getDAO('IssueDAO');
+		$articleDao = DAORegistry::getDAO('ArticleDAO');
+		$publishedArticleDao = DAORegistry::getDAO('PublishedArticleDAO');
+
+		// get all published previous article versions
+		$this->previousRevisions = $articleDao->getPublishedSubmissionRevisions($articleId);
+
+		// get the most resent article version
+		$this->latestSubmissionRevision = $articleDao->getLatestPublishedRevisionId($articleId, $this->journal->getId());
```
bozana: You can get the latest revision id from this->previousRevisions, I suppose

pages/article/ArticleHandler.inc.php
```diff
@@ -71,9 +96,35 @@ function initialize($request, $args) {
 			$article = $articleDao->getById((int) $articleId, $journal->getId(), true);
```
bozana: $journal variable does not exist here.

bozana: Hmmm... maybe to provide the submission revision here as parameter? 
This part of the code is executed when an editor previews an unpublished article, I believe. Maybe it is a good feature for previewing the unpublished versions?

dbscripts/xml/upgrade/3.0.0_versioning.xml
asmecher: And I've recently changed

> Copyright (c) xxxx-yyyy Simon Fraser University Library

...to...

> Copyright (c) xxxx-yyyy Simon Fraser University

...as SFU is the legal entity, not the SFU library.

pages/article/ArticleHandler.inc.php
```diff
@@ -57,11 +69,24 @@ function initialize($request, $args) {
 		$articleId = isset($args[0]) ? $args[0] : 0;
 		$galleyId = isset($args[1]) ? $args[1] : 0;
```
bozana: If function "version" is used the second argument is not the galleyId

pages/article/ArticleHandler.inc.php
```diff
@@ -57,11 +69,24 @@ function initialize($request, $args) {
...
-		$publishedArticleDao = DAORegistry::getDAO('PublishedArticleDAO');
-		$publishedArticle = $publishedArticleDao->getPublishedArticleByBestArticleId((int) $journal->getId(), $articleId, true);
+		$this->journal = $request->getContext();
 
 		$issueDao = DAORegistry::getDAO('IssueDAO');
+		$articleDao = DAORegistry::getDAO('ArticleDAO');
+		$publishedArticleDao = DAORegistry::getDAO('PublishedArticleDAO');
+
+		// get all published previous article versions
+		$this->previousRevisions = $articleDao->getPublishedSubmissionRevisions($articleId);
```
bozana: should we not get here just submission revisions instead of published submission revisions? -- because it could be that the editor previews a not published article and because you search the published articles a few lines below

classes/article/ArticleDAO.inc.php
asmecher: Better still, the Request object shouldn't be used in the DAOs and DataObjects at all -- that's kind of mixing the Controller with the Model. Isn't the `context_id` (`journal_id`) available as a column here already?

classes/article/PublishedArticleDAO.inc.php
asmecher: @lilients, standard SQL syntax is for strings to be quoted with `'` -- MySQL is OK with `"` as well, but I think PostgreSQL isn't.

templates/frontend/objects/galley_link.tpl
```diff
@@ -48,8 +48,21 @@
 	{/if}
 {/if}
 
+{* Revision? *}
+{if $isPreviousRevision}
+	{assign var=path value=$parentId|to_array:$submissionRevision:$galley->getBestGalleyId()}
+	{assign var=label value=$galley->getGalleyLabel()}
+{else}
+	{assign var=path value=$parentId|to_array:$galley->getBestGalleyId()}
+	{assign var=label value=$galley->getGalleyLabel()}
```
bozana: the label is always the same i.e. two same lines of code. thus maybe not to use a variable for that?

templates/frontend/objects/galley_link.tpl
```diff
@@ -48,8 +48,21 @@
...
+{/if}
+
 {* Don't be frightened. This is just a link *}
-<a class="obj_galley_link {$type}{if $restricted} restricted{/if}" href="{url page=$page op="view" path=$parentId|to_array:$galley->getBestGalleyId()}">
+{if $isPreviousRevision}
+	<a class="obj_galley_link {$type}{if $restricted} restricted{/if}" href="{url page=$page op="version" path=$path}">
+{else}
+	<a class="obj_galley_link {$type}{if $restricted} restricted{/if}" href="{url page=$page op="view" path=$path}">
+{/if}
```
bozana: maybe to set the operation as a variable above, so that here only one line of the code can be used?

templates/frontend/objects/article_details.tpl
```diff
@@ -64,6 +64,16 @@
  * @uses $ccLicenseBadge string An image and text with details about the license
  *}
 <article class="obj_article_details">
+
+	{* Notification if there is a more recent version of this article *}
+	{if $isPreviousRevision}
+		<div class='cmp_notification'>
+			{translate key="submission.versioning.linkToLatest"}
+			{assign var=newVersionLink value=$article->getBestArticleId($currentJournal)}
```
bozana: getBestArticleId does not need a parameter

templates/frontend/objects/article_details.tpl
```diff
@@ -227,6 +240,22 @@
...
+				{* Display article versions *}
+				{if $versioningEnabled && (@count($previousRevisions) > 1)}
+					{if $previousRevisions|@count > 0}
+						<div class="item versioning">
+							<div class="label">
+								{translate key="submission.versioning.versionHistory"}
+							</div>
+							<div class="value">
+								{foreach from=$previousRevisions item=submissionRevision}
+									<a href="{url op="version" path=$article->getBestArticleId($currentJournal)|to_array:$submissionRevision->getSubmissionRevision() escape=false}">{translate key="submission.versioning.version" version=$submissionRevision->getSubmissionRevision() date=$submissionRevision->getDatePublished()|date_format:$dateFormatShort}</a><br>
```
bozana: getBestArticleId does not have arguments/parameters

templates/frontend/objects/article_details.tpl
```diff
@@ -205,14 +215,17 @@
 			{/if}
 
 			{* Article Galleys *}
+			{assign var=hasRevisions value=false}
```
bozana: this variable seems to be not used

templates/frontend/objects/article_details.tpl
```diff
@@ -64,6 +64,16 @@
  * @uses $ccLicenseBadge string An image and text with details about the license
  *}
 <article class="obj_article_details">
+
+	{* Notification if there is a more recent version of this article *}
+	{if $isPreviousRevision}
+		<div class='cmp_notification'>
+			{translate key="submission.versioning.linkToLatest"}
+			{assign var=newVersionLink value=$article->getBestArticleId($currentJournal)}
+			<a href="{url op="view" path=$newVersionLink escape=false}">{translate key="submission.versioning.latest"}</a>.
```
bozana: why do you use escape=false, here and elsewhere?

templates/controllers/tab/workflow/production.tpl
```diff
@@ -11,9 +11,41 @@
...
 
+<script type="text/javascript">
+	// Attach the JS file tab handler.
+	$(function() {ldelim}
+		$('#submissionRevisions').pkpHandler(
+			'$.pkp.controllers.TabHandler',
+			{ldelim}
+				{assign var=versionIndex value=$latestSubmissionRevision-1}
+				selected: {$versionIndex},
+				disabled: [{$submissionRevision}]
```
bozana: what contains this variable submissionRevision here?

templates/controllers/tab/workflow/production.tpl
```diff
@@ -11,9 +11,41 @@
...
+			{rdelim}
+		);
+	{rdelim});
+</script>
 <div id="production">
-{include file="controllers/notification/inPlaceNotification.tpl" notificationId="productionNotification" requestOptions=$productionNotificationRequestOptions}
+	{if $versioningEnabled || $submissionRevisions|@count > 1}
```
bozana: maybe to align this if with else and closing if, for easier reading

templates/controllers/tab/workflow/production.tpl
```diff
@@ -11,9 +11,41 @@
...
+			{ldelim}
+				{assign var=versionIndex value=$latestSubmissionRevision-1}
+				selected: {$versionIndex},
+				disabled: [{$submissionRevision}]
+			{rdelim}
+		);
+	{rdelim});
+</script>
 <div id="production">
-{include file="controllers/notification/inPlaceNotification.tpl" notificationId="productionNotification" requestOptions=$productionNotificationRequestOptions}
```
bozana: what happened with notifications -- do we not need them any more?

pages/article/ArticleHandler.inc.php
```diff
@@ -71,9 +96,35 @@ function initialize($request, $args) {
...
+		$articleId = $args[0];
+		$this->submissionRevision = $args[1];
+		$galleyId = isset($args[2]) ? $args[2] : 0;
+		array_splice($args, 1, 1);
+
+		// check if submission revision exists
+		$submissionDao = Application::getSubmissionDAO();
+		$submission = $submissionDao->getById($articleId, null, false, $this->submissionRevision);
+
+		if($submission->getDatePublished()){
```
bozana: What if the submission revision ($submission) does not exist

pages/article/ArticleHandler.inc.php
```diff
@@ -71,9 +96,35 @@ function initialize($request, $args) {
...
+	 */
+	function version($args, $request){
+		$articleId = $args[0];
+		$this->submissionRevision = $args[1];
+		$galleyId = isset($args[2]) ? $args[2] : 0;
+		array_splice($args, 1, 1);
+
+		// check if submission revision exists
+		$submissionDao = Application::getSubmissionDAO();
+		$submission = $submissionDao->getById($articleId, null, false, $this->submissionRevision);
```
bozana: I think this should be already considered in initialize function, so that it is not double (here and there)
I also find the order in initialize slightly better: first look after the published article and then just article

pages/article/ArticleHandler.inc.php
```diff
@@ -17,18 +17,30 @@
...
+	/** @var issue object: issue associated with the request **/
 	var $issue;
 
-	/** article associated with the request **/
+	/** @var article object: article associated with the request **/
 	var $article;
 
-	/** galley associated with the request **/
+	/** @var galley object: galley associated with the request **/
 	var $galley;
```
bozana: Galley is always get anew, in every function, thus I am not sure we need this class variable

controllers/tab/workflow/VersioningTabHandler.inc.php
```diff
@@ -0,0 +1,51 @@
...
+	// Extended methods from Handler
+	//
+	/**
+	 * @copydoc PKPHandler::authorize()
+	 */
+	function authorize($request, &$args, $roleAssignments) {
+		$stageId = (int) $request->getUserVar('stageId'); // This is validated in WorkflowStageAccessPolicy.
+
+		import('lib.pkp.classes.security.authorization.WorkflowStageAccessPolicy');
+		$this->addPolicy(new WorkflowStageAccessPolicy($request, $args, $roleAssignments, 'submissionId', $stageId));
```
bozana: Should other roles that have access to the production stage be able to create new versions?

controllers/grid/articleGalleys/form/ArticleGalleyForm.inc.php
```diff
@@ -20,17 +20,22 @@ class ArticleGalleyForm extends Form {
...
+	/** @var the article version */
+	var $_submissionRevision = null;
+
 	/** @var ArticleGalley current galley */
 	var $_articleGalley = null;
 
 	/**
 	 * Constructor.
 	 * @param $submission Submission
+	 * @param $submissionRevision SubmissionRevision (optional)
```
bozana: variable type is int

controllers/grid/articleGalleys/form/ArticleGalleyForm.inc.php
```diff
@@ -20,17 +20,22 @@ class ArticleGalleyForm extends Form {
 	/** @var the article */
 	var $_submission = null;
 
+	/** @var the article version */
+	var $_submissionRevision = null;
```
bozana: it would be nice to correct the comments for the variables here: name type description

controllers/grid/articleGalleys/ArticleGalleyGridRow.inc.php
```diff
@@ -19,12 +19,16 @@ class ArticleGalleyGridRow extends GridRow {
 	/** @var Submission **/
 	var $_submission;
 
+	/** @var Submission revision */
+	var $_submissionRevision;
```
bozana: it would be nice to correct the variables here: name type description

controllers/grid/articleGalleys/ArticleGalleyGridRow.inc.php
```diff
@@ -99,12 +107,21 @@ function getSubmission() {
 	}
 
 	/**
+	 * Get the submission revision for this row (already authorized)
+	 * @return Submission revision
```
bozana: return value is an integer, correct?

controllers/grid/articleGalleys/ArticleGalleyGridCellProvider.inc.php
```diff
@@ -24,9 +24,10 @@ class ArticleGalleyGridCellProvider extends DataObjectGridCellProvider {
 	 * Constructor
 	 * @param $submission Submission
 	 */
-	function __construct($submission) {
+	function __construct($submission, $submissionRevision) {
 		parent::__construct();
 		$this->_submission = $submission;
+		$this->_submissionRevision = $submissionRevision;
```
bozana: Where is this class variable declared?

controllers/grid/articleGalleys/ArticleGalleyGridCellProvider.inc.php
```diff
@@ -59,6 +60,7 @@ function getTemplateVarsFromRowColumn($row, $column) {
 	function getRequestArgs($row) {
 		return array(
 			'submissionId' => $this->_submission->getId(),
+			'submissionRevision' => $this->getSubmissionRevision(),
```
bozana: Where is this function defined?

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
```diff
@@ -341,6 +344,21 @@ function execute($request) {
 				$submission->setLicenseURL(null);
 			}
 
+			// Versioning
+
+			// set publication date
+			$submission->setDatePublished($this->getData('datePublished') ? $this->getData('datePublished') : Core::getCurrentDate());
```
bozana: Because this is happening here, and not above as earlier i.e. because it does not considers if an article is unassigned from an issue, an article can have the datePublished although it is not assigned to an issue i.e. although not published

classes/article/ArticleDAO.inc.php
lilients: I can read the contextId from the article: `$article->getContextId()` ok?

bozana: I think you do not need that parameter at all -- if you see my other comments... Also, Alec proposed to have another DB column in the submissions table with the current/latest version, I believe.... So maybe to first go roughly over all comments and then start with changes... ?

classes/article/ArticleDAO.inc.php
lilients: yep, seems to be redundant. Removed it. 🐛

classes/article/ArticleGalleyDAO.inc.php
lilients: Yes, thanks. Its gone now.

classes/article/ArticleGalleyDAO.inc.php
lilients: Because you can only add galleys to the most recent version it is always the recent version id (Old versions can not be edited). Maybe this should be made clearer a the comments of the function ...

classes/article/ArticleGalleyDAO.inc.php
lilients: ok

pages/article/ArticleHandler.inc.php
lilients: True. Removed it.

pages/article/ArticleHandler.inc.php
lilients: I set 1 as default value.

pages/article/ArticleHandler.inc.php
lilients: Hmm, the problem is, that the version number can only be read at the version function. So the function at initialize can only get the default values (the latest version) but for the display of the version specific data you need to call getPublishedArticleByBestArticleId again at version().

pages/article/ArticleHandler.inc.php
bozana: I mean if the object $submission, that the function is being called on, does not exist -- it would lead to an PHP error... ?

classes/article/PublishedArticleDAO.inc.php
lilients: I changed `_fromRow()` to get the `latestRevisionId` if `submissionRevision==null`

controllers/tab/workflow/VersioningTabHandler.inc.php
lilients: I think, every role that has access should be able to create new versions. What do you think? Do I need to change something here to accomplish that?

controllers/grid/articleGalleys/ArticleGalleyGridCellProvider.inc.php
lilients: added the declaration

templates/controllers/tab/workflow/production.tpl
lilients: moved it to the part of the template that is beeing displayed if versioning is not enabled

bozana: can they not be included if versioning is there -- do they not make sense any more? at least in the first step coming from copyediting they might be good to have?

lilients: ok

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -286,17 +339,18 @@ function updateObject($galley) {
 	 * @param $galley ArticleGalley
 	 */
 	function deleteObject($galley) {
-		return $this->deleteById($galley->getId());
+		return $this->deleteById($galley->getId(), null, $galley->getCurrentVersionId());
```
bozana: maybe to provide the articleId ($galley->getSubmissionId()) as well -- is more secure...

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -307,7 +361,9 @@ function deleteById($galleyId, $articleId = null) {
 			$params
 		);
 		if ($this->getAffectedRows()) {
-			$this->update('DELETE FROM submission_galley_settings WHERE galley_id = ?', array((int) $galleyId));
+			$this->update('
+				DELETE FROM submission_galley_settings
+				WHERE galley_id = ? AND version = ?', array((int) $galleyId, (int)$submissionRevision));
```
bozana: should here (by deletion of a galley) also the new table submission_galley_files be considered?

classes/article/ArticleGalleyDAO.inc.php
```diff
@@ -493,23 +549,26 @@ function getExportable($contextId, $pubIdType = null, $title = null, $author = n
 			FROM	submission_galleys g
 				JOIN submissions s ON (s.submission_id = g.submission_id)
 				LEFT JOIN published_submissions ps ON (ps.submission_id = g.submission_id)
+				JOIN submission_settings ss ON (ss.submission_id = ps.submission_id)
 				JOIN issues i ON (ps.issue_id = i.issue_id)
-				LEFT JOIN submission_files sf ON (g.file_id = sf.file_id)
-				LEFT JOIN submission_files nsf ON (nsf.file_id = g.file_id AND nsf.revision > sf.revision AND nsf.file_id IS NULL )
+				LEFT JOIN submission_galley_files sfg ON (sfg.galley_id = g.galley_id)
+				LEFT JOIN submission_files sf ON (sfg.file_id = sf.file_id)
+				LEFT JOIN submission_files nsf ON (nsf.file_id = sfg.file_id AND nsf.revision > sf.revision AND nsf.file_id IS NULL )
```
bozana: hmmm... how can these two cases coexist: `nsf.file_id = sfg.file_id AND nsf.file_id IS NULL`?

classes/article/ArticleGalleyDAO.inc.php
bozana: maybe to add articleId ($galley->getSubmissionId()) as well -- it is more safe...

classes/article/ArticleGalleyDAO.inc.php
bozana: should also the new table submission_galley_files be considered when deleting a galley?

classes/article/ArticleGalleyDAO.inc.php
bozana: hmmm.... do these two conditions together make sense: `nsf.file_id = sfg.file_id AND nsf.file_id IS NULL`?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -125,6 +128,81 @@ function getPublishedArticles($issueId) {
...
+				LEFT JOIN submissions s ON ps.submission_id = s.submission_id
+			WHERE 	ss.submission_id = ?'.
+				($contextId ? ' AND s.context_id = ?' : '').'
+				AND ss.setting_name = \'datePublished\'
+			ORDER BY ss.submission_revision ' . $this->getDirectionMapping($order);
+
+		$result = $this->retrieve($sql, $params);
+		$submissionRevisions = array();
+
+		foreach($result as $submission) {
```
bozana: hmmm... $result is a ADORecordSet object and not array, so does this work?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -125,6 +128,81 @@ function getPublishedArticles($issueId) {
...
+		$params = array((int) $submissionId);
+		if ($contextId) {
+			$params[] = (int) $contextId;
+		}
+
+		$sql = 'SELECT MAX(ss.submission_revision)
+			FROM 	submission_settings ss
+				INNER JOIN submissions s ON ss.submission_id = s.submission_id
+			WHERE 	ss.submission_id = ?
+				AND ss.setting_name = \'datePublished\'
```
bozana: could it be that the setting_value is empty i.e. do we have to ensure that it is not? and in all other cases here...

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -125,6 +128,81 @@ function getPublishedArticles($issueId) {
...
+			$params[] = (int) $contextId;
+		}
+
+		$sql = 'SELECT DISTINCT ss.submission_revision
+			FROM 	submission_settings ss
+				LEFT JOIN published_submissions ps ON ps.submission_id = ss.submission_id
+				LEFT JOIN submissions s ON ps.submission_id = s.submission_id
+			WHERE 	ss.submission_id = ?'.
+				($contextId ? ' AND s.context_id = ?' : '').'
+				AND ss.setting_name = \'datePublished\'
```
bozana: could it be that the setting_value is empty i.e. do we have to ensure that it is not?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -105,10 +105,13 @@ function getPublishedArticles($issueId) {
 				' . $this->getFetchColumns() . '
 			FROM	published_submissions ps
 				LEFT JOIN submissions s ON ps.submission_id = s.submission_id
+				JOIN submission_settings ss ON (ss.submission_id = s.submission_id)
 				' . $this->getFetchJoins() . '
 				LEFT JOIN custom_section_orders o ON (s.section_id = o.section_id AND o.issue_id = ?)
 			WHERE	ps.submission_id = s.submission_id
 				AND ps.issue_id = ?
+				AND ss.setting_name = \'datePublished\'
```
bozana: could it be that the setting_value is empty i.e. do we have to ensure that it is not?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -154,12 +232,14 @@ function getPublishedArticlesByJournalId($journalId = null, $rangeInfo = null, $
 				' . $this->getFetchColumns() . '
 			FROM	published_submissions ps
 				LEFT JOIN submissions s ON ps.submission_id = s.submission_id
+				LEFT JOIN submission_settings ss ON ss.submission_id = ps.submission_id
```
bozana: shouldn't we then consider DISTINCT because we could get several submission_settings because of the revisions? Also elsewhere?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -154,12 +232,14 @@ function getPublishedArticlesByJournalId($journalId = null, $rangeInfo = null, $
...
 				LEFT JOIN issues i ON ps.issue_id = i.issue_id
 				' . $this->getFetchJoins() . '
 			WHERE 	i.published = 1
 				' . ($journalId?'AND s.context_id = ?':'') . '
 				AND s.status <> ' . STATUS_DECLINED . '
-			ORDER BY ps.date_published '. ($reverse?'DESC':'ASC'),
+				AND ss.setting_name = \'datePublished\'
```
bozana: should here also AND MAX(ss.submission_revision) not be checked?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -192,9 +272,11 @@ function getPublishedArticlesInSections($issueId, $useCache = false) {
 				' . $this->getFetchColumns() . '
 			FROM	published_submissions ps
 				JOIN submissions s ON (ps.submission_id = s.submission_id)
+				LEFT JOIN submission_settings ss ON (ss.submission_id = ps.submission_id)
 				' . $this->getFetchJoins() . '
 				LEFT JOIN custom_section_orders o ON (s.section_id = o.section_id AND ps.issue_id = o.issue_id)
 			WHERE	ps.issue_id = ?
+				AND ss.setting_name = \'datePublished\'
```
bozana: should here also AND MAX(ss.submission_revision) not be checked? And elsewhere?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -221,6 +303,10 @@ function getPublishedArticlesInSections($issueId, $useCache = false) {
 					$publishedArticles[$currSectionId]['title'] = $publishedArticle->getSectionTitle();
 				}
 			}
+
+			// get galleys of the current version
+			$publishedArticle->setGalleys($this->galleyDao->getBySubmissionId($row['submission_id'], null, $publishedArticle->getCurrentVersionId())->toArray());
```
bozana: do we not get this already above with _fromRow? -- if we look after the MAX(ss.submission_revision)?

classes/article/PublishedArticleDAO.inc.php
```diff
@@ -273,15 +361,17 @@ function getPublishedArticlesBySectionId($sectionId, $issueId) {
 	 */
 	function getPublishedArticleById($publishedArticleId) {
```
bozana: i think this function is not used anywhere, right? maybe to remove it then?

pages/article/ArticleHandler.inc.php
```diff
@@ -357,8 +422,9 @@ function cite($args, $request) {
 		$router = $request->getRouter();
 		$this->setupTemplate($request);
 		$articleId = isset($args[0]) ? $args[0] : 0;
-		$citeType = isset($args[1]) ? $args[1] : null;
-		$returnFormat = isset($args[2]) ? $args[2] : null;
+		$version = isset($args[1]) ? $args[1] : null;
+		$citeType = isset($args[2]) ? $args[2] : null;
```
bozana: this is probably OK, because these old URLs are not so important as the "download" URLs

templates/controllers/modals/submissionMetadata/form/issueEntrySubmissionReviewForm.tpl
```diff
@@ -48,6 +48,13 @@
 	<input type="hidden" name="displayedInContainer" value="{$formParams.displayedInContainer|escape}" />
 	<input type="hidden" name="tab" value="submission" />
 
+	{* Versioning *}
+	{assign var=submissionRevision value=$formParams.revision}
+	{if !$formParams.revision}
+		{assign var=submissionRevision value=$latestRevisionId}
+	{/if}
+	<input type="hidden" name="submissionRevision" id="submissionRevision" value="{$submissionRevision}" />
```
bozana: maybe to escape the submissionRevision

templates/controllers/tab/settings/versioning/form/versioningForm.tpl
```diff
@@ -0,0 +1,37 @@
+{**
+ * templates/controllers/tab/settings/versioning/form/versioningForm.tpl
+ *
+ * Copyright (c) 2014-2016 Simon Fraser University Library
```
bozana: year

templates/frontend/objects/article_details.tpl
```diff
@@ -227,6 +239,22 @@
 						{$article->getDatePublished()|date_format:$dateFormatShort}
 					</div>
 				</div>
+
+				{* Display article versions *}
+				{if $versioningEnabled && (@count($previousRevisions) > 1)}
+					{if $previousRevisions|@count > 0}
```
bozana: why double check of the count (> 1 and > 0) ? -- it seems that it is not necessary....

controllers/tab/workflow/VersioningTabHandler.inc.php
bozana: OK, lets leave it as it is for now :-)

classes/article/ArticleGalleyDAO.inc.php
lilients: same here https://github.com/pkp/ojs/blob/master/classes/article/ArticleGalleyDAO.inc.php#L498
should I change that, although it has nothing to do with versioning?

pages/article/ArticleHandler.inc.php
```diff
@@ -224,8 +285,12 @@ function downloadSuppFile($args, $request) {
 	 */
 	function download($args, $request) {
 		$articleId = isset($args[0]) ? $args[0] : 0;
-		$galleyId = isset($args[1]) ? $args[1] : 0;
-		$fileId = isset($args[2]) ? (int) $args[2] : 0;
+		$submissionRevision = isset($args[1]) ? (int) $args[1] : 0;
```
lilients: Yes, goot that you point it out. I did not see that. I removed the `$submissionRevision` parameter form download, so the old urls are still working. But this way the download of old versions is not working, because the redirect (https://github.com/pkp/ojs/blob/master/pages/article/ArticleHandler.inc.php#L171) calls the initialize function which overwrites the `$submissionRevision` and the `download()` function does only hand out the most recent submission revision. Do you know if the redirect is needed or could I replace it with a normal function call like `$this->download()`?

pages/article/ArticleHandler.inc.php
```diff
@@ -55,25 +67,61 @@ function authorize($request, &$args, $roleAssignments) {
...
+
+		// get data of publishedArticle
 		if (isset($publishedArticle)) {
 			$issue = $issueDao->getById($publishedArticle->getIssueId(), $publishedArticle->getJournalId(), true);
 			$this->issue = $issue;
 			$this->article = $publishedArticle;
 		} else {
 			$articleDao = DAORegistry::getDAO('ArticleDAO');
-			$article = $articleDao->getById((int) $articleId, $journal->getId(), true);
+			$article = $articleDao->getById((int) $articleId, $this->journal->getId(), true);
```
lilients: hm, I did not see a such function in ArticleDao. Also this function is only used for unpublished articles, so the articleId can only be an int, I guess. 🐰

pages/article/ArticleHandler.inc.php
```diff
@@ -55,25 +67,61 @@ function authorize($request, &$args, $roleAssignments) {
...
+		$this->previousRevisions = $publishedArticleDao->getPublishedSubmissionRevisions($publishedArticle->getId(), $this->journal->getId(), SORT_DIRECTION_DESC);
+
+		// get the most recent article version
+		$this->latestSubmissionRevision = $publishedArticleDao->getLatestRevisionId($publishedArticle->getId(), $this->journal->getId());
+
+		// set latest submission revision as default
+		$this->submissionRevision = $this->latestSubmissionRevision;
+
+		// get metadata of the latest version
+		$publishedArticle = $publishedArticleDao->getPublishedArticleByBestArticleId((int) $this->journal->getId(), $articleId, false, $this->latestSubmissionRevision);
```
lilients: yes, very good idea. I also changed that at getByBestGalleyId 👍

classes/article/PublishedArticleDAO.inc.php
lilients: changed it

classes/plugins/PubIdPlugin.inc.php
```diff
@@ -136,17 +136,21 @@ function getPubId($pubObject) {
...
 			assert(is_a($pubObject, 'Representation') || is_a($pubObject, 'SubmissionFile'));
-			$submissionDao = Application::getSubmissionDAO();
 			$submission = $submissionDao->getById($pubObject->getSubmissionId(), null, true);
+
 			if (!$submission) return null;
 			// Now we can identify the context.
 			$contextId = $submission->getJournalId();
 		}
+		// Get the latest submission revision
+		$latestRevision = $submissionDao->getLatestRevisionId($submission->getId());
```
lilients: hmmm, the function getCurrentVersionId also uses $submissionDao->getLatestRevisionId. So there is no advantage here ... 🤔

controllers/tab/issueEntry/form/IssueEntryPublicationMetadataForm.inc.php
```diff
@@ -284,11 +289,13 @@ function execute($request) {
 
 			if ($issue) {
 
+				// set publication date
+				$submission->setDatePublished($this->getData('datePublished') ? $this->getData('datePublished') : Core::getCurrentDate());
```
lilients: I don't think the date should be changed or deleted.

bozana: why not?

lilients: If you remove a submission (or a version of a submission) from an issue, the (whole) article does not show at the frontend. I think this makes sense. Also if you add the article again all old dates are still there, which is also good, I think.

bozana: Hmmm... Would that date then be the (first) date_published for that submission? If an editor assigns a submission to a future issue, then unassigns it and assigns it later to the correct (and lets say published) issue, both dates will be in the DB (which I do not think is necessary?) and would then the first date be the date of article first publication (which would be wrong, I think, no?)?

lilients: The dates are stored for each submission revision (in submission_settings). The date for the latest version is set, wenn this version is published. Unassigning a submission does not change the publication date, because this is at submission level at not at submission revision level. I will add a conceptional description of the versioning and the design decisions we made to the pkp wiki soon, so there will be a place to look stuff like this up...  📝

bozana: Sorry @lilients, one more question: is that date set when you assign a submission to a future issue?

bozana: Ah, maybe I should test it first... then I will probably understand... :-) 
Earlier, the date_published was in the published_submissions table and thus if a submission was unassigned from an issue, it would be deleted. And, as I said, I am a little bit worried about the case above: if I first assigned the article to a future issue and then to a real issue, which publication date will this article (in the version 1) have...

lilients: Here my first draft of a technical description: https://pkp.sfu.ca/wiki/index.php?title=Versioning

> And, as I said, I am a little bit worried about the case above: if I first assigned the article to a future issue and then to a real issue, which publication date will this article (in the version 1) have...

I tested it: the date remains the same. But I am not sure if we can/should prevent that ... at least the user can always change the date by hand ... 🤔

bozana: ok, thanks!!!

classes/article/PublishedArticleDAO.inc.php
lilients: I don't think the setting_value can be empty, because the publication date can not be removed once it has been set (only changed to another date).

classes/article/PublishedArticleDAO.inc.php
lilients: I don't think the setting_value can be empty, because the publication date can not be removed once it has been set (only changed to another date).

classes/article/PublishedArticleDAO.inc.php
lilients: I don't think the setting_value can be empty, because the publication date can not be removed once it has been set (only changed to another date).

classes/article/PublishedArticleDAO.inc.php
lilients: hmmm, no sure why, but the function seems to return the latest revision because the issue page displays the latest. 😖

classes/article/ArticleGalleyDAO.inc.php
bozana: sorry for late response @lilients :-( Leave it as it is then... :-)

pages/article/ArticleHandler.inc.php
bozana: Maybe to provide the submissionRevision in a different way to the download function and optional? -- so that the function could be self contained i.e. that it could be called directly and not only from the function view. Then the redirect could stay and you could only provide the parameter if necessary?

pages/article/ArticleHandler.inc.php
bozana: OK, lets leave it as it is for now... :-)

dbscripts/xml/upgrade.xml
```diff
@@ -43,6 +43,7 @@
 	<upgrade minversion="2.4.0.0" maxversion="3.0.9.9">
 		<data file="dbscripts/xml/upgrade/3.1.0_preupdate_review_assignments.xml" condition="return $installer->columnExists('review_assignments', 'cancelled');" />
 		<data file="dbscripts/xml/upgrade/3.1.0_preupdate_notes.xml" condition="return $installer->columnExists('notes', 'file_id');" />
+		<data file="dbscripts/xml/upgrade/3.1.0_preupdate_versioning.xml" condition="return $installer->columnExists('published_submissions', 'date_published');" />
```
bozana: there is no such table published_submissions in OJS 2.4.x

bozana: ah, sorry, there is a rename before, so it does exists...

PR #1281, opened 2017-02-27
id_ID translation of pubprocesslarge.png

mohammadrizki:
@mtub, here the translation of pubprocesslarge.png for id_ID


mtub:
Thank you, @mohammadrizki !

PR #1282, opened 2017-02-27
pkp/pkp-lib#2173 migrate review files

bozana:
s. 1) here: https://github.com/pkp/pkp-lib/issues/2173 
This migrates correctly the review files. This cannot be considered for those already migrated to and using OJS 3.


asmecher:
Yes, that query looks good. I do wish we had some way of keeping the old data around for users who had already upgraded -- would you mind posting a cross-DB version of that query to this issue, so that users who have already upgraded can execute it?

bozana:
@asmecher , do you mean like an extra PHP script?

asmecher:
Nope, just a SQL statement. See [here](http://stackoverflow.com/questions/13733415/mysql-cross-db-table-update-is-blowing-my-mind) for an example of a cross-database update. Someone who wanted to apply this would load their old OJS 2.x database into an empty DB, then use this query to update the OJS 3.x database with the missing data.

bozana:
OK :-) Shall this be integrated into the upgrade script (in that case, can I assume the DB name?) or just an extra SQL posted in the GitHub Issue for example?

asmecher:
Just post it here, thanks! I don't think we can code it into the upgrade script with any kind of certainty.

bozana:
Here the SQL, for those that already upgraded and using OJS 3:
`INSERT INTO ojs_pkp_stable_test.review_files (review_id, file_id) SELECT DISTINCT ra.review_id, afm.file_id FROM ojs_stable.article_files afm, ojs_stable.articles am, ojs_stable.review_assignments ra WHERE ra.submission_id = am.article_id AND am.review_file_id = afm.file_id`

PR #1289, opened 2017-03-02
pkp/pkp-lib#2173 (4) migrate date_response_due

bozana:
s. 4) here: https://github.com/pkp/pkp-lib/issues/2173


dbscripts/xml/upgrade.xml
```diff
@@ -158,6 +158,7 @@
 
 	<upgrade minversion="2.4.0.0" maxversion="3.0.9.9">
 		<data file="dbscripts/xml/upgrade/3.1.0_update.xml" />
+		<data file="dbscripts/xml/upgrade/3.0.0_review_assignments.xml" />
```
asmecher: This should either be named `3.1.0_review_assignments.xml`, or better yet, drop it into the existing `3_1_0_update.xml` script (since it's a single query).

bozana: Ah, yes, sure... :-\

PR #1306, opened 2017-03-09
Fix line-breaks in bibtex citation template

mittinatten:
Some line breaks were missing, and some lines were included as blanks when fields were missing.


mittinatten:
Apparently I used the name bibsys instead of bibtex in the commit (it's the name of our retired library system). I can redo the PR if that's something that matters.

asmecher:
Thanks, @mittinatten, that's fine. I'll forward-port to the `master` branch as well.

PR #1309, opened 2017-03-10
New Serbian locales for OJS 3.x

vedranvucic:
new translated locale files


mtub:
This is for pkp/pkp-lib#2337

asmecher:
There are some XML validation problems here -- see e.g. https://travis-ci.org/pkp/ojs/jobs/209839293

mtub:
Fixed the XML errors (cc: @vedranvucic). Mind having another quick look, @asmecher ?

vedranvucic:
Which files I should check?

mtub:
Hi @vedranvucic , nothing to check for you :) - there where two errors in both of the variants with XML tags not closed, but I fixed those already (e1a4621, 7c7785d, 968e28f,  94f9b62). I was just asking Alec to have another look because this is a huge commit, but the automatic test has gone through without problems.

bozana:
@asmecher will be less available for a few weeks, so I can take a look at this...

bozana:
@vedranvucic, why is the file locale/sr_RS@cyrillic/emailTemplates.xml  not in Cyrillic?

bozana:
@mtub, @vedranvucic, what about the current sr_SR folder? Should it be removed?

mtub:
Hi @bozana, let me take care of the old folder and of the locale registry file. will send a PR.

vedranvucic:
Hello, Mail templates are usually heavily edited by editorial boards so we left that file untranslated since we do not see purpose of translating something that will be radically changed anyway.

vedranvucic:
The current sr_RS loacle should be replaced and new locales will take place instead since there are two scripts in Serbian language.

PR #1317, opened 2017-03-17
pkp/pkp-lib#2173 (3) remove migrated cancelled review assignments

bozana:
s. number 3) here: https://github.com/pkp/pkp-lib/issues/2173
Because you can not delete rows in several tables with one SQL statement in PostgreSQL, I am using the DAO function for it...

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -1908,6 +1908,23 @@ class Upgrade extends Installer {
 		return true;
 	}
 
+	/**
+	 * For 2.4.x - 3.0.2 upgrade: remove cancelled review assignments.
+	 * @return boolean
+	 */
+	function removeCancelledReviewAssignments() {
+		$reviewAssignmentDao = DAORegistry::getDAO('ReviewAssignmentDAO');
+		// get cancelled review assignemnts
+		$result = $reviewAssignmentDao->retrieve('SELECT review_id FROM review_assignments WHERE cancelled = 1');
+		while (!$result->EOF) {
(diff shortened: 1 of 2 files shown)


asmecher:
Two suggestions:

- If we're not going to use the `cancelled` column in `review_assignments`, then remove it fully (e.g. from the DataObject, DAO and database table)
- Rather than deleting these without a trace, it would probably be a good idea to move them into the submission event log. I presume that's what happens with reviewer unassignments in 3.x?

bozana:
I agree with the suggestion number 1. 
Concerning suggestion number 2: If a review assignment is cancelled in OJS 2.4.x, the review assignment will not be removed (just "cancelled = 1" set) and there is the event log "The round X review by Y for submission Z has been cancelled". In OJS 3.x there is no cancel function and this attribute "cancelled" is not used in OJS 3.x. Instead there is the function "unassignReviewer" and in that case the review assignment will be removed and there will be the event log "The round X review by Y for submission Z has been cleared". Thus, after the migration from 2.4.x to 3.x the event logs for cancelled review assignments will be there -- they have slightly different text ("cancelled" vs. "cleared"), but... maybe no additional event log is needed when now clearing them?

asmecher:
Yes, that sounds OK -- as long as we don't need to maintain code in OJS 3.x for situations that could only arise in OJS 2.x. If we did that it would likely lead to breakage later down the line.

bozana:
closing this PR, because of these PRs, that only go to master:
ojs master: pkp/ojs#1321
reviewReport master: pkp/reviewReport#2
pkp-lib master: #2366

PR #1322, opened 2017-03-20
pkp/pkp-lib#2312 Add 'context' to list of managed pages

defstat:


pages/management/SettingsHandler.inc.php
```diff
@@ -47,6 +47,7 @@ function settings($args, $request) {
 			case 'index':
 			case '':
 			case 'journal':
```
asmecher: I would suggest renaming the endpoint to `context` instead of `journal`/`press` throughout -- no sense in maintaining both, and having two working endpoints for the same place will probably lead to inadvertent breakage later.

defstat: Fixed.

PR #1329, opened 2017-03-23
clean the upgrade scripts

bozana:
This reorganizes the upgrade scripts a little bit cleaner for 3.1. upgrade


bozana:
@asmecher, could you please take a look at this PR?
The upgrade scripts can be sorted differently than in the ojs-stable-3_0_2 branch, above all because we added a few fixes (upgrade scripts) after the release. 
Thanks!

bozana:
@asmecher, I will merge this, because I need it for further upgrade scripts... If not OK, I can reverse it, but I think it is good so... :-)

asmecher:
Sure, no objections -- sorry for the wait.

PR #1344, opened 2017-04-01
pkp/pkp-lib#2407 Remove mutex work-around for UNION queries

asmecher:


classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -246,6 +196,74 @@ function setOAIData($record, $row, $isRecord = true) {
...
+			FROM
+				published_submissions pa
+				JOIN submissions a ON (a.submission_id = pa.submission_id)
+				JOIN issues i ON (i.issue_id = pa.issue_id)
+				JOIN sections s ON (s.section_id = a.section_id)
+				JOIN journals j ON (j.journal_id = a.context_id)
+			WHERE	i.published = 1 AND j.enabled = 1 AND a.status <> ' . STATUS_DECLINED . '
+				' . ($journalId?' AND j.journal_id = ?':'') . '
+				' . ($sectionId?' AND s.section_id = ?':'') . '
+				' . ($from?' AND LEAST(a.last_modified, i.last_modified) >= ' . $this->datetimeToDB($from):'') . '
```
bozana: In the old code, it seems that generally always the latest/greatest/most recent date modified is used. Only for the function `getEarliestDatestamp`, that is only used for the `repositoryInfo` OAI function, the earliest date is important. 
I am not sure, but maybe it makes sense... For example: if we are harvesting every day and in one day either article or issue changes, we would like to catch that change... ?
Thus, maybe we have to differentiate this function (getEarliestDatestamp) from the other queries?

asmecher: @bozana, I think you're thinking of `MIN`/`MAX` rather than `GREATEST`/`LEAST`. The `getEarliestDatestamp` feature uses the sort order to fetch the single oldest record, using the `$orderBy` parameter to this function.

bozana: Hmmm... I am not sure I understand :-\ What does then this LEAST do here i.e. why not use GREATEST? 
I'll take a further look... but I actually believe you :-)
I just wanted to say that I think that the latest date modified should always be considered, for from as well until... right?

asmecher: I think this already does what you're suggesting -- we go by the earliest of the two dates, so if either an article or issue  modification happens, the `WHERE` clause (from/to dates) will be inclusive. Maybe you mean the `LEAST` call in the `SELECT` column list?

bozana: I think I suggest using not the earliest but the latest i.e. GREATEST date :-) For example, if both article and issue date modified are the same and = the day before yesterday, and a service provider harvested it so. Then one of them changed yesterday (in the evening). When the service provider harvests them next with from=yesterday (in the morning) it would not find them using the earliest i.e. LEAST, right?

asmecher: AH, I have finally gotten this through my thick head. Thanks for your patience, @bozana!

PR #1345, opened 2017-04-04
Consider  in article_summary.tpl and search.tpl, change searchHandler…

ajnyga:
… journalmustpublispolicy conditional

See: https://github.com/pkp/pkp-lib/issues/2010

### DIFF
--- a/pages/search/SearchHandler.inc.php
+++ b/pages/search/SearchHandler.inc.php
@@ -29,7 +29,7 @@ class SearchHandler extends Handler {
 	 */
 	function authorize($request, &$args, $roleAssignments) {
 		import('classes.security.authorization.OjsJournalMustPublishPolicy');
-		$this->addPolicy(new OjsJournalMustPublishPolicy($request));
+		if ($request->getContext()) $this->addPolicy(new OjsJournalMustPublishPolicy($request));
 
 		return parent::authorize($request, $args, $roleAssignments);
 	}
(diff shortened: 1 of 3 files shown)


get04:
Dear Ajnyga, 
The site wide search was not working and after doing the changes you suggested it searches , but the problem is the path with the search results. I made the changes from your commit, but still no success. Could you please help? btw i am using ojs 3.0.2
regards

ajnyga:
I wonder if it is a cache issue with the old article_summary.tpl?

get04:
http://10.6.71.154/index.php/index/user/authorizationDenied?message=user.authorization.contextRequired
I cleared the cache and checked , but no success
Thanks

ajnyga:
I mean the link in the list of search results, the link that should lead to the article.

get04:
10.9.6.23/index.php/index/article/view/9

ajnyga:
So does line 24 look like this:
`<a {if $journal}href="{url journal=$journal->getPath() page="article" op="view" path=$articlePath}"{else}href="{url page="article" op="view" path=$articlePath}"{/if} class="file">`

get04:
yes,it looks the same

ajnyga:
and you have also added the changes to templates/frontend/pages/search.tpl

get04:
Ok, Thanks. Then i wait to hear any update after your test.

regards

ajnyga:
do you use some custom theme, like bootstrap https://github.com/NateWr/bootstrap3/

get04:
yes, i used that

ajnyga:
Hi,

The problem is either
1) an issue I fixed with https://github.com/pkp/ojs/pull/1371/files
2) if you use the bootstrap theme, the template files there will be used instead and at the moment they do not include this modification. @NateWr

PR #1347, opened 2017-04-05
Corrected some glitches affecting issue view in vanilla theme

martinpub:
I noticed some inconsistency in the Vanilla theme, so here's a minor fix of the theme's CSS. (Don't know how many use this, but on our server we have several journals who use this theme.)


asmecher:
Merged -- thanks, @martinpub!

PR #1353, opened 2017-04-16
Allow artwork files for XML galleys

ajnyga:
HI @asmecher 

Based on you suggestion here http://forum.pkp.sfu.ca/t/artwork-files-for-xml-galleys/29587 I made a pr concerning the artwork files XML galleys. Allowing these is crucial for the implementation of XML workflows.

However, I started to think that should the `templates/controllers/grid/articleGalleys/form/articleGalleyForm.tpl` actually check for file type at all? I mean, why not allow artwork files for ANY galley? I realize that many usual galleys do not need or support separate artwork (like pdf), but still.

I am also preparing support for the XML artwork files both in my embedGalley plugin and the lensGalley plugin. However, I noticed that you have decided to stick with the old version of lens because it is missing support for mobile layouts. I have, however, used the latest lens release because it supports features likes footnotes. Making changes to the original lens.js files is very difficult. I am working with the lens and lens-starter repositories to create whole new lens.js files. In these versions I have also stripped most of the eLife filters that exist in the current lensGalley release. But more on those later.

### DIFF
--- a/templates/controllers/grid/articleGalleys/form/articleGalleyForm.tpl
+++ b/templates/controllers/grid/articleGalleys/form/articleGalleyForm.tpl
@@ -39,7 +39,7 @@
 		{/fbvFormSection}
 	{/fbvFormArea}
 
-	{if $articleGalleyFile && $articleGalleyFile->getFileType()=='text/html'}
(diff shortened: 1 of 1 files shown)


asmecher:
Merged, thanks. Agreed that this is currently in the wrong place, but it'll serve until we get the chance to move it elsewhere.

asmecher:
Also cherry-picked to `ojs-stable-3_0_2`.

PR #1357, opened 2017-04-20
pkp/pkp-lib#2392 migrate removed to the existing fields on the about …

bozana:
…page

s. https://github.com/pkp/pkp-lib/issues/2392

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -1952,6 +1952,277 @@ class Upgrade extends Installer {
 		return true;
 	}
 
+	/**
+	 * For 2.4.x - 3.1.0 upgrade: concatenate removed journal setting fields into the new journal setting 'about'.
+	 * @return boolean
+	 */
+	function concatenateIntoAbout() {
+		$journalDao = DAORegistry::getDAO('JournalDAO');
+		$journalSettingsDao = DAORegistry::getDAO('JournalSettingsDAO');
+		$journals = $journalDao->getAll();
+		while ($journal = $journals->next()) {
+			$supportedFormLocales = $journal->getSupportedFormLocales();
+			$focusAndScope = $journal->getSetting('focusScopeDesc');
+			$focusAndScope['localKey'] = 'about.focusAndScope';
+			$reviewPolicy = $journal->getSetting('reviewPolicy');
+			$reviewPolicy['localKey'] = 'about.peerReviewProcess';
+			$pubFreqPolicy = $journal->getSetting('pubFreqPolicy');
+			$pubFreqPolicy['localKey'] = 'about.publicationFrequency';
+			$oaPolicy = array();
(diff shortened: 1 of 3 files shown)


classes/install/Upgrade.inc.php
```diff
@@ -1930,6 +1930,277 @@ function removeCancelledReviewAssignments() {
...
+	 * @return boolean
+	 */
+	function concatenateIntoAbout() {
+		$journalDao = DAORegistry::getDAO('JournalDAO');
+		$journalSettingsDao = DAORegistry::getDAO('JournalSettingsDAO');
+		$journals = $journalDao->getAll();
+		while ($journal = $journals->next()) {
+			$supportedFormLocales = $journal->getSupportedFormLocales();
+			$focusAndScope = $journal->getSetting('focusScopeDesc');
+			$focusAndScope['localKey'] = 'about.focusAndScope';
```
asmecher: I think `localKey` should be `localeKey` throughout, no?

classes/install/Upgrade.inc.php
```diff
@@ -1930,6 +1930,277 @@ function removeCancelledReviewAssignments() {
...
+						if (!empty($customItem['title']) && !empty($customItem['content'])) {
+							$aboutJournal[$locale] .= '
+								<div id="custom-'.$index.'">
+								<h3>'.$customItem['title'].'</h3>
+								<p>'.nl2br($customItem['content']).'</p>
+								</div>';
+						}
+					}
+				}
+				// finaly, concatenate the other settings
```
asmecher: *finally

classes/install/Upgrade.inc.php
```diff
@@ -1930,6 +1930,277 @@ function removeCancelledReviewAssignments() {
...
+		}
+		return true;
+	}
+
+	/**
+	 * For 2.4.x - 3.1.0 upgrade: concatenate editorialTeam and displayMembership page to new journal setting 'masthead'.
+	 * @return boolean
+	 */
+	function concatenateIntoMasthead() {
+		$roleIdEditor = 0x00000100;
```
asmecher: I'd suggest putting the OJS 2.x constant names in comments -- it'll help when grepping and explain where the magic numbers come from.

asmecher: (I still prefer `define` statements for these, but it's not a big enough deal to warrant changing)

bozana: They are defined in a function above, so I thought it is more secure like this, depending on where and when they are needed, which function is when executed... to have them defined in the appropriate function... OK?

asmecher: Ah, I see. Hmm, maybe we should define them once at the top of the file, something like `OLD_ROLE_ID_...`? That way grepping will work, the constants will just be defined once, and magic numbers get documented.

bozana: OK, I can do that :-) THANKS!!!

PR #1358, opened 2017-04-20
pkp/pkp-lib#2235 submodule update

bozana:
just a submodule update


bozana:
ok, tests are failing, I'll take a look...

PR #1361, opened 2017-04-20
pkp/pkp-lib#2337 update locale registry and remove outdated Serbian locale

mtub:
this also addresses the latest comments at #1309


bozana:
Thanks a lot @mtub!!! :-)))

PR #1370, opened 2017-05-03
pkp/pkp-lib#1791 COUNTER plugin

bozana:
s. https://github.com/pkp/pkp-lib/issues/1791


plugins/reports/counter/templates/sushixml.tpl
```diff
@@ -1,8 +1,8 @@
 {**
- * plugins/generic/counter/sushixml.tpl
+ * plugins/reports/counter/templates/sushixml.tpl
  *
- * Copyright (c) 2014-2017 Simon Fraser University
- * Copyright (c) 2003-2017 John Willinsky
+ * Copyright (c) 2013-2017 Simon Fraser University
+ * Copyright (c) 2003-2016 John Willinsky
```
asmecher: I'm not sure if these changes are inadvertent -- but all files should be -2017.

bozana: Ah, that's why I copied the files from ojs-dev-2_4... I'll change it...

PR #1371, opened 2017-05-03
Fix article title link in article_summary.tpl

ajnyga:
@asmecher, an addition to: https://github.com/pkp/ojs/pull/1345

The original pr changed only the url for the cover image, the title link was unaffected. oops.


asmecher:
Merged, thanks!

PR #1374, opened 2017-05-08
pkp/pkp-lib#2500: Correct filehandling for Native Import/Export

ctgraham:
Ensure NativeImportExportPlugin only checks file permissions as needed for import and export from the CLI; add support for NativeXMLSumbissionFileFilter filehandling via lib/pkp update.

Resolves pkp/pkp-lib#2500 for OJS.


bozana:
This is all file, it would just need a rebase... THANKS!!! :-)

ctgraham:
I've dropped the explicit add of lib/pkp, but this will depend on https://github.com/pkp/pkp-lib/pull/2502 .

PR #1375, opened 2017-05-08
pkp/pkp-lib#2465 Permit author read access to copyedited and representation grids

asmecher:


templates/controllers/tab/authorDashboard/production.tpl
```diff
@@ -0,0 +1,22 @@
+{**
+ * lib/pkp/templates/controllers/tab/authorDashboard/production.tpl
```
bozana: also here...

PR #1376, opened 2017-05-10
Remove unnecessary print_r from ArticleHander

ajnyga:
@asmecher, there is an extra print_r command left probably from debugging which is causing headers already sent errors.


asmecher:
Oops! Yes, thanks.

PR #1380, opened 2017-05-12
pkp/pkp-lib#2516: Flag an error in Native Import if article has no date_published, but is a member of an issue

ctgraham:
Resolves a part of pkp/pkp-lib#2516.

Raises an error if the import has a `<issue>` element which contains `<article>` elements which do not have a `date_published` attribute.  Articles cannot be added to an issue without a published date.

### DIFF
--- a/plugins/importexport/native/filter/NativeXmlArticleFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlArticleFilter.inc.php
@@ -72,7 +72,7 @@ class NativeXmlArticleFilter extends NativeXmlSubmissionFilter {
 	}
 
 	/**
-	 * Populate the submission object from the node
+	 * Populate the submission object from the node, checking first for a valid section and published_date/issue relationship
 	 * @param $submission Submission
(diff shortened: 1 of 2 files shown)


plugins/importexport/native/filter/NativeXmlArticleFilter.inc.php
```diff
@@ -89,6 +89,12 @@ function populateObject($submission, $node) {
 				$submission->setSectionId($section->getId());
 			}
 		}
+		$datePublished = $node->getAttribute('date_published');
+		$issue = $deployment->getIssue();
```
bozana: An import could also be just import of one published article that would then have the child element `issue_identification`, that would be handled later i.e. there would be no `$issue` at this place yet. Thus, could we maybe also check if the child element exists and consider it in the `if` statement below i.e. if there is child element `issue_identification` and no `$datePublished` add the error?

ctgraham: Yes, this could be more complete.

PR #1382, opened 2017-05-12
pkp/pkp-lib#2449: Document missing parameters and return types in SectionDAO

ctgraham:
Part of cleanup for pkp/pkp-lib#2449.


asmecher:
Merged, thanks!

PR #1394, opened 2017-05-19
Change DOAJ API protocol to https

ajnyga:
Hi @bozana 

If a journal is SSL the DOAJ API returns 301 error when trying to export articles. This can be fixed by chancing the API address to https.

However, I do not know whether you want to use `//doaj.org/api/v1/` instead? @asmecher? But if so, note that most of the export plugins define the API protocoll at the moment.


ajnyga:
I actually think that the API is currently only available in https, they just did not upgrade the documentation https://doaj.org/api/v1/docs

PR #1397, opened 2017-05-22
Prevent duplicate load of issue in Native Import with identical issue identification

ctgraham:
This pull resolves pkp/pkp-lib#2462 by checking if a matching issue_identification is found before creating a brand new issue.

Some problems with this:
* The entire import process is far too dependent on reporting errors based on imported object id; IMHO this should be based on the XML input (node lineage? line/char?).  Here I've just reported the raw XML.
* There is an unfortunate dependency throughout the import process on the magic of the `$deployment`'s shared variables for context, issue, submission, etc.  I'm unhappy with the `$addProcessedObject` parameter in `setIssue()` as a workaround for this.
* Reusing an existing issue based on issue_identification means that the issue will be updated, with no path to rollback possible in the event of error!

### DIFF
--- a/plugins/importexport/native/NativeImportExportDeployment.inc.php
+++ b/plugins/importexport/native/NativeImportExportDeployment.inc.php
@@ -69,7 +69,6 @@ class NativeImportExportDeployment extends PKPNativeImportExportDeployment {
 	 */
 	function setIssue($issue) {
 		$this->_issue = $issue;
-		if ($issue) $this->addProcessedObjectId(ASSOC_TYPE_ISSUE, $issue->getId());
 	}
 
 	/**
--- a/plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
@@ -64,25 +64,33 @@ class NativeXmlIssueFilter extends NativeImportFilter {
 	function handleElement($node) {
 		$deployment = $this->getDeployment();
 		$context = $deployment->getContext();
-		$user = $deployment->getUser();
 
-		// Create and insert the issue (ID needed for other entities)
 		$issueDao = DAORegistry::getDAO('IssueDAO');
-		$issue = $issueDao->newDataObject();
-		$issue->setJournalId($context->getId());
-		$issue->setPublished($node->getAttribute('published'));
-		$issue->setCurrent($node->getAttribute('current'));
-		$issue->setAccessStatus($node->getAttribute('access_status'));
-
-		$issueDao->insertObject($issue);
+		// if the issue identification matches an existing issue, flag to process only child objects
+		$issueExists = false;
+		$issue = $this->_issueExists($node);
+		if ($issue) {
+			$issueExists = true;
+		} else {
+			// Create and insert the issue (ID needed for other entities)
+			$issue = $issueDao->newDataObject();
+			$issue->setJournalId($context->getId());
+			$issue->setPublished($node->getAttribute('published'));
+			$issue->setCurrent($node->getAttribute('current'));
+			$issue->setAccessStatus($node->getAttribute('access_status'));
+			$issueDao->insertObject($issue);
+			$deployment->addProcessedObjectId(ASSOC_TYPE_ISSUE, $issue->getId());
+		}
 		$deployment->setIssue($issue);
 
 		for ($n = $node->firstChild; $n !== null; $n=$n->nextSibling) {
 			if (is_a($n, 'DOMElement')) {
-				$this->handleChildElement($n, $issue);
+				$this->handleChildElement($n, $issue, $issueExists);
 			}
 		}
-		$issueDao->updateObject($issue); // Persist setters
+		if (!$issueExists) {
+			$issueDao->updateObject($issue); // Persist setters
+		}
 		return $issue;
 	}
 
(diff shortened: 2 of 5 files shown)


plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
```diff
@@ -74,8 +73,21 @@ function handleElement($node) {
 		$issue->setCurrent($node->getAttribute('current'));
 		$issue->setAccessStatus($node->getAttribute('access_status'));
 
-		$issueDao->insertObject($issue);
-		$deployment->setIssue($issue);
+		$foundIssue = null;
+		foreach ($node->getElementsByTagName('issue_identification') as $n) {
+			$searchIssue = $issueDao->newDataObject();
+			$this->parseIssueIdentification($n, $searchIssue, false);
```
bozana: How about letting this function return either an issue, if such an issue is found, or null? So that we would not create newDataObject above, but first below, if returned issue is null?

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
```diff
@@ -74,8 +73,21 @@ function handleElement($node) {
...
+			$foundIssues = $issueDao->getIssuesByIdentification($context->getId(), $searchIssue->getVolume(), $searchIssue->getNumber(), $searchIssue->getYear(), $searchIssue->getTitle(null));
+			foreach ($foundIssues->toArray() as $foundIssue) {
+				$deployment->addWarning(ASSOC_TYPE_ISSUE, $foundIssue->getId(), __('plugins.importexport.native.import.error.issueIdentificationDuplicate', array('issueId' => $foundIssue->getId(), 'issueIdentification' => $n->ownerDocument->saveXML($n))));
+			}
+		}
+		if ($foundIssue) {
+			$issue = $foundIssue;
+		} else {
+			$issueDao->insertObject($issue);
```
bozana: here we would then create newDataObject, set the attributes from above (journalId, published, current, access_status) and insert it?

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
```diff
@@ -74,8 +73,21 @@ function handleElement($node) {
...
 
-		$issueDao->insertObject($issue);
-		$deployment->setIssue($issue);
+		$foundIssue = null;
+		foreach ($node->getElementsByTagName('issue_identification') as $n) {
+			$searchIssue = $issueDao->newDataObject();
+			$this->parseIssueIdentification($n, $searchIssue, false);
+			$foundIssues = $issueDao->getIssuesByIdentification($context->getId(), $searchIssue->getVolume(), $searchIssue->getNumber(), $searchIssue->getYear(), $searchIssue->getTitle(null));
+			foreach ($foundIssues->toArray() as $foundIssue) {
+				$deployment->addWarning(ASSOC_TYPE_ISSUE, $foundIssue->getId(), __('plugins.importexport.native.import.error.issueIdentificationDuplicate', array('issueId' => $foundIssue->getId(), 'issueIdentification' => $n->ownerDocument->saveXML($n))));
```
ctgraham: I included this for two reasons:
1) it is a change from earlier behavior, where a new duplicate issue with the same identification would be created
2) any attributes of the issue added/modified by this import will *not* be rolled back in the event of error.

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
ctgraham: +1 to this.  I'm thinking too hierarchically, and too oriented to backwards compatibility.  This would be a step in the right direction.

plugins/importexport/native/NativeImportExportDeployment.inc.php
```diff
@@ -66,10 +66,11 @@ function getSchemaFilename() {
 	/**
 	 * Set the import/export issue.
 	 * @param $issue Issue
+	 * @param $addProcessedObject boolean Should this issue be identified as a newly processed object?
 	 */
-	function setIssue($issue) {
+	function setIssue($issue, $addProcessedObject = true) {
 		$this->_issue = $issue;
-		if ($issue) $this->addProcessedObjectId(ASSOC_TYPE_ISSUE, $issue->getId());
+		if ($issue && $addProcessedObject) $this->addProcessedObjectId(ASSOC_TYPE_ISSUE, $issue->getId());
```
ctgraham: Really, it is the `NativeXmlIssueFilter` which should act on its `$deployment` object to call `addProcessedObjectId()`, similar to the way we handle sections there.  That's where we really decide if we're adding a new Issue.

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
ctgraham: Actually, if we've found a matching issue, maybe we should avoid modifying it, and that would be the real meaning of this message.

plugins/importexport/native/filter/NativeXmlIssueFilter.inc.php
ctgraham: I tried doing this a couple of times, but each time ended up looking a lot like the current code.  I'll push a halfway-to-this change; let me know what you think.

ctgraham:
Improved with respect to my concerns about with the `$addProcessedObject` parameter in `setIssue()`, and the problem of updating an issue, with no path to rollback possible in the event of error.

bozana:
Thanks a lot @ctgraham !!!

PR #1401, opened 2017-05-23
pkp/pkp-lib#2537 article_galley_settings migration

bozana:
s. https://github.com/pkp/pkp-lib/issues/2537

### DIFF
--- a/dbscripts/xml/upgrade/3.1.0_update.xml
+++ b/dbscripts/xml/upgrade/3.1.0_update.xml
@@ -69,5 +69,11 @@
 		<!-- remove old, removed journal settings migrated/concatenated to the new, existing -->
 		<query>DELETE FROM journal_settings WHERE setting_name IN ('focusScopeDesc', 'reviewPolicy', 'pubFreqPolicy', 'openAccessPolicy', 'customAboutItems', 'contributorNote', 'contributors', 'sponsorNote', 'sponsors', 'history')</query>
 	</sql>
+	<sql><!-- pkp/pkp-lib#2537 -->
+		<!-- copy article_galley_settings into submission_galley_settings -->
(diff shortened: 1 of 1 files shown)


dbscripts/xml/upgrade/3.1.0_update.xml
```diff
@@ -69,5 +69,11 @@
 		<!-- remove old, removed journal settings migrated/concatenated to the new, existing -->
 		<query>DELETE FROM journal_settings WHERE setting_name IN ('focusScopeDesc', 'reviewPolicy', 'pubFreqPolicy', 'openAccessPolicy', 'customAboutItems', 'contributorNote', 'contributors', 'sponsorNote', 'sponsors', 'history')</query>
 	</sql>
+	<sql><!-- pkp/pkp-lib#2537 -->
+		<!-- copy article_galley_settings into submission_galley_settings -->
+		<query>INSERT INTO submission_galley_settings (galley_id, locale, setting_name, setting_value, setting_type) SELECT galley_id, locale, setting_name, setting_value, setting_type FROM article_galley_settings WHERE setting_name &lt;&gt; 'excludeDoi' AND setting_name &lt;&gt; 'excludeURN'</query>
```
asmecher: It looks like this setting is moved from article-specific to galley-specific; I guess that's intentional? Is there no more role for those settings on the article level? (If it's something we might re-add to articles later, in addition to the galley level, it might be worth leaving the `submission_settings` entries around.)

bozana: Hmmm... @asmecher, I am not sure I understand :-( These are all galley settings, just that they were not migrated into the corresponding submission_galley_settings at all, which is now happening here. Or do you mean the next row? -- Those two settings are not needed any more...

asmecher: Oops, I did slightly misread the code. Yes, I was wondering about the deletions from `submission_settings` -- are these truly not needed, or is it something that might get re-added later?

bozana: This excludeDOI and excludeURN option is truly not needed any more, because our model/workflow changed -- now the DOIs/URNs have to be explicitly assigned...

asmecher: OK, great -- please go ahead!

PR #1404, opened 2017-05-25
pkp/pkp-lib#2548: When logging reviewer actions, check if the user is logged in or is using an access key to ensure user is associated with log event

ctgraham:
Resolves pkp/pkp-lib#2548.  Suitable for cherry-pick across to stage 2.4.8.

### DIFF
--- a/classes/submission/reviewer/ReviewerAction.inc.php
+++ b/classes/submission/reviewer/ReviewerAction.inc.php
@@ -68,7 +68,13 @@ class ReviewerAction extends Action {
 
 				// Add log
 				import('classes.article.log.ArticleLog');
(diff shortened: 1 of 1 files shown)


asmecher:
Not a big deal -- but why not just call `logEventHeadless` rather than the near-duplication of calls to either `logEvent` or `logEventHeadless`?

ctgraham:
Preserving the original `logEvent` calls made sense to me at the time, out of some desire to preserve the original function if the user was logged in.  Re-reading it, however... you are right.  We can always just call `logEventHeadless` with a ternary operator on the user to guarantee the same backwards compatibility.

ctgraham:
I think if I close and reopen this, the Travis CI build restarts?

asmecher:
Travis testing on the stable branch is currently broken since they rolled out a VM update. Unfortunately it's not an easy fix -- we went through all kinds of LAMP stack hell on the master branch to get it going again -- so I suspect we'll probably leave it that way.

PR #1406, opened 2017-05-29
Fix DOAJ export article URL's when using Acron

ajnyga:
Hi @bozana 

While working with this https://github.com/pkp/pkp-lib/issues/2543, I noticed that the journal article url's depend on the journal that triggers the export. The pr fixes that, but does not solve the issue linked.


bozana:
Thanks a lot @ajnyga!

PR #1408, opened 2017-05-29
remove pubIds importexport plugin

bozana:
this plugin is not migrated to ojs 3.x and because it seems to be rarely used, it will be removed for now and then we will see if there are requirements to re-implement it for 3.x...


bozana:
@asmecher, could you take a look? THANKS!!!

asmecher:
Please go ahead!

PR #1414, opened 2017-06-06
pkp/pkp-lib#2569 OAI fixes

bozana:
s. https://github.com/pkp/pkp-lib/issues/2566


asmecher:
The material change is the one related to `dot.set_spec`, right? Does this affect both journal and section sets?

bozana:
@asmecher, what do you exactly mean? :-\

asmecher:
Ah, OK, just making sure that collisions aren't possible -- and it looks like they aren't. Please go ahead!

PR #1428, opened 2017-06-16
Ojs stable 3 0 2

ronniebrito:
pt_BR locale update


asmecher:
Looking at the commit list, I think this might've been sent to the wrong branch?

PR #1434, opened 2017-06-21
Update ru_RU locale after #1430

ppv1979:


ppv1979:
@mtub, can you check this PR?

PR #1438, opened 2017-06-22
pkp/pkp-lib#2602 Small update for fr_CA translation plugins

pnault:
-block/../../../locale.xml: wording
-importexport/../../../local.xml: all new file, adding 13 keys


mtub:
This is for pkp/pkp-lib#2602

PR #1449, opened 2017-07-11
Small corrections in en_US strings

martinpub:
Hi, just to suggest two corrections (whitespace and duplicate line) found during the work with the Swedish translation.


asmecher:
Thanks, @martinpub!

PR #1451, opened 2017-07-13
Update ru_RU locale after i1816-fix2 and #1449

ppv1979:


asmecher:
Thanks, @ppv1979!

mtub:
pkp/pkp-lib#2911

PR #1455, opened 2017-07-18
Restore automatic selection of the section when submitting an article

ptrckb:
OJS 2.4 automatically selects the section of a submission if there is a single journal section defined. This pull request replicate this behaviour with OJS 3.


asmecher:
Looks good -- thanks, @ptrckb!

PR #1456, opened 2017-07-19
Fix user merge on master

ptrckb:
It is not possible to merge users anymore since the renaming of methods in `classes/subscription/IndividualSubscriptionDAO.inc.php` (pkp/ojs@70ddecd).

Maybe a similar glitch on https://github.com/pkp/ojs/blob/master/classes/journal/JournalDAO.inc.php#L87 ?


asmecher:
Merged -- and also fixed the JournalDAO case, thanks. This code is a work in progress.

PR #1459, opened 2017-07-20
pkp/pkp-lib#2663: Prevent IssueDAO::resequenceCustomIssueOrders() from creating a new arbitrary ordering where none existed

ctgraham:
Resolves pkp/pkp-lib#2663.

An alternate strategy could be to continue to wrap every call to `IssueDAO::resequenceCustomIssueOrders()` with a check, but that seems unnecessary.

Also considered was adding `i.date_published DESC` to the `$this->retrieve()` of 670 so that even if called it would at least order them correctly, but this also seems unnecessary.


asmecher:
Looks good, and net decrease in code size -- thanks, @ctgraham!

PR #1341, opened 2017-03-30
Translation into Ukrainian

alexukua:
Revised, stylistically edited, translated into Ukrainian and structured according English version

PR #1465, opened 2017-07-26
new uk_UA update

mtub:
* updates from @alexukua, see #1341
* this is for pkp/pkp-lib#2670
* will need updates, see original PR and tracking issue


mtub:
I will merge this - we received the files quite some time ago, it took a few updates until they were ready. In the meantime, we received another translation, which we will now have to compare to this one.

PR #1468, opened 2017-07-26
pkp/pkp-lib#2671 crossref deposit setting name migration

bozana:
s. https://github.com/pkp/pkp-lib/issues/2671


asmecher:
Merged, thanks! (Not sure what the testing problem is, but it's unrelated.)

PR #1470, opened 2017-07-27
Update ru_RU locale after #2163 and #2434

ppv1979:


mtub:
Thanks, Pavel!

mtub:
pkp/pkp-lib#2911

PR #1476, opened 2017-08-03
pkp2017-sprint Issue export change

defstat:
Added some changes to export all issues if the "php importExport.php export ..." command line tool has an "issues" switch without issue ids.

### DIFF
--- a/plugins/importexport/native/NativeImportExportPlugin.inc.php
+++ b/plugins/importexport/native/NativeImportExportPlugin.inc.php
@@ -260,10 +260,16 @@ class NativeImportExportPlugin extends ImportExportPlugin {
 		$exportFilter = array_shift($nativeExportFilters);
 		$exportFilter->setDeployment(new NativeImportExportDeployment($context, $user));
 		$issues = array();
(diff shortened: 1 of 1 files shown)


asmecher:
The tool's usage documentation would need updating to describe this.

defstat:
@asmecher Do you mean https://github.com/pkp/ojs/blob/master/docs/IMPORTEXPORT#L132? If yes doesn't this usage statement declares issueIds as optional?

asmecher:
There's usage information in the locale files, e.g. `plugins/importexport/native/locale/en_US/locale.xml`.

defstat:
@asmecher My question remains: doesn't this usage statement declares issueIds as optional? 
(so the usage: `export [xmlFileName] [journal_path] issues [issueId1] [issueId2]` is valid even if we type 
`export [xmlFileName] [journal_path] issues` as a command? )

bozana:
@defstat, no it is the syntax for a variable, I think -- to differentiate the command and its options from the variables that user has to type in... If no issue ID is given nothing will be exported...

PR #1478, opened 2017-08-07
pkp/pkp-lib#2570: Escape display of references rather than striping HTML-like tags.

ctgraham:
Consider Harvard and ABNT citation styles as examples of why `strip_unsafe_html` should not be used.


asmecher:
Merged, thanks!

PR #1482, opened 2017-08-10
pkp/pkp-lib#2603 new pl_PL locale

mtub:
see pkp/pkp-lib#2603 and commit messages


mtub:
will tackle a few more issues with this PR before merging…

mtub:
Hi @szmigieldesign, a few comments so far:

* in the future, we'd like you to follow the en_US file structure, especially the XML file headers (copyright dates etc.)
* you've included lots of files that are not part of the OJS release, some PKP plugins hosted independently and also some plugins that are not from PKP - I will strip them from your submission
* your files still had lots of English text, in the future I would ask that you removed this yourself before submitting
* I will continue to prepare the plugins, but in the future I'd strongly preferred if you could upload everything to GitHub (as pull requests for the correct repositories)

szmigieldesign:
Hello,
thanks @mtub. It was a rather wild ride with translating those XML files since we've started with 2.x a long time ago and hence several outdated files in this release. Also, I was unable to use translator plugin's built in export function as it's blocked on the hosting I was working on and I had to manually prepare the ZIP.

PR #1502, opened 2017-08-22
Update version.xml

marcbria:
Why include version in application name? 
Instead of "ojs2" why not "ojs"?

### DIFF
--- a/dbscripts/xml/version.xml
+++ b/dbscripts/xml/version.xml
@@ -11,7 +11,7 @@
   -->
 
 <version>
-	<application>ojs2</application>
+	<application>ojs</application>
 	<type>core</type>
 	<release>3.0.2.0</release>
 	<tag>ojs-3_0_2-0</tag>


marcbria:
A few more files also refer to "ojs2":
- classes/core/Application.inc.php
- docs/README
- lib/pkp/xml/schema/common.xml
- lib/pkp/xml/schema/plugins.xsd
- lib/pkp/classes/site/Version.inc.php (and the regexp)
- plugins/gateways/resolver/README
- plugins/importexport/medra/tests/functional/serial-issue-as-work.xml
- ...

marcbria:
I asked about this change not only for consistency... 

When you make a clean install of ojs3, the database username, username, OAI... are taken from there "'databaseUsername' => Application::getName()" and references to "ojs2" on "ojs3" looks like we missed something.

BTW, "ojs3" is also an option, but I prefer a permanent solution.

asmecher:
This isn't going to change for a while. The whole version check infrastructure depends on the application being `ojs2`, and the issue is cosmetic and internal -- OJS 3.x doesn't identify itself as `ojs2` anywhere externally. I'm open to contributions here -- but they have to work fully :)

asmecher:
(I've committed changes to a number of the other issues you flagged, however!)

marcbria:
I full understand Alec.

>  I'm open to contributions here.
It's just a grep&replace task or it's much complex than this?

> (I've committed changes to a number of the other issues you flagged, however!)
yeah. thanks for this.

asmecher:
It'll take more than that -- the bootstrapping code that fetches the version descriptor expects the product to be `ojs2`, so changing it to something else will cause OJS to have an identity crisis. It would take some fancy footwork in the upgrade code to get around this for little gain, so I consider it a low priority (though, granted, an annoying inconsistency).

marcbria:
Thanks for the explanation Alec. I can live with this. ;-)

PR #1535, opened 2017-09-19
Fix the construction of the url in notification emails for editors.

ptrckb:
Instead of testing the role of the recipient of the email, it was testing the role of the current user.

This resulted in a bad URL in the emails received by editors : for example `https://.../index.php/JOURNAL/reviewer/submissionReview/123#peerReview` in a notification email for an uploaded review file.


asmecher:
Merged -- thanks, @ptrckb!

PR #1555, opened 2017-10-03
pkp/pkp-lib#2832 remove citationFormats in code

bozana:
https://github.com/pkp/pkp-lib/issues/2832


bozana:
@NateWr, could you please also take a look at this PR?

PR #1567, opened 2017-10-09
pkp/pkp-lib#2850 remote galleys display in layout.tpl

bozana:
s. https://github.com/pkp/pkp-lib/issues/2850


bozana:
@asmecher, any idea what is wrong with the tests i.e. could I ignore it here? THANKS!!!

asmecher:
The `ojs-dev-2_4` and `ojs-stable-2_4_8` tests are broken since the Travis upgrade a few months back, and I don't think it's worth the time investment to fix them as we have for `master` and `ojs-stable-3`xyz.

PR #1447, opened 2017-07-04
pkp/pkp-lib#2604 update pt_BR locale

mtub:
files received from Ronnie, see https://github.com/pkp/ojs/pull/1444

this is for pkp/pkp-lib#2604

PR #1570, opened 2017-10-09
pkp/pkp-lib#2604 new updates for pt_BR

ronniebrito:


mtub:
Please don't commit submodules in the future (`lib/pkp`, `plugins/generic/pdfJsviewer`). Will fix this locally.

mtub:
Why did you remove the &nbsp; entities?

mtub:
Thanks, @ronniebrito, for your updates. There were again lots of errors probably introduced by your text editor's settings. Also again duplicate keys, strings in the wrong keys. In the future, I won't be able to track all of these down. We will talk about workflows after this release with all translators. But in either case we will expect cleaner updates in the future, because it sometimes just takes too much time to tidy the files.

closing, superseded by #1447

PR #1574, opened 2017-10-11
Add citation_journal_abbrev meta to Google Scholar plugin

ajnyga:
See request here: https://forum.pkp.sfu.ca/t/include-abbreviation-of-journal-title-in-the-html-header/34172/2

Uses same logic as here: https://github.com/pkp/ojs/blob/master/plugins/importexport/crossref/filter/IssueCrossrefXmlFilter.inc.php#L152-L156


asmecher:
Perfect -- thanks, @ajnyga!

PR #1577, opened 2017-10-11
Small fixes for de_DE and nl_NL locales

stweil:


locale/de_DE/manager.xml
```diff
@@ -320,7 +320,7 @@
 	<message key="manager.setup.typeProvideExamples">Beispiele für im Fachgebiet relevante Forschungsansätze, -methoden oder -verfahren angeben.</message>
 	<message key="manager.setup.uniqueIdentifier">Eindeutige Kennung</message>
 	<message key="manager.setup.uniqueIdentifierDescription">Artikel und Ausgaben können mit einer Identifikationsnummer oder einer Zeichenkette gekennzeichnet werden, wobei ein System, z.B. das DOI (Digital Object Identifier)-System, verwendet werden sollte.</message>
-	<message key="manager.setup.useEditorialReviewBoard">Die Zeitschrift hat eine Redaktion/ein Herausgeber/innen- und Gutachter/innengremium.</message>
+	<message key="manager.setup.useEditorialReviewBoard">Die Zeitschrift hat eine Redaktion / ein Herausgeber/innen- und Gutachter/innengremium.</message>
```
mtub: Please revert this change. No spaces before/after the slashes in the German locale.

mtub:
Fixed that sentence and another phrase. Will merge now.

locale/de_DE/manager.xml
stweil: Sure? I followed the suggestion of the [German Wikipedia](https://de.wikipedia.org/wiki/Schr%C3%A4gstrich#Typografie) and several other references here. They say that sequences of words which are separated by a slash should add spaces around the slash.

As a native German I find that more legible. I know that there are more places where the spaces would have to be added, but I thought that I'd like to start here.

Of course I can revert that part for now (and send a pull request for it later if needed) if you prefer that.

stweil:
Thanks. 'bzw.' is indeed much better than a slash.

PR #1581, opened 2017-10-12
[OJS] Pkp/pkp lib#2178 add nav menu manage

defstat:


NateWr:
Looks good. I'm just getting one missed locale string:

selection_150

defstat:
@NateWr OK

PR #1595, opened 2017-10-18
Adding missing sv_SE plugin translation file.

martinpub:


mtub:
This is for pkp/pkp-lib#2864

PR #1604, opened 2017-10-20
update ru_RU locale after #1963

ppv1979:


ppv1979:
Shibboleth plugin was deleted, so this PR don't need anymore

PR #1610, opened 2017-10-23
update locale registry

mtub:
completing translation updates for OJS 3.1 :-)


asmecher:
This isn't nearly big enough but :tada: :tada: :tada:!

PR #1613, opened 2017-10-23
pkp/pkp-lib#2506 fix genre id in file names

bozana:
s. https://github.com/pkp/pkp-lib/issues/2506


bozana:
I cannot wait for the test here, I will merge, the next PR is coming so maybe I will have more luck there...

PR #1614, opened 2017-10-23
Spanish translation for plugins/themes/default/locale/

jorgeandr3s:
Spanish translation for plugins/themes/default/locale/


mtub:
@marcbria , please let me know if there's any update…

PR #1619, opened 2017-10-23
pkp/pkp-lib#2582 fix repairImageAssociations

bozana:
s. https://github.com/pkp/pkp-lib/issues/2582

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -2306,6 +2306,8 @@ class Upgrade extends Installer {
 	 * @return boolean
 	 */
 	function repairImageAssociations() {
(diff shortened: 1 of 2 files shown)


bozana:
@asmecher, I believe something is wrong with the tests -- they do not start :-( Else, if you have time to take a look at this PR or just merge it -- I've tested it, so I believe everything is ok... :-)

bozana:
I'll merge this and the test problem can be solved later then...

asmecher:
We're quite back-logged on our tests on the Travis service -- https://travis-ci.org/pkp/ojs/builds shows 10 builds of OJS, with 2 of them running. Unfortunately we'll just have to be patient!

PR #1629, opened 2017-10-24
Update German translations and add missing ones

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>


stweil:
They are missing in `ojs-stable-3_1_0`, too.

There is a `manager.paymentTypes` and a `manager.payment.paymentTypes` which looks redundant.

stweil:
I added some more files now.

PR #1630, opened 2017-10-24
Update Upgrade.inc.php (issue #2506)

Vitaliy-1:
Tagging @bozana 
Related to: https://github.com/pkp/pkp-lib/issues/2506#issuecomment-339044181


bozana:
@Vitaliy-1, thanks a lot for the PR! :star: I would however close it, because the missing of the genre keys in the DB should be repaired, as discussed further in the issue https://github.com/pkp/pkp-lib/issues/2506

PR #1634, opened 2017-10-25
pkp/pkp-lib#2954 Possible error on line 113 of UserAction.inc.php

jpgrego:
While attempting to merge two users, an error has occured because it tries to
invoke the "next()" method on an array. In this commit we suggest a possible fix
by invoking the next php array function, passing the array as argument.


asmecher:
Thanks! Merged and cherry-picked to `ojs-stable-3_1_0`.

PR #1646, opened 2017-10-30
Add an example .htaccess

marcbria:
An example .htaccess will help those who want to enable restful_urls.
This sample.htaccess will be also useful for containers/virtualization automations.

### DIFF
--- a/sample.htaccess
+++ b/sample.htaccess
@@ -1,0 +1,6 @@
+<IfModule mod_rewrite.c>
+   RewriteEngine on
+   RewriteCond %{REQUEST_FILENAME} !-d
+   RewriteCond %{REQUEST_FILENAME} !-f
+   RewriteRule ^(.*)$ index.php/$1 [QSA,L]
+</IfModule>


asmecher:
The `config.inc.php` file already refers to `docs/FAQ` in the `restful_urls` documentation, and the sample config you have here is already in the FAQ: https://github.com/pkp/ojs/blob/ojs-3_1_0-0/docs/FAQ#L282..L287

I'd rather not duplicate these areas if not necessary -- maybe a reference to the FAQ entry from somewhere else you expect to see it?

marcbria:
Hi @asmecher 

Let me explain the motivation to see if I can convince you. :-)

As you know, I'm working on docker files for OJS and one rule in docker is keep those files to minimum.

This is an example my docker recipe to download and install OJS (ready for 2.4.5, 2.4.8, 3.0.2 and 3.0.1): https://hub.docker.com/r/marcbria/docker-pkp-ojs/~/dockerfile/

The LAMP stack of the containers are ready for restfull_urls so to complete the process, I need an htaccess and enable apache. Right now is done as follows:

```
# Setting Apache
RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf
COPY default.htaccess /var/www/html/.htaccess
RUN a2enmod rewrite \
    && service apache2 restart 
```
Of course, I can keep a htaccess in my repository (as I'm doing now), but I think people feels more comfortable if files are download from the official PKP repo, instead from an anonymous guy.

Nowadays, a site without clean_urls, looks like from the last century, so this is not an original idea: For instance, drupal distributes a very well documented htaccess (enabled by default): https://github.com/drupal/drupal/blob/8.4.x/.htaccess

So I thought a middle way between "no file and enabled by default" was including an example htaccess file in ojs repo, and modify FAQs to rename this file.

In summary:

- Easier to enable restful_urls.
- Make orchestration tools (over docker/vagrant/whatever...) easier.
- First steep in the way to restful_urls=On as default.

What do you think?

asmecher:
How about putting that file in `docs/`, and editing the FAQ entry so that it refers to it? That would meet your needs, and avoid duplication of config samples.

marcbria:
I buy it. ;-)

README/UPGRADE can refer to this file too.

PR #1666, opened 2017-11-03
pkp/pkp-lib#3002 submodule update ##bozana/ojs-stable-3_1_0##

bozana:
s. https://github.com/pkp/pkp-lib/issues/3002

only submodule update


bozana:
This was the wrong solution, thus closing...

PR #1680, opened 2017-11-11
Ojs stable 3 1 0

rainwolf60:
how does it possible to make phone filed in user contacts ,to display on registration form as a required field?


asmecher:
@rainwolf60, if you're wanting to ask a question, post it on our forum at http://forum.pkp.sfu.ca/. If you're trying to make a code contribution, double-check which branches you're using.

PR #1686, opened 2017-11-13
pkp/pkp-lib#3052 consider url encoding for oai record identifiers

bozana:
s. https://github.com/pkp/pkp-lib/issues/3052


bozana:
Wrong solution thus closing

PR #1693, opened 2017-11-14
pkp/pkp-lib#2989 correct PostgreSQL update queries

bozana:
s. https://forum.pkp.sfu.ca/t/error-while-upgrading-ojs-from-2-4-8-to-3-1-postgres-only/34627/9


bozana:
I've just tested the upgrade: yes, the aliases work... so closing...

PR #1695, opened 2017-11-14
pkp/pkp-lib#2989 correct PostgreSQL update queries

bozana:
s. https://forum.pkp.sfu.ca/t/error-while-upgrading-ojs-from-2-4-8-to-3-1-postgres-only/34627/9


asmecher:
Hmm, this looks like an exact revert of the change proposed in the top of the thread, which appears to have been tested... Maybe this is a PostgreSQL version issue?

bozana:
@asmecher, it is not the same -- in the earlier version, before you change it, there was a alias -- maybe that's why it did not work -- there should be full main table name i.e. submission_files...

bozana:
@asmecher, I have PostgreSQL 9.4.12 and I get this error:
```
ERROR:  syntax error at or near "sf"
LINE 1: ...T COUNT(*) AS total FROM (UPDATE submission_files sf SET gen
```
when I execute the update statement with the alias:

`UPDATE submission_files sf SET genre_id = g.genre_id FROM genres g, submissions s, article_galleys_migration agm WHERE g.entry_key = 'STYLE' AND g.context_id = s.context_id AND s.submission_id = sf.submission_id AND sf.file_id = agm.style_file_id`

Although the syntax in the documentation allows it, s. https://www.postgresql.org/docs/9.4/static/sql-update.html

Hmmm... :-\

bozana:
Hmmm... now it worked... I will test the whole upgrade tomorrow...
Also if it works we should then remove `sf.submission_files.submission_id` i.e. using both alias and table name...

bozana:
I've just tested the upgrade: yes, the aliases work... so closing...

PR #1696, opened 2017-11-14
fix an upgrade typo and remove unused filters on time

bozana:


bozana:
@asmecher, could you please take a look at this?

asmecher:
Wow, I totally missed that double alias. Weird that some versions of PostgreSQL didn't mind. I've also back-ported that single commit back to `ojs-stable-3_1_0` -- does the other query need porting too?

bozana:
Thanks @asmecher!
No the other commit is only for master... :-)

PR #1697, opened 2017-11-14
pkp/pkp-lib#2989 fix a typo, using alias and table name

bozana:


bozana:
Apparently this is fixed/merged... thus closing...

PR #1701, opened 2017-11-15
Fix oai_marc template for schema compliance in dev-2_4

ctgraham:
Replace instances of marcxml tags/attributes with oai_marc tags/attributes.  (ojs-dev-2_4)

See: pkp/pkp-lib#3062


bozana:
Thanks a lot @ctgraham!!!

PR #1734, opened 2017-11-29
3125

bozana:
s. https://github.com/pkp/pkp-lib/issues/3125

only submodule update


bozana:
irrelevant thus closing....

bozana:
ah reopening because of the submodule update...

PR #1757, opened 2017-12-12
Add section id to TOC

dastewo:
Hello there,

as mentioned [https://forum.pkp.sfu.ca/t/add-section-id-to-toc/36173](url) it would be a good feature to add section ids to the issue tocs, so people might be able to style every section in a different way and it might be a foundation for anchor links.

edit: I'm new to Github and I think I might have done this pull request the wrong way. I'm sorry!

Kind regards
Daniela


asmecher:
@dastewo, I think this was sent into the wrong branch. If you're working with a branch of `ojs-stable-3_1_0`, then make sure when you create the PR to select the right branch. This one looks like it was based on `ojs-stable-3_1_0` but sent in against the `master` branch. If you get it right, you should see only your commits listed.

PR #1758, opened 2017-12-13
pkp/pkp-lib#3181 Add id attribute to TOC sections for styling

dastewo:
Hello there,

as mentioned https://forum.pkp.sfu.ca/t/add-section-id-to-toc/36173 it would be a good feature to add section ids to the issue tocs, so people might be able to style every section in a different way and it might be a foundation for anchor links.

Kind regards
Daniela

### DIFF
--- a/templates/frontend/objects/issue_toc.tpl
+++ b/templates/frontend/objects/issue_toc.tpl
@@ -94,8 +94,8 @@
 
 	{* Articles *}
 	<div class="sections">
-	{foreach name=sections from=$publishedArticles item=section}
-		<div class="section">
+	{foreach name=sections key=currSectionId from=$publishedArticles item=section}
+		<div class="section" id="sectionId_{$currSectionId}">
 		{if $section.articles}
 			{if $section.title}
 				<h2>


gemusehandler:
I would like to underscore the need for something that help to apply specific styles to specific sections. The approach here is rather complicated. I successfully use a class for this purpose:
<div class="{$section.title|escape}">

NateWr:
@gemusehandler We'd be happy to receive a PR that added a unique class name. You can see some discussion on this at https://github.com/pkp/pkp-lib/issues/3181.

Using `<div class="{$section.title|escape}">` won't work for multilingual journals, since the class name will change depending on the language being shown.

Ideally, I'd like to see two class names applied: one that uses the section abbreviation in the journal's primary language and one that uses the id. The first will resolve challenges between dev/staging/production that are mentioned in the issue. The second will work even in cases where the section abbreviation will not generate a workable class name (ie - when it uses cyrillic, hebrew or arabic characters).

gemusehandler:
I painfully became aware myself that the $section.title didn’t work in the one Russian journal I have. The section abbreviation might do the trick but I am not able to call it.

NateWr:
In order to get this, we'll need to pass a sanitized section abbreviation for each section here: https://github.com/pkp/ojs/blob/master/pages/issue/IssueHandler.inc.php#L326-L327.

Perhaps what we can do is call it a `slug` and use a sanitized section abbreviation. If no section abbreviation exists, it can fallback to the `id`. That should give us the best of both worlds: the ability to use a stable class name across dev/staging/production builds, and a reliable fallback when abbreviations are not available.

PR #1759, opened 2017-12-13
Instead of issue title, show issue identification in statistics output

ajnyga:
Hi @asmecher 

At the moment if you output statistics that have issue information, the report generator will either show the issue title or just the id. This does not tell a lot for the editors.

Instead, getIssueIdentification should be used.


bozana:
Yes, @ajnyga, this makes sense! I will allow myself to merge this...

asmecher:
Thanks, @ajnyga and @bozana!

PR #1782, opened 2018-01-05
pkp/pkp-lib#3242 Upgrade Smarty to v3

asmecher:


asmecher:
I grepped for those and replaced them on Friday and forgot to push up the changes.

It sounds like the big thing that needs testing/fixing is the plugin template override stuff. I'll have a look at that next.

PR #1790, opened 2018-01-13
[ojs] #3056 Port StaticPages as NMIs

defstat:


classes/install/Upgrade.inc.php
```diff
@@ -2575,6 +2575,33 @@ function repairSuppFilesFilestage() {
 		return true;
 	}
 
+	/**
+	 * If StaticPages table exists we should port the data as NMIs
```
asmecher: Missing `@return`

classes/install/Upgrade.inc.php
```diff
@@ -2575,6 +2575,33 @@ function repairSuppFilesFilestage() {
...
+			$contextDao = Application::getContextDAO();
+			$navigationMenuItemDao = DAORegistry::getDAO('NavigationMenuItemDAO');
+
+			import('plugins.generic.staticPages.classes.StaticPagesDAO');
+
+			$staticPagesDao = new StaticPagesDAO();
+
+			$contexts = $contextDao->getAll();
+			while ($context = $contexts->next()) {
+				$contextStaticPages = $staticPagesDao->getByContextId($context->getId())->toAssociativeArray();
```
asmecher: Not a big deal, but why not just iterate rather than converting to an array?

PR #1805, opened 2018-01-24
pkp/pkp-lib#3311 rename journal setting subject into subjects

bozana:
s. https://github.com/pkp/pkp-lib/issues/3311


dbscripts/xml/upgrade/3.1.1_update_a.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE data SYSTEM "../../../lib/pkp/dtd/xmlData.dtd">
+
+<!--
+  * 3.1.1_update_a.xml
```
asmecher: I think this could probably just be included in the existing `3.1.1_update.xml` script.

asmecher: (I think that's probably what you intended, since this isn't referenced from `dbscripts/xml/upgrade.xml`)

bozana: It is referenced form upgrade.xml, but you are right -- I could include it in 3.1.1_update.xml because it does not matter if those migrating/upgrading from OJS 2.4.x do not have those settings... :-)

PR #1806, opened 2018-01-24
pkp/pkp-lib#2291: Use the localized page title for the site, not the page header

ctgraham:
Resolves pkp/pkp-lib#2291


asmecher:
Yep, I remember some conversation about this distinction. Thanks for the quick turn-around, @ctgraham.

PR #1836, opened 2018-02-13
pkp/pkp-lib#2683: remove password value from user block

ctgraham:
Resolves pkp/pkp-lib#2683.

Like the username, the password field should never be pre-populated in the block.


bozana:
Thanks @ctgraham!

PR #1849, opened 2018-02-22
Fixed parameter names in getSummaryProperties

MarHerUMR:
The parameter should be @$section@ and not @$ection@ in @getSummaryProperties@.


asmecher:
Good spotting -- thanks, @MarHerUMR.

PR #1850, opened 2018-02-24
changing the accept title from skipReview to accept

johanesmikhael:
Dear all, I suggest to not use editor.submission.decision.skipReview for submission stage, because it will be recorded as (Accept and Skip Review) in editorial history (even with the review process has been conducted). I did a quick fix but if there is a more elegant solution it will be better I think.

### DIFF
--- a/classes/workflow/EditorDecisionActionsManager.inc.php
+++ b/classes/workflow/EditorDecisionActionsManager.inc.php
@@ -135,7 +135,7 @@ class EditorDecisionActionsManager {
 			SUBMISSION_EDITOR_DECISION_ACCEPT => array(
 				'name' => 'accept',
 				'operation' => 'promote',
-				'title' => 'editor.submission.decision.skipReview',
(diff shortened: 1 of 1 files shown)


asmecher:
@johanesmikhael, if I understand your point correctly: The "Accept and skip review" will send the submission directly to proofreading, thus skipping review as advertised -- but since we don't prevent users from going back to earlier stages and working with them further, it's still possible to use the review tools when the submission has already been sent to copyediting.

johanesmikhael:
The problem is, even if all of the review processes has been done, and the button from editors page already changed to "Send To Copyediting" (I forgot the exact title), it still logged it as "Skip and Review" in the Editorial History. How was that happen?

bozana:
@johanesmikhael, I will close this PR, because the problem should be solved differently -- I opened a new GitHub Issue for that: https://github.com/pkp/pkp-lib/issues/3543 -- track that for the solution...
THANKS A LOT!!!

PR #1857, opened 2018-02-28
pkp/pkp-lib#3416 migrate sr_SR to sr_RS@latin

bozana:
s. https://github.com/pkp/pkp-lib/issues/3416

I put everything in the PHP function instead of an XML file for the users to be able to manually change the new locale to sr_RS@cyrillic if wished so.

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -2604,6 +2604,132 @@ class Upgrade extends Installer {
 
 		return true;
 	}
+
+	/**
+	 * Migrate sr_SR locale to the new sr_RS@latin.
+	 * @return boolean
+	 */
+	function migrateSRLocale() {
+		$oldLocale = 'sr_SR';
+		$newLocale = 'sr_RS@latin';
+
+		$oldLocaleStringLength = 's:5';
+
+		$journalSettingsDao = DAORegistry::getDAO('JournalSettingsDAO');
+
+		// Consider all DB tables that have locale column:
+		$dbTables = array(
+			'announcement_settings', 'announcement_type_settings', 'author_settings', 'books_for_review_settings', 'citation_settings', 'controlled_vocab_entry_settings',
+			'data_object_tombstone_settings', 'email_templates_data', 'email_templates_default_data', 'external_feed_settings', 'filter_settings', 'genre_settings', 'group_settings',
+			'issue_galleys', 'issue_galley_settings', 'issue_settings',	'journal_settings', 'library_file_settings', 'metadata_description_settings',
+			'navigation_menu_item_assignment_settings', 'navigation_menu_item_settings', 'notification_settings', 'referral_settings',
+			'review_form_element_settings', 'review_form_settings', 'review_object_metadata_settings', 'review_object_type_settings', 'rt_versions', 'section_settings', 'site_settings',
+			'static_page_settings', 'submissions', 'submission_file_settings', 'submission_galleys', 'submission_galley_settings', 'submission_settings', 'subscription_type_settings',
+			'user_group_settings', 'user_settings',
+		);
+		foreach ($dbTables as $dbTable) {
+			if ($this->tableExists($dbTable)) {
+				$journalSettingsDao->update('UPDATE '.$dbTable.' SET locale = \''.$newLocale.'\' WHERE locale = \''.$oldLocale.'\'');
+			}
+		}
+		// Consider other locale columns
+		$journalSettingsDao->update('UPDATE journals SET primary_locale = \''.$newLocale.'\' WHERE primary_locale = \''.$oldLocale.'\'');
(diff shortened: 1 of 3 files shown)


classes/install/Upgrade.inc.php
```diff
@@ -2604,6 +2604,132 @@ function migrateStaticPagesToNavigationMenuItems() {
...
+
+		$oldLocaleStringLength = 's:5';
+
+		$journalSettingsDao = DAORegistry::getDAO('JournalSettingsDAO');
+
+		// Consider all DB tables that have locale column:
+		$dbTables = array(
+			'announcement_settings', 'announcement_type_settings', 'author_settings', 'books_for_review_settings', 'citation_settings', 'controlled_vocab_entry_settings',
+			'data_object_tombstone_settings', 'email_templates_data', 'email_templates_default_data', 'external_feed_settings', 'filter_settings', 'genre_settings', 'group_settings',
+			'issue_galleys', 'issue_galley_settings', 'issue_settings',	'journal_settings', 'library_file_settings', 'metadata_description_settings',
```
asmecher: Misplaced tab :)

classes/install/Upgrade.inc.php
```diff
@@ -2604,6 +2604,132 @@ function migrateStaticPagesToNavigationMenuItems() {
...
+			'navigation_menu_item_assignment_settings', 'navigation_menu_item_settings', 'notification_settings', 'referral_settings',
+			'review_form_element_settings', 'review_form_settings', 'review_object_metadata_settings', 'review_object_type_settings', 'rt_versions', 'section_settings', 'site_settings',
+			'static_page_settings', 'submissions', 'submission_file_settings', 'submission_galleys', 'submission_galley_settings', 'submission_settings', 'subscription_type_settings',
+			'user_group_settings', 'user_settings',
+		);
+		foreach ($dbTables as $dbTable) {
+			if ($this->tableExists($dbTable)) {
+				$journalSettingsDao->update('UPDATE '.$dbTable.' SET locale = \''.$newLocale.'\' WHERE locale = \''.$oldLocale.'\'');
```
asmecher: I know it's not strictly needed in this case, but it's still better to bind in values using `?` to let ADODB handle the escaping, and use `qStr` to wrap `$dbTable`.

asmecher: (Actually, no, I think `qStr` won't work -- that one probably needs to stay literal.)

asmecher:
Phheewww, that's (necessarily) an ugly bit of upgrade script. I think it needs just a couple of things:

- Conversion of glued-together SQL with `?` bindings, where possible (e.g. not on table names, but definitely on string values etc)
- I think it would be possible to check whether the journal used `sr_SR` e.g. in the journal/site settings before converting any data, and if not, to skip the entire block. I suggest this because a small minority of journals will use `sr_SR`, and the migration code is funky enough that I'd be happier not running it at all if possible.

Thanks, @bozana!

bozana:
@asmecher, regarding the second point: I first also thought so, but then I realized that it could be that one used sr_SR and it does not use it any more, so that it still has some data in the DB. However this could be a very rare/marginal case, if it exists at all. So OK, yes, will do so.
Another option would be to provide a CLI script instead of that function, that only the users who would need it would have to run. 
What do you prefer? 
THANKS!!!

bozana:
Ah, I think it is better we provide it like this -- this way we are sure we fixed it at this point...
Would it be enough to check if sr_SR is in site's installed languages and just in that case to run the script?

bozana:
@asmecher, I've added the changes, would you like to take one more look? -- I don't know if I can provide $settingNames that appear in the "... WHERE setting_name IN $settingNames..." as parameter to the DB function? Is it OK so?

asmecher:
This comes down to taste, but I prefer either...

    query('SELECT ... FROM ... WHERE columnName IN (?, ?, ?)', array('val1', 'val2', 'val3'));

...or...

    query("SELECT ... FROM ... WHERE columnName IN ('val1', 'val2', 'val3')");

...rather than...

    $someVar = '\'val1\', \'val2\', \'val3\';
    query('SELECT ... FROM ... WHERE columnName IN ('.$someVar.')');

For readability, I also prefer to use `"` quotes rather than `'` when `'` will require escaping in order to use it in SQL.

But these are all discretionary and I don't think it's worth spending too much time tinkering with it, especially if you're happy with what's already written!

asmecher:
> Would it be enough to check if sr_SR is in site's installed languages and just in that case to run the script?
Yes, I think so. It's possible there will be `sr_SR` content in the database after disabling the locale, but I would argue that this is a rare marginal case on a rare marginal case, and that the user isn't working with that content actively anyway.

bozana:
Thanks @asmecher, OK, then I will only change "" instead of '' in $settingNames but leave it as it is and merge... THANKS!!!

PR #1862, opened 2018-03-02
pkp/pkp-lib#3416 migrate sr_SR to sr_RS@latin

bozana:
This is a fix for those using OJS 3.1.0-1 i.e. ojs-stable-3_1_0


bozana:
I will not merge this, but just that it is there for those who need it... The changes will be included in the next OJS release...

PR #1871, opened 2018-03-12
Extend delayed open access months

tgc:
This was done in 2.4.x but not in OJS 3.
For reference:
https://forum.pkp.sfu.ca/t/delayed-open-access-in-ojs/28930


asmecher:
Thanks, @tgc!

PR #520, opened 2015-06-25
pkp/pkp-lib#519 Implement submission queries copyediting process

asmecher:

PR #1878, opened 2018-03-16
[OJS] #520 hide attachment on email not send

defstat:


bozana:
Ah, I merged the pkp-lib PR but now I see that I cannot merge this one -- could you rebase and add the new submodule update but then to ##pkp/master## ? THANKS!!!

bozana:
Ah, I think you do not have to update the submodule -- someone will do it with the next PR or so... So I will just close this PR...

PR #1880, opened 2018-03-19
Move self archiving settings under distribution->permissions

ajnyga:
Self archiving settings are at the moment under Payment settings, which does not make sense, because self archiving policies are not connected to subscription settings.

So, move settings concerning self archiving under settings=>distribution=>permissions


asmecher:
Merged, thanks, @ajnyga!

PR #1888, opened 2018-03-23
Fi locale for 3.1.1

ajnyga:
@asmecher Finnish locale for OJS 3.1.1


ajnyga:
ah ok sorry, I used the translate plugin and forgot to check the dates!

PR #1898, opened 2018-03-29
use constant instead of number

bozana:


bozana:
@asmecher, if this fix can still go.., else, no problem, after the release... :-)

PR #1901, opened 2018-04-09
pkp/pkp-lib#3556 submodule update ##bozana/3556##

bozana:
s. https://github.com/pkp/pkp-lib/issues/3556

only submodule update


bozana:
wrong branch....

PR #1917, opened 2018-04-16
pkp/pkp-lib#3593 fix issue galley locale validation

bozana:
s. https://github.com/pkp/pkp-lib/issues/3593


bozana:
@asmecher, any idea why that test is failing? :-\

asmecher:
That looks like a temporary network hiccough -- I've restarted that test, but I think it'll finish fine.

PR #1944, opened 2018-05-01
Include institution name and IP range variables in subscription locales

kaitlinnewson:
Noticed these were missing in our French and Spanish language journals, and thought it may be helpful to include them in all missing locale files as well.


asmecher:
Thanks and good spotting, @kaitlinnewson! Merged and will cherry-pick to `ojs-stable-3_1_1` in a sec.

PR #1947, opened 2018-05-02
Html galleys display typo

Vitaliy-1:
Suppose, there is a typo in `display.tpl` file in the htmlArticleGalley plugin.


asmecher:
Thanks, @Vitaliy-1, good catch. I'll cherry-pick to `ojs-stable-3_1_1` too.

PR #1948, opened 2018-05-03
 Fix missing submission id event log parameters

ptrckb:
With OJS 2.4, the `log.author.submitted` events of submissions prior to 642fa8543782c3df69e2e0f77d08d6d43bbe753d are missing the `submissionId` param.


asmecher:
Ah, I finally get what you were suggesting -- disregard my last two comments. However, one change I'd request: this should be executed inside an `<upgrade>` clause in the upgrade script so that it's limited to the affected versions.

ptrckb:
Moved the script inside an `upgrade` element.

asmecher:
Thanks, @ptrckb, and sorry for the delay! Merged and ported to `ojs-dev-2_4`.

PR #1949, opened 2018-05-03
pkp/pkp-lib#3663 log event entry parameter articleId

ptrckb:


asmecher:
@ptrckb, is the `articleId` setting used in OJS 2.x, or is this just preparation for a migration to 3.x?

ptrckb:
Moved the script in a `upgrade` element.

`articleId` is used in OJS 2.x messages. With OJS 3, log events use `submissionId` instead, see #1950 and pkp/pkp-lib#2160 was supposed to update from one to the other.

PR #1950, opened 2018-05-03
pkp/pkp-lib#3663 fix missing submission id params of migrated event l…

ptrckb:
…og entries


asmecher:
This one needs the same change as the `master` branch PR -- moving the script into the `<upgrade>` clause. Thanks, by the way, these look good and upgrade scripts are not easy!

ptrckb:
Moved the script inside an `upgrade` element.

PR #1951, opened 2018-05-03
fix individual subscription dao

bozana:
s. https://github.com/pkp/pkp-lib/issues/3664


bozana:
Nothing major to review here I believe, so I will merge it...

PR #1953, opened 2018-05-03
pkp/pkp-lib#3601 Extend article report

asmecher:


bozana:
Maybe only one comment/question/suggestion here: maybe to only consider the article locale, as this was done for additional metadata (e.g. keywords), instead of using getLocalized... ? -- Maybe it will be more consistent that way? And if it turns out that the users would need to have everything i.e. also article data in other languages i.e. the translations, maybe we could consider having them in a separate row?

asmecher:
Manually merged to `ojs-stable-3_1_1` and `master`.

PR #1954, opened 2018-05-06
[OJS] #3462 Refine password element for user import (and possibly export)

defstat:


plugins/importexport/users/UserImportExportPlugin.inc.php
```diff
@@ -75,15 +75,15 @@ function executeCLI($scriptName, &$args) {
 
 		switch ($command) {
 			case 'import':
-				$this->importUsers(file_get_contents($xmlFile), $journal, null);
+				$this->importUsers(file_get_contents($xmlFile), $journal, null, null);
```
asmecher: I think this will cause a warning, because `null` isn't a reference and the function expects one.

defstat: @asmecher you are right. I changed the `importUsers`, `exportUsers` and `exportAllUsers` methods to have the reference parameter null as default, and removed the passed null from `UserImportExportPlugin`

PR #1967, opened 2018-05-14
Update locale.xml

zuphilip:
I came across this sentence today and find it hard to understand. Hope this PR clarifies this. CC @mtub


mtub:
Thanks @zuphilip. I am not sure the new translation will be much clearer - "Bearbeitungszeit" sound like an editorial task, but what is meant in both the English and the German version is that DOIs are not being registered instantly because of cronjob configuration options, connection problems etc. I have opened an issue https://github.com/pkp/pkp-lib/issues/3720 and will provide an update to the English and the German strings.

Thanks!

PR #1968, opened 2018-05-16
pkp/pkp-lib#3578 Resolve PHP warnings

crism:
Clean-up done, at least as far as the admin side goes, which is where it was messiest. Fixes pkp/pkp-lib#3578.


crism:
Abandoning in favor or more focused PRs.

PR #1978, opened 2018-05-22
[OJS] #3261 review assignments export

defstat:


plugins/importexport/native/filter/filterConfig.xml
```diff
@@ -124,6 +124,62 @@
 			description="plugins.importexport.native.description"
 			inputType="xml::schema(plugins/importexport/native/native.xsd)"
 			outputType="class::classes.article.ArticleGalley[]" />
+		<!-- ReviewRounds native XML export -->
+		<filterGroup
```
asmecher: Hmm, a lot of these are going to be duplicated in OJS and OMP. I think it should be possible to move filter configurations into pkp-lib -- some of the other plugins already have these. Mind giving it a try?

defstat: @asmecher I have migrated the newly created filters to a new filterConfig.xml file on pkp-lib level. Maybe we could migrate Author and the various SubmissionFiles [SupplementaryFile, SubmissionArtworkFile, ] filters as well.

asmecher: That makes sense; feel free to tackle it if you like.

PR #1983, opened 2018-05-23
de_DE locale update

mtub:
completing a file-by-file comparison of de_DE and en_US


bozana:
Thanks @mtub!!!
I'll see if it should be applied to the ojs-stable-3_1_1 as well...

PR #1998, opened 2018-06-04
Fix word form of "verwenden" (=use) in de_DE

zuphilip:
CC @mtub


mtub:
I remember when we introduced this… sounds rather old-fashioned today, thanks for the suggestion :)

zuphilip:
@mtub My first association, when reading it, was "verwandt" like "Verwandtschaft" and not "verwenden".

PR #2002, opened 2018-06-04
Update UPGRADE

zuphilip:
See also https://forum.pkp.sfu.ca/t/solved-ojs-3-1-1-0-cannot-see-submissions-and-user-roles/45494/4


asmecher:
Merged -- thanks, @zuphilip!

zuphilip:
BTW is there any reason that this is text file and not a markdown file which would be shown nicer here in GitHub? @asmecher

asmecher:
Strictly historical (we migrated from CVS to github some years ago). PRs welcome as always!

PR #2003, opened 2018-06-05
Convert some docs into markdown

zuphilip:
Discussed in #2002.

### DIFF
--- a/docs/REPORTING-BUGS.md
+++ b/docs/REPORTING-BUGS.md
@@ -1,13 +1,13 @@
-Reporting Bugs
---------------
+## Reporting Bugs
+
 Bugs can be reported through Git issues at:
-<https://github.com/pkp/pkp-lib#issues>
+https://github.com/pkp/pkp-lib#issues
 
-Support Forum
--------------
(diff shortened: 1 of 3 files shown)


mtub:
I like this. If we don't just want to rely on GitHub rendering, maybe also transform the external links? Like  "https://github.com/pkp/pkp-lib#issues" etc. in REPORTING-BUGS.md - although this will make reading the raw file without any rendering a little more difficult. I'd still go with putting the links in there…

zuphilip:
Are relative links only working in GitHub? Relative links work also for different branches, but maybe this is not a common scenario... I can do either way.

asmecher:
Thanks, @zuphilip, and this is pretty uninvasive for non-github users. Merging now, but for extra bonus points, can you make the equivalent change in the OMP repository?

zuphilip:
@asmecher Okay, I can also look for the other files in `doc` here first and then switch to OMP for the extra points :sparkles: .

asmecher:
That would be fantastic, thanks!

PR #2006, opened 2018-06-07
pkp/pkp-lib#3781 Overriding recommendByAuthor plugin's template from the theme

Vitaliy-1:
Suppose, it should take place until an upgrade to Smarty v3 as per Nate's suggestion.
https://github.com/pkp/pkp-lib/issues/3781


asmecher:
I think this is already covered by 9c2d9580b007a41f4c187c9960501cbf02d6eb63 (recently merged), thanks!

PR #2012, opened 2018-06-12
Proper migration of no_NO to nb_NO

tgc:
I noticed that there where hundreds of no_NO references still in the database after a test upgrade from 2.4.8.1 to 3.1.1-1.
This patch series attempts to fix this by using the code from pkp/pkp-lib#3416 to do a more complete migration.

In my testing this has converted all traces of no_NO to nb_NO as expected.


bozana:
Thanks @tgc!!!

PR #2014, opened 2018-06-14
pkp/pkp-lib#3794 Add Reference Number column to subscription overview

tgc:
This is a response to a long standing issue we've had which @neffe first brought up here:
https://forum.pkp.sfu.ca/t/show-reference-numbers-on-the-summary-page/13794


asmecher:
Created an issue for this: https://github.com/pkp/pkp-lib/issues/3794

PR #2021, opened 2018-06-21
pkp/pkp-lib#3016 patch not to migrate supp files as galleys but as su…

bozana:
…bmissions files on OJS 3.1.1-2

this is the patch for those that do not want to migrate supp files form OJS 2 to the galleys in OJS 3. here, the supp files will be in the submission stage, in the "submission files" grid.


bozana:
Closing, because this is only a patch for others to use... It should not be merged...

luizbgomide:
Is that still appllicable to the latest version (3.2.1.1)??

PR #2027, opened 2018-06-28
Workflow variation implementation with english and french translations.

jdpro:
@asmecher @bozana 
This PR is related to this debate : https://forum.pkp.sfu.ca/t/2-books-for-review-questions/29234/18
The feature has been requested by other OFR users.
It has been presented here : https://github.com/pkp/pkp-lib/issues/2325
We have this slight variation on a production instance and we would like it not to remain a fork.


asmecher:
Thanks, @jdpro -- we're likely not to release any more OJS 2.x packages, and have essentially decided not to add new functionality to OJS 2.x, but as you note this is a useful input for the OJS 3.x objects for review rewrite which is still under consideration. Tagging @sfbetz, who has been considering design.

jdpro:
Thank you for this return @asmecher. The integration of this branch would be very useful for us: the plugin with the variation is used by our main social science journal. If the fork is not integrated, they will be stuck on OJS2 event if you port the plugin to OJS 3. You spoke me about automatic migration opportunities last year : https://forum.pkp.sfu.ca/t/2-books-for-review-questions/29234/9. I communicated about the existence of our variant in October: https://github.com/pkp/pkp-lib/issues/2325. Now the journal is in a delicate situation, because it uses an incompatible fork. I don't know if @sfbetz has the opportunity to merge, but that would solve a big problem...

asmecher:
@jdpro, if the OJS3 OFR plugin takes shape it may not be a straight-forward port from the OJS2 code -- it might be a bit of a redesign as well. This PR is flagged in the BFR/OFR rewrite issue (https://github.com/pkp/pkp-lib/issues/2325) for consideration so it shouldn't be forgotten in that process.

PR #2047, opened 2018-07-10
Fix typo in CrossrefInfoSender

ajnyga:
@bozana this is a typo right?


bozana:
Thanks a lot @ajnyga!!!

PR #2057, opened 2018-07-12
pkp/pkp-lib#3578: harmonizing GridHandler signatures

crism:
Fixes part of pkp/pkp-lib#3578.


asmecher:
Tests failed here -- I think it's because you have two submodule hash updates. Squash 'em together.

PR #2069, opened 2018-07-18
pkp/pkp-lib#3935 get the submisison after saving the schedule for pub…

bozana:
…lication form again

s. https://github.com/pkp/pkp-lib/issues/3935


asmecher:
Manually merged to master.

PR #2072, opened 2018-07-25
Proposal: Changed display of workflow buttons to highlight current status

retostauffer:
Changed how the workflow buttons (decline, request revision, accept) will be displayed to highlight the current state. Allows the editor to change decisions but hides the buttons as soon as a decision has been recorded. Has been discussed in https://github.com/pkp/pkp-lib/issues/3386


asmecher:
@retostauffer, thanks! I suspect this pull request was sent against the wrong branch -- make sure it's created against the same branch that your modifications are based on. You should only see your additions listed on the commits list / files changed.

asmecher:
@retostauffer, I think this needs to be re-created against the right branch; let me know if you get stuck.

PR #2083, opened 2018-08-01
Update locale.xml in es_ES lang

gonetil:
Improved already existing translations and added missing ones


mtub:
@marcbria , could you have a look?

marcbria:
Good enough.
Thanks @gonetil for the work done.

PR #2091, opened 2018-08-08
Improved cs_CZ translation

Jirrka:
Here are some files with updated Czech translation - I am totally new to GitHub so I am not sure this is the correct way to submit the translation


Jirrka:
Updated pull request already applied

PR #2094, opened 2018-08-13
Improved cs_CZ translation

Jirrka:


plugins/generic/lucene/locale/cs_CZ/locale.xml
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mtub: This is just a copy of the English file. Please remove.

PR #2096, opened 2018-08-16
Check that pubIdPlugins array is set before foreach

ajnyga:
Fix https://github.com/pkp/pkp-lib/issues/3992


asmecher:
(Also cherry-picked back to `ojs-stable-3_1_1`.)

PR #2099, opened 2018-08-17
pkp/pkp-lib#3578: harmonizing Plugin signatures

crism:
Fixes some of pkp/pkp-lib#3578. Some other specific plugins still have problems in `getContents()`, `getEnabled()`, and `register()`.


asmecher:
Merged, thanks!

PR #2100, opened 2018-08-17
pkp/pkp-lib#3578: harmonizing PKPUserSettingsDAO signatures

crism:
Fixes part of pkp/pkp-lib#3578.


asmecher:
@crism, I think I'd find this change confusing -- the `assocId` parameter is forcibly overridden, so it's effectively noise anywhere it's being used (since all calls come through the child implementation).

Instead, I'd suggest renaming the parent class implementation of getSetting to `getByAssoc`. That way both function signatures could coexist without raising warnings.

asmecher:
Merged manually (submodule conflict).

PR #2105, opened 2018-08-25
Make sure contextPath does not collide with an url in removeBaseUrl

ajnyga:
See: https://github.com/pkp/pkp-lib/issues/4009

ping @asmecher 

Two things to take into account here:
1. This does not fix the collisions issue, if the contextPath is "article" or "issue". Not sure if this is an issue (pi). Is there a black list of contextPath names that OJS will not allow?
2. There are several other places where contextPath is being compared to an url. Could be that they have a similar issue?


asmecher:
(This is just a submodule commit -- I've added that manually, thanks!)

asmecher:
@bozana, when you're available, would you mind taking a look at @ajnyga's comments here? We might need to file a follow-up issue. I've merged the proposed changes here (https://github.com/pkp/pkp-lib/pull/4020) but I don't know the stats code well enough to comment more broadly.

PR #2106, opened 2018-08-25
Add cookie consent option for GDPR compliance

ajnyga:
@NateWr 

Issue: https://github.com/pkp/pkp-lib/issues/3624

third time the charm? Closed https://github.com/pkp/ojs/pull/2036


ajnyga:
so it is failing almost in the same place here: https://github.com/pkp/pkp-lib/blob/master/tests/WebTestCase.inc.php#L129

The earlier versions failed on row 127. The commits are basically identical so no idea why it fails on that row now. I can not even see how the changes would affect the login?

ajnyga:
@NateWr if you want to test this here are the branches:
OJS: https://github.com/pkp/ojs/pull/2106
PKP-LIB: https://github.com/pkp/pkp-lib/pull/4015

I do not understand the test failure

PR #2119, opened 2018-09-15
best order and more concordance in Spanish emails

carzamora:
I ordered according to English emails, I use "Estimado/a" at first of any email for more concordance between all of them. The string "SUBMISSION_ACK" was repeated so I made a fusion between both repeated strings.

### DIFF
--- a/locale/es_ES/emailTemplates.xml
+++ b/locale/es_ES/emailTemplates.xml
@@ -12,6 +12,17 @@
   -->
 
 <email_texts locale="es_ES">
+	<email_text key="NOTIFICATION">
+		<subject>Nueva notificación desde {$siteTitle}</subject>
+		<body><![CDATA[Tiene una nueva notificación desde {$siteTitle}:<br />
+<br />
+{$notificationContents}<br />
+<br />
+Enlace: {$url}<br />
+<br />
+{$principalContactSignature}]]></body>
+		<description>El email es enviado a usuarios registrados que han seleccionado recibir notificaciones.</description>
+	</email_text>
 	<email_text key="PASSWORD_RESET_CONFIRM">
 		<subject>Confirmación de cambio de contraseña</subject>
 		<body><![CDATA[Hemos recibido una petición para cambiar su contraseña en {$siteTitle}.<br />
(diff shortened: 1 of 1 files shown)


mtub:
@marcbria , do you mind to have a look?

marcbria:
Hi @t4x0n ,

Thanks for the contrib.

I suspect you sorted chains following en_US order (what is the right way) but it makes difficult for our translator to trace real changes so we won't be able to take a proper look to this till the end of the month (after Heidelberg meeting).

In the other hand I have seen style changes, that need a deeper look. EmailTemplates is a delicate file and the tone need to fit a great variety of contexts... from very formal journals to the more "casual" ones.

With Marco we are making some research to improve the translation workflow. I'm sure we will find a better tool to let us collaborate in a more fruitfully way.

My apologies for this lag and thanks a lot for your help.

Cheers,
m.

t4x0n:
Is not problem @marcbria  ;)  the changes are minimal and I have my own version with more changes for the journals I admin (each editor have his/her own requirements) 

In this case, I add "Estimado/a " because some emails have this and other not, so I just made all emails more similar, and delete the repeated string "SUBMISSION_ACK" 

If you want I can make other PR only with the order changed according to en_US and putting together the repeated string "SUBMISSION_ACK" for you can see this issue, this repeated string is more important that add or not add "Estimado/a "  

I think is not necessary make any improvements in the translation workflow if you have this in order with the translator, I just inform you the real issues, the translate you are doing the last month is very very well... so... proceed as you want.

If you want a new PR only with order changes first I have no problems, and in another, I can submit other PR with more changes, so in this order, you and Marco (the translator I think?) can see more easily others changes.

And of course, if you are in a meeting and have no time don't worry, this Issue and Commit will keeping here until you have more time ;)  

saludos,
Carlos.

mtub:
I think it would be good to separate reordering and updating in two PRs so that reviewing gets a little easier…

marcbria:
It will really help... first updating (to let our translator review the changes) and then reordering.
@carzamora, I know it means time for you but do you think is it feasible?
Thanks,
m.

carzamora:
I made the first PR for reordering to identifying the string "SUBMISSION_ACK" as repeated, now the order is the same as English. This is the PR: https://github.com/pkp/ojs/pull/2130

mtub:
Now that we have the file reordered, could you maybe open a new pull request with your suggested changes to the translation? This PR gets a little messy…

PR #2122, opened 2018-09-25
lib/pkp#4089 remove line breaks from the report titles

ajnyga:
see https://github.com/pkp/pkp-lib/issues/4089


asmecher:
@ajnyga, I don't think this is multibyte-safe -- try using `PKPString::regexp_replace` instead?

ajnyga:
check @asmecher

asmecher:
Thanks as always, @ajnyga!

PR #2125, opened 2018-09-29
Updating OJS Arabic language files before the end of September

vormia:
Arabic locale files have to be updated after the release of OJS 3.1.1-4


mtub:
Thank you so much, @vormia !

PR #2129, opened 2018-10-08
readme updated

lilients:
see also https://github.com/pkp/omp/pull/594


asmecher:
Looks good, but I'd recommend just adding a reference to `getcomposer.org` rather than including the Composer installation instructions themselves. Composer's installation details will depend on the platform and it's best to leave that guidance up to them IMO.

PR #2130, opened 2018-10-08
reordering emailTemplate (new)

carzamora:
This RP is for reordering email templates according to the English version   (derived from https://github.com/pkp/ojs/pull/2119)

Lines 127 and 141 are both "SUBMISSION_ACK" (repeated)

this is the first step, @marcbria if you say me what of the options is the best I can edit the PR and delete one of them or make a new one like a fusion between both.

Something like this:

Gracias por enviarnos su manuscrito &quot;{$submissionTitle}&quot; a {$contextName}. Gracias al sistema de gestión de publicaciones en línea que utilizamos podrá seguir su progreso a través del proceso editorial, iniciando sesión en el sitio web de la revista:

Si tiene cualquier pregunta no dude en contactar con nosotros/as. Gracias por considerar esta revista para publicar su trabajo.

I thought it would be easier to detect changes in blocks, but I see the PR and only some strings are in the block and others appear like changes in lines, but trust me, I downloaded the last version from the master and put it in order.

### DIFF
--- a/locale/es_ES/emailTemplates.xml
+++ b/locale/es_ES/emailTemplates.xml
@@ -12,6 +12,17 @@
   -->
 
 <email_texts locale="es_ES">
+	<email_text key="NOTIFICATION">
+		<subject>Nueva notificación desde {$siteTitle}</subject>
+		<body><![CDATA[Tiene una nueva notificación desde {$siteTitle}:<br />
+<br />
+{$notificationContents}<br />
+<br />
+Enlace: {$url}<br />
+<br />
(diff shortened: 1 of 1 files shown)


carzamora:
@marcbria do not forget this PR, take a look to Lines 127 and 141, both are "SUBMISSION_ACK" (repeated string)

mtub:
Thanks, @carzamora, for aligning this file with the en_US version. I will merge this so that we can move to the next step. As soon as @marcbria says which version of the duplicate email he likes better we can remove the other one.

PR #2131, opened 2018-10-10
pkp/pkp-lib#4131 use function to update 531 assoc_id's in metrics table

ajnyga:
See https://github.com/pkp/pkp-lib/issues/4131

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -2856,7 +2856,7 @@ class Upgrade extends Installer {
 	}
 
 	/**
-	* Update assoc_id for assoc_type 531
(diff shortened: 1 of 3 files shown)


asmecher:
@bozana, could you have a look at this one? (Thanks!)

classes/install/Upgrade.inc.php
```diff
@@ -2855,6 +2855,47 @@ function migrateUserAndAuthorNames() {
 		return true;
 	}
 
+	/**
+	* Update assoc_id for assoc_type 531
```
bozana: One minor thing here: could we add ASSOC_TYPE_SUBMISSION_FILE_COUNTER_OTHER = 531 or so, so that we know all the places that are to be considered if a change for that assoc type is needed.

ajnyga: Thanks, is that what you meant?

PR #2132, opened 2018-10-11
Section policies display

ViRuSTriNiTy:
see https://github.com/pkp/pkp-lib/issues/2638

### DIFF
--- a/classes/submission/form/SubmissionSubmitStep1Form.inc.php
+++ b/classes/submission/form/SubmissionSubmitStep1Form.inc.php
@@ -35,7 +35,7 @@ class SubmissionSubmitStep1Form extends PKPSubmissionSubmitStep1Form {
 
 		// Get section options for this context
 		$sectionDao = DAORegistry::getDAO('SectionDAO');
-		$sectionOptions = array('0' => '') + $sectionDao->getTitles($this->context->getId(), !$canSubmitAll);
+		$sectionOptions = array('0' => '') + $sectionDao->getTitlesByContextId($this->context->getId(), !$canSubmitAll);
 		$templateMgr = TemplateManager::getManager($request);
 		$templateMgr->assign('sectionOptions', $sectionOptions);
 		$templateMgr->assign('sectionId', $request->getUserVar('sectionId'));
--- a/classes/submission/form/SubmissionSubmitStep1Form.inc.php
+++ b/classes/submission/form/SubmissionSubmitStep1Form.inc.php
@@ -38,6 +38,7 @@ class SubmissionSubmitStep1Form extends PKPSubmissionSubmitStep1Form {
 		$sectionOptions = array('0' => '') + $sectionDao->getTitles($this->context->getId(), !$canSubmitAll);
 		$templateMgr = TemplateManager::getManager($request);
 		$templateMgr->assign('sectionOptions', $sectionOptions);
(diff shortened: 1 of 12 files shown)


classes/submission/form/SubmissionSubmitStep1Form.inc.php
```diff
@@ -38,6 +38,20 @@ function fetch($request) {
...
+		$templateMgr->assign('sectionId', $request->getUserVar('sectionId'));
+
+		// Get section policies for this context
+		$sectionPolicies = array();
+		foreach ($sectionOptions as $sectionId => $sectionTitle) {
+			$section = $sectionDao->getById($sectionId);
+
+			$sectionPolicy = $section ? $section->getLocalizedPolicy() : null;
+			$sectionPolicyPlainText = trim(PKPString::html2text($sectionPolicy));
+			if (strlen($sectionPolicyPlainText) > 0)
```
asmecher: (See https://github.com/pkp/pkp-lib/pull/4135/files#r224624538)

locale/en_US/locale.xml
```diff
@@ -101,6 +101,7 @@
 	<message key="section.title">Section title</message>
 	<message key="section.abbreviation">Abbreviation</message>
 	<message key="section.abbreviation.example">(For example, Articles=ART)</message>
+	<message key="section.policy">Section Policy</message>
```
asmecher: Does this also require a PR for OMP?

templates/submission/form/section.tpl
```diff
@@ -14,3 +14,43 @@
...
+
+{foreach from=$sectionPolicies key="sectionPolicySectionId" item="content"}
+	{assign var=sectionPolicyHidden value=false}
+	{if $sectionPolicySectionId != $sectionId}
+		{assign var=sectionPolicyHidden value=true}
+	{/if}
+	{include file="submission/form/sectionPolicy.tpl" sectionId=$sectionPolicySectionId content=$content hidden=$sectionPolicyHidden}
+{/foreach}
+
+<script type="text/javascript">
```
asmecher: We're between two Javascript frameworks -- our own homebrew one, which we're gradually working to replace, and a new vue.js-based one. The old framework discouraged Javascript code in template files, and the new one favours vue templates. Maybe @NateWr has a recommendation for this...?

classes/submission/form/SubmissionSubmitStep1Form.inc.php
ViRuSTriNiTy: See my comments in the linked PR

templates/submission/form/section.tpl
ViRuSTriNiTy: Yep, i definitely need some help here on how to implement the dynamic refresh without a javascript block. It is somewhat a quick and dirty solution.

locale/en_US/locale.xml
ViRuSTriNiTy: > Does this also require a PR for OMP?

I don't think so as the change is part of pkp/ojs and is used in `sectionPolicy.tpl` which is also part of pkp/ojs only.

templates/submission/form/section.tpl
ViRuSTriNiTy: Not relevant anymore due to PR provided by @NateWr

PR #2139, opened 2018-10-17
Fix some small issues reported by LGTM

stweil:


stweil:
When did you start requiring a signed CLA? I think that's a really bad idea. My contributions to free software are under free licenses, here the GPL, so it should not be necessary to read and sign long legal texts.

asmecher:
@stweil, this was added about a month ago. There's a description of the agreement and the reasons we added it here: https://pkp.sfu.ca/2018/10/01/pkps-new-contributor-license-agreement/
I'm happy to talk about it further (both philosophy and practical stuff) if you're unsure about it.

stweil:
So the main reason was to be able to switch from GPL2 to GPL3? Then it would be sufficient to clearly state that new contributions must accept GPL2+ ("GPL 2 or newer"). That's the way how the QEMU project handled this. They also use the Signed-off-by in commits which was introduced initially for the Linux kernel. This is also fine for me – I always sign my contributions to free software.

My main concern with a long CLA is that it is a legal text which is not in my native language, full of details which I might not fully understand. Especially those parts about patents are problematic, as I have patents, but are not free to decide about their use. That's why I only signed a single CLA up to now. It was much shorter and without a patents clause. For all other projects I had to stop contributing. :-(

stweil:
All my contributions to any projects of PKP are licensed according to the GNU General Public License Version 2 or newer.

asmecher:
I am not a lawyer either, but your intentions are clear and fine with me, so I'll take the liberty of merging this without the CLA. If you include a similar note in whatever you contribute going forward that'll remind me. Thanks for contributing!

PR #2142, opened 2018-10-18
pkp/pkp-lib#4161 support generating DMARC compliant mails

tgc:
This provides the config.inc.php part of the changes introduced in https://github.com/pkp/pkp-lib/pull/4165


tgc:
Update the config template with mention of using %s to insert the sitename.

PR #2153, opened 2018-11-02
Update of it_IT for pkp/ojs repo

ztajoli:


mtub:
@ztajoli, thank you very much. There are a few issues with wrong copyright dates, I will take the chance and fix a few things.

PR #2177, opened 2018-11-27
Clean up template without purpose

MarHerUMR:
The ojs/templates/user/profile.tpl was used to pass an additional tab to pkp-lib/templates/user/profile.tpl

Since 1b0e416910a185ddfbb54464df17f4ad975b59a5 the tab "notificationSettings" is included in the profile.tpl from pkp-lib and the {$additionalProfileTabs} are no longer accessed. With that the profile.tpl from ojs does nothing more than including the profile.tpl from pkp-lib.

So the template can be removed.


asmecher:
You're right! Thanks, @MarHerUMR, good catch.

PR #2181, opened 2018-11-27
Syntax issue on string #18

kant:


asmecher:
Merged manually, as this plugin has been moved to its own repository. Thanks, @kant!

PR #2183, opened 2018-11-27
Fix typo on string #23 & 26. Update URL link

kant:


asmecher:
Sorry for the delay in assessing this, @kant! phpMyVisites was merged into Matomo -- see this issue: https://github.com/pkp/pkp-lib/issues/4487

PR #2185, opened 2018-11-27
Fix typos on strings #21 & #23 & #38

kant:
* plus minor proposals


asmecher:
@kant, with apologies for the delay -- which has probably caused the conflict -- there is now a merge issue related to other changes that got merged into that file. Would it be possible for you to reconcile the changes against a newer copy of that XML file?

PR #2188, opened 2018-11-27
Minor fix on string #16

kant:


asmecher:
Merged manually to `stable-3_1_2` instead. Thanks, @kant!

PR #619, opened 2015-09-27
OJS locales updated

mtub:
* de_DE locale: updated
* es_ES locale: minor update from translation server
* pt_PT locale: manually updated plugin translation, see #533
* ru_RU locale: manually updated key, see #579
* usageStats plugin: updated locales (en_US, de_DE)
* locales.xml in registry: sorted alphabetically

PR #2194, opened 2018-11-27
Syntax issues. Typo fixed on string #619

kant:
* Plus minor fixes (proposals)
* update URL on string #119
* on string 452: is mandatory that should be  in english?


asmecher:
With apologies, @kant, this needs to be reconciled against the latest copy of the email templates in the `master` branch. Would you mind taking a look?

PR #2196, opened 2018-11-27
Fix typo on string #29

kant:
* most `http` become `https`


mtub:
@kant I will merge this. But in the future please check if URL changes you propose should also apply to other locales, especially the English one. I'd rather not have different URLs in the individual languages. (Google Scholar link is not https in English file e.g.)

PR #2197, opened 2018-11-28
User id in event log entries of one-click reviews

ptrckb:


dbscripts/xml/upgrade/event_log_oneclickuserid.xml
```diff
@@ -0,0 +1,23 @@
...
+  * Distributed under the GNU GPL v2. For full terms see the file docs/COPYING.
+  *
+  * Add missing user id on event logs from 'one-click' reviews.
+  *
+  -->
+
+<data>
+	<sql>
+		<query>
+			UPDATE event_log el JOIN ( event_log_settings els JOIN review_assignments ra ) ON ( els.log_id=el.log_id AND els.setting_name='reviewId' AND ra.review_id=els.setting_value )
```
asmecher: I think this query may have to be written separately for PostgreSQL and MySQL. Suggestions:
- Break it into separate implementations for the JOIN syntax, e.g.:
  ```
  <query driver="mysql">...</query>
  <query driver="mysqli">...</query>
  <query driver="postgres7">...</query>
  ```
  (The `mysqli` and `mysql` implementations should be the same, just for different driver names.)
- Use decimal numbers instead of hex, but include the constant names for the purposes of grepping the code, e.g.:
  ```
  <query ...>... WHERE el.event_type IN (1073741830, ...)</query><!-- SUBMISSION_LOG_REVIEW_ACCEPT -->
  ```

ptrckb:
@asmecher, yes it is intentional. We do use OJS 3 for our new journals, but the old ones are still on OJS 2.4.

asmecher:
Thanks, @ptrckb -- have you checked whether this affects OJS 3.x as well?

ptrckb:
It does not affect the new log entries with OJS 3 for `ARTICLE_LOG_REVIEW_DECLINE/ACCEPT`. And it seems there is more event logged for new recommendations (`ARTICLE_LOG_REVIEW_RECOMMENDATION`).

However databases already upgraded for OJS 3.x from 2.4 will be missing the user ids on the old events.

asmecher:
Merged -- thanks, @ptrckb!

PR #2198, opened 2018-11-28
Replace generic exception "getStatus() on null"

gustavotonini:
A generic error (Call to a member function getStatus() on null) was shown when articles didn't have a properly section set. This commit intends to log the correct message showing the root cause of the problem.

### DIFF
--- a/classes/install/Upgrade.inc.php
+++ b/classes/install/Upgrade.inc.php
@@ -1056,6 +1056,9 @@ class Upgrade extends Installer {
 				$creatorUserId = $managerUsers->next()->getId();
 			}
 			$article = $articleDao->getById($row['article_id']);
+			if (!$article){
+                                error_log("ERROR: Article ".$row['article_id']." could not be found in the original OJS database. Check if the article and submission exist and if the section_id field is correctly set");
+                        }
 
 			// if it is a remote supp file and article is published, convert it to a remote galley
 			if (!$row['file_id'] && $row['remote_url'] != '' && $article->getStatus() == STATUS_PUBLISHED) {


classes/install/Upgrade.inc.php
```diff
@@ -1056,6 +1056,9 @@ function convertSupplementaryFiles() {
 				$creatorUserId = $managerUsers->next()->getId();
 			}
 			$article = $articleDao->getById($row['article_id']);
+			if (!$article){
+                                error_log("ERROR: Article ".$row['article_id']." could not be found in the original OJS database. Check if the article and submission exist and if the section_id field is correctly set");
```
asmecher: `Article` and `submission` are the same thing -- `Article` is the OJS-specific term, `monograph` is the OMP-specific term, and `submission` is the generic term used in pkp-lib (which both OMP and OJS rely on). I'd suggest changing...
```
Check if the article and submission exist and if the section_id field is correctly set
```
...to...
```
Check if the section_id column is correctly set in the articles table of your original database. All articles must have section_id columns matching entries in the sections table.
```
This will be a helpful addition, thanks!

gustavotonini: But I don't want other users making the same mistake I did: to check articles table for null section ids and forget submission's table.

So, It could be:

`Check if the section_id column is correctly set in the articles/submissions tables of your original database. All articles must have section_id columns matching entries in the sections table.`

asmecher: The migration from OJS 2.x to 3.x effectively renames the `articles` table (OJS 2.x) to `submissions` (OJS 3.x). Before the upgrade runs there shouldn't be a `submissions` table at all; the `articles` table is the only place where the data should need fixing.

PR #2204, opened 2018-12-04
User id in migrated event log entries of one-click reviews

ptrckb:


asmecher:
Thanks, @ptrckb, I've cherry-picked the commit into `master` rather than merging into `ojs-stable-3_1_1`. (We're going to release OJS 3.1.1-5 soon, and then OJS 3.2, and generally speaking we try to avoid database changes between builds.

PR #2207, opened 2018-12-05
pkp/pkp-lib#4215 Fixed missing submission parameter in issueentrysubmissionreviewform::execute hook call

ViRuSTriNiTy:
see pkp/pkp-lib#4215


bozana:
:+1:  thanks @ViRuSTriNiTy !

PR #2215, opened 2018-12-11
pkp/pkp-lib#1908 fix CLOCKSS and LOCKSS manifest display issues

jmacgreg:


pages/gateway/GatewayHandler.inc.php
```diff
@@ -68,7 +68,7 @@ function lockss($args, $request) {
 					'SELECT * FROM issues WHERE journal_id = ? AND year = ? AND published = 1 ORDER BY current DESC, year ASC, volume ASC, number ASC',
 					array($journal->getId(), $year)
 				);
-				$issues = new DAOResultFactory($result, $issueDao, '_returnIssueFromRow');
+				$issues = $issueDao->getPublishedIssuesByNumber($journal->getId(), null, null, $year);
```
asmecher: This makes the above query unnecessary, I think -- you should be able to remove those 4 lines.

asmecher: Also, I think these two lines (fetching `issues` and assigning to the template) are already coded below, so you should be able to remove these as well, I think.

asmecher:
Looks good, thanks, @jmacgreg -- I'll get these forward-ported.

PR #2219, opened 2018-12-12
Correções de traduções

rhuancampos:


asmecher:
@rhuancampos, is this a test PR? It's just a spacing change that's not consistent with the rest of the file.

PR #2249, opened 2019-01-22
Reference linking stable 3 1 2

bozana:
new crossref deposit api


bozana:
Always a different test fails :-P Thus I will now merge this, and hope to see later what is going with the tests... :-\

asmecher:
@bozana, I've noticed Travis failing more often recently -- I don't think it's a change on our part. However, you can restart individual jobs via Travis (it's a little more reliable than rolling 6 dice and hoping for the same result :)

ctgraham:
@bozana, this leaves the production constants pointing to the test APIs.  Is this intended?
https://github.com/pkp/ojs/pull/2249/files#diff-3d6a2901be051c2b5d25cdb4ded5c438R25
https://github.com/pkp/ojs/pull/2249/files#diff-3d6a2901be051c2b5d25cdb4ded5c438R30

bozana:
Ah, good catch @ctgraham! Phew... :+1:

PR #2252, opened 2019-01-23
fix crossref new deposit api production url

bozana:


asmecher:
@bozana, do you see the button in the top right of the job view? Here's what mine looks like...

image

bozana:
@asmecher, hmmm... no, I do not have that button "Restart job" -- that corner is just empty for me :-\

PR #2283, opened 2019-02-12
Greek locale

kstamatis:
Greek translations for the main OJS 3.x user interface


mtub:
Thank you, @kstamatis! There are a few issues that I will tackle after merging this.

PR #2288, opened 2019-02-14
pkp/pkp-lib#4487 removing phpMyVisites plugin

jamshidhashimi:
s. https://github.com/pkp/pkp-lib/issues/4487


asmecher:
Thanks, @jamshidhashimi! Two tweaks:
- It looks like `plugins/importexport/quickSubmit` got included in the PR by accident
- Don't forget to remove the exclusion from `tools/buildpkg.sh`, now that the plugin is gone.

jamshidhashimi:
@asmecher done now. After many tries, finally was able to get rid of the invisible `plugins/importexport/quicksubmit` submodule from the commit history. 

Also removed the plugin from `tools/buildpkg.sh`.

PR #2298, opened 2019-02-19
[Ver. 2.4.8.] Updated counter robots list

meinbrodt:
Added new version of offical counter robots list from https://github.com/atmire/COUNTER-Robots/blob/master/generated/COUNTER_Robots_list.txt


asmecher:
@bozana, I think you've been doing some related work recently... Mind taking a look?

PR #2321, opened 2019-03-06
Update documentation’s URL

sssoz:
Noticed there was a new link there while doing my setup :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ If your PHP version supports built-in development server :
 
     php -S localhost:8000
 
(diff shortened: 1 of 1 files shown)


asmecher:
Good catch, @sssoz, but we probably shouldn't be referring to it as a wiki anymore either -- would you mind adjusting the name of the link from "wiki" to "documentation hub"? Thanks!

sssoz:
@asmecher updated + force pushed

asmecher:
Thanks, @sssoz! Merged.

PR #2332, opened 2019-03-17
Change default db driver

ajnyga:
Since we are only supporting PHP7 now.


asmecher:
Yup, definitely a good idea. Ported to OMP master.

PR #2336, opened 2019-03-20
Add citation_reference tags to GoogleSchorlaPlugin output

ajnyga:
Show unstructured citation_reference metatags with GoogleScholar Plugin. Also add a hook that allows other plugins dealing with references to add structured references if needed.

https://github.com/pkp/pkp-lib/issues/4619

### DIFF
--- a/plugins/generic/googleScholar/GoogleScholarPlugin.inc.php
+++ b/plugins/generic/googleScholar/GoogleScholarPlugin.inc.php
@@ -117,6 +117,24 @@ class GoogleScholarPlugin extends GenericPlugin {
 			}
 		}
 
+		// Citations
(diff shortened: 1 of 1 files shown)


asmecher:
This looks clear to me, @ajnyga, though I haven't tested it with any data. Have you run a few documents through it?

ajnyga:
yes, I did test this with a couple of journals that have saved citations. But the hook there is purely hypothetical, no plugin of course uses it at the moment. Of course if the internal citation handling changes towards structured citations the code needs to be changed. Google would prefer structured citations.

asmecher:
Thanks, @ajnyga! Structured citations would be great but this is a good start.

PR #2343, opened 2019-03-23
Updated templates using the AjaxFormHandler

jonasraoni:
- Templates using the AjaxFormHandler had their content moved into the `<form>` to avoid displaying duplicated content in case there's a submit error (pkp/pkp-lib#4533)


jonasraoni:
@asmecher I was going to ask for a review, but I just saw that one test failed. I'll check how to run the test suite later.

jonasraoni:
Hi @asmecher, I'm resurrecting forgotten things.

tl;dr = Content of forms handled by AjaxFormHandler should be inside the <form> tag to avoid duplicating content after submitting.

This PR failed one test when I created it. I've just merged the latest changes and it passed.

PR #2351, opened 2019-04-03
Replace call to deprecated split() function with explode() for PHP 7

Ianvdl:
Hello

I'm running OJS 2.4.8-4 on Ubuntu 16.04, with PHP 7.0, and noticed that there is a call to split() in the code for step 5 of the setup page. It causes any changes to the existing configuration on that page to fail with a fatal error.

This change replaces the call to split() with a call to explode().


asmecher:
Merged -- thanks, ~~Captain Haddock~~@Ianvdl!

PR #2354, opened 2019-04-09
Remove reference in call to callbackBaseUrl to avoid warning

majaroenne:
We are getting a PHP warning (PHP Warning:  Parameter 2 to rebuildSearchIndex::callbackBaseUrl() expected to be a reference, value given in/lib/pkp/classes/plugins/HookRegistry.inc.php on line 107
) when runing rebuildSearchIndex therefor I have removed the reference in the call to callbackBaseUrl  in tools/rebuildSearchIndex.php


asmecher:
Merged, and will forward-port to `stable-3_1_2` and `master`. Thanks, @majaroenne!

PR #2356, opened 2019-04-10
Added call_hooks for AdditionalFilters and PreResults

Stefanvdb:
@asmecher 
Hi Alec,
In order to be able to show Spellingsuggestion and  Additional Filters from facets, I added two call_hooks to search.tpl. Would be great if they could be added to OJS 3.1.2-1 release, so the Lucene Plugin could function from that release on when it gets released.


asmecher:
Thanks, @Stefanvdb, I'll also cherry-pick this over to `stable-3_1_2` for the 3.1.2-1 release.

PR #2371, opened 2019-04-29
update fr_FR

forgive38:
update locale fr_FR


mtub:
Hi @forgive38 , could you sign the CLA (see above)?

PR #2380, opened 2019-05-08
#4741: Add autocompletion of search phrase to lucene plugin

Stefanvdb:


templates/frontend/components/searchForm_simple.tpl
```diff
@@ -12,7 +12,12 @@
 {if !$currentJournal || $currentJournal->getData('publishingMode') != $smarty.const.PUBLISHING_MODE_NONE}
 	<form class="pkp_search" action="{url page="search" op="search"}" method="post" role="search">
 		{csrf}
-		<input name="query" value="{$searchQuery|escape}" type="text" aria-label="{translate|escape key="common.searchQuery"}">
+		{capture assign="queryFilter"}{call_hook name="Templates::Search::SearchResults::FilterInput" filterName="simpleQuery" filterValue=$query}{/capture}
```
asmecher: Just thinking of a slight change that would allow plugins access to the "default" markup. Something like:
```
{capture assign="queryControl"}
    <input name="query" value="{$searchQuery|escape}" type="text" aria-label="{translate|escape key="common.searchQuery"}">
{/capture}
{call_hook name="Templates::Search::SearchResults::FilterInput" filterName="simpleQuery" filterValue=$query}
{$queryControl}
```
I'm slightly concerned about adding a complex formulation like this to specifically allow the override of a single element in a template -- it's much nicer for the plugin to use, but won't scale well (you can imagine how hard to read the templates will become if this formulation becomes popular in a large plugin ecosystem).

The current alternative is to use Smarty filters to augment/alter template content before it's sent to the browser, then using something like a regexp to mutate it. Have you looked at that option? It doesn't require any hooks to be explicitly added.

(Altering template content on the fly is something for which I've never been satisfied that we have a good solution -- I think Smarty3 might now offer some better options but haven't looked into it!)

Stefanvdb: Hi @asmecher 
how about putting a {block} around it?  I can then override the template in the plugin with a template file like: 
`{extends file='frontend/pages/search.tpl'}
{block name=searchQuery}
//do my stuff 
{/block}
`

asmecher: Yes, this is something I've heard is available in Smarty3 but haven't looked into. It appears to do just what we're looking for. If this works well it could make a good general pattern. Please let me know how it goes!

Stefanvdb: It works well for me in the plugin, got everything working there.

PR #2382, opened 2019-05-08
Correct a tag on pt_BR translation

dancaixeta:
On file locale\manager.xml

The wrong line is:
`<message key="manager.setup.customTagsDescription"><![CDATA[Tags META, inseridas no cabeçalho HTML das páginas do sistema.</message>` 
The correct should be:
`<message key="manager.setup.customTagsDescription"><![CDATA[Tags META, inseridas no cabeçalho HTML das páginas do sistema.]]></message>`

Is missing the tag closure.

PS: This is my first contribution, I'm still learning how to work with Git. Thanks.


asmecher:
Thanks, @dancaixeta -- it looks like the pull request was sent in against the wrong branch, so it includes a bunch of commits that it shouldn't. However, your description of the problem is clear and I'll fix it over here: https://github.com/pkp/pkp-lib/issues/4743

Thanks for contributing!

PR #2396, opened 2019-05-14
pkp/pkp-lib#4754 Skip file when its stage cannot be ascertained

ptrckb:


asmecher:
@ptrckb, thanks, that looks like a clean fix. I've cherry-picked https://github.com/pkp/pkp-lib/commit/92f45e19b251abf228d57e9facb49b014f05d275 to `stable-3_1_2` and `master` from your fork, and added submodule updates all around manually. Will follow up on the issue entry. Thanks again!

PR #2409, opened 2019-05-24
Fix published article cache collision bug

juanmaiquez:
Problem description:
- We have a multi-journal installation. Object cache using memcached.
- We are looking for an article published with id 109 (journal 7).
- A cached article is returned with pub-id::published-id = 109 (journal 571).
- The journals do not match, so null is setted and returned, causing an error 500.

The workaround is to add "if (!empty($returner))" to the return forcing the recovery of the article when the cache fails.

### DIFF
--- a/classes/article/PublishedArticleDAO.inc.php
+++ b/classes/article/PublishedArticleDAO.inc.php
@@ -301,7 +301,7 @@ class PublishedArticleDAO extends ArticleDAO {
 			$cache = $this->_getPublishedArticleCache();
 			$returner = $cache->get($articleId);
 			if ($returner && $journalId != null && $journalId != $returner->getJournalId()) $returner = null;
-			return $returner;
+			if (!empty($returner)) return $returner;
 		}
 
 		$params = $this->getFetchParameters();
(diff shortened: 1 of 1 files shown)


asmecher:
I would recommend disabling memcache support; it has not been maintained for a while and I expect it to be unreliable. I doubt it has much impact on performance either. We'll address it more fully as part of https://github.com/pkp/pkp-lib/issues/3304.

juanmaiquez:
> I would recommend disabling memcache support; it has not been maintained for a while and I expect it to be unreliable. I doubt it has much impact on performance either. We'll address it more fully as part of [pkp/pkp-lib#3304](https://github.com/pkp/pkp-lib/issues/3304).

We will keep in mind you have said about the memcache driver status. Thank you!
However I think this bug affects to every object cache driver implemented in lib/pkp (memcache, opc and xcache), regardless of the choosen config.

asmecher:
Yes, all the object caches are currently not maintained and should be avoided.

Leave `object_cache` set to `none` and `web_cache` set to `Off`. The file-based caching performs well enough for large installations.

PR #2412, opened 2019-05-28
Add lang attribute to language selector

sssoz:
See https://github.com/pkp/pkp-lib/issues/4798

To be added to other custom themes


asmecher:
Thanks, @sssoz! Tagging @NateWr and @Vitaliy-1, in case they need to make similar changes to themes. I'll also back-port this to `stable-3_1_2`.

PR #2426, opened 2019-06-26
pkp/pkp-lib#4021 Use full titles in DC.Title (meta), Title (OAI DC), and title (DataCite/CrossRef) formats

asmecher:
Use full titles (including subtitles) for the following metadata extracts:
- DC.Title (DC meta tag plugin)
- Dublin Core title (OAI)
- DataCite/CrossRef


zuphilip:
Thank you very much for this implementation!

PR #2435, opened 2019-07-23
pkp/pkp-lib#3948: Ensure RSS2 validates with RDF namespace and author/creator distinction

ctgraham:
This addresses strict validation for RSS2: 
* RDF namespace was missing. (pkp/pkp-lib#3948)
* RSS `author` element should be an email or emails; `dc:creator` can be used for names; both should not be used together.


asmecher:
Thanks, @ctgraham! Cherry-picked back to `stable-3_1_2` too.

PR #2436, opened 2019-07-24
var 'total' should be passed by reference to get resume buttons for O…

ohilbig01:
the variable 'total' should be passed/assigned by reference to get the resume button working when using oai sets


asmecher:
Thanks, @ohilbig01! I'll also cherry-pick this over to `stable-3_1_2`, since I think it'll apply cleanly there too.

PR #2448, opened 2019-08-20
pkp/pkp-lib#4844 Editorial report

jonasraoni:
Here most of the code is to extract data... I've decided to add it to the metrics class, because it sounds like a good place, even though the class has another type of task.

While I was updating the email templates, I fixed some things in the pt-BR locale, but didn't read the file until the end =]


classes/statistics/MetricsDAO.inc.php
```diff
@@ -94,9 +94,241 @@ protected function getAssocObjectInfo($submissionId, $contextId) {
...
+		// Access submission constants.
+		import('lib.pkp.classes.submission.Submission');
+
+		// Access decision actions constants.
+		import('classes.workflow.EditorDecisionActionsManager');
+
+		$params = $where = $whereMinimum = [];
+		if($dateStart || $dateEnd) {
+			$where[] = 'ed.date_decided ' . ($dateStart && $dateEnd ? 'BETWEEN ? AND ?' : ($dateStart ? '>= ?' : '<= ?'));
+			$dateStart && $params[] = $dateStart;
```
jonasraoni: I need to check what framework you've adopted to handle databases... To avoid being too slow, I just followed the standard, but I'd like to use at least named parameters here.

classes/statistics/MetricsDAO.inc.php
```diff
@@ -94,9 +94,241 @@ protected function getAssocObjectInfo($submissionId, $contextId) {
...
+	 * Retrieves several statistics about submissions, with sums and averages
+	 * @param $contextId int The context id
+	 * @param $dateStart string The start date filter
+	 * @param $dateEnd string The end date filter
+	 * @param $sectionIds array A list of section IDs to filter the result set
+	 * @return array Returns a dictionary with statistics
+	 */
+	public function getSubmissionStatistics($contextId = null, $dateStart = null, $dateEnd = null, $sectionIds = null) {
+		// Access submission constants.
+		import('lib.pkp.classes.submission.Submission');
```
jonasraoni: I saw there's an autoloader in the application, but didn't check its "requirements". Should I replace these import calls?!

classes/statistics/MetricsDAO.inc.php
```diff
@@ -94,9 +94,241 @@ protected function getAssocObjectInfo($submissionId, $contextId) {
...
+		}
+
+		if (count($sectionIds)) {
+			$where[] = 's.section_id IN (' . substr(str_repeat(',?', count($sectionIds)), 1) . ')';
+			$params = array_merge($params, $sectionIds);
+		}
+
+		$params = array_merge($params, [$dateEnd, $dateStart]);
+
+		$sql = '
```
jonasraoni: The SQL clauses need a good review, I lost edit access to the specification (so I can't see the comments) and I remember of meeting some strange conditions in the database (e.g. published submission with declined as last decision), so I've decided to let some conditions less strict.

locale/pt_BR/emailTemplates.xml
```diff
@@ -16,30 +16,32 @@
 		<subject>Solicitação de atualização de senha</subject>
```
jonasraoni: Maybe to create another PR for this?!
p.s.: I'm not sure I should spend time with locales, but I'm ok to review the rest of the text.

classes/statistics/MetricsDAO.inc.php
asmecher: For now, the DAOs use ADODB and the service classes use Illuminate. We're gradually trying to phase out the ADODB-based code in favour of the Illuminate-based code, but don't worry for now about using ADODB for new code in DAOs.

classes/statistics/MetricsDAO.inc.php
asmecher: The autoloader is in `lib/pkp/includes/functions.inc.php` in the `customAutoload` class. Off the top of my head, yes, you can remove `import` statements that would bring in classes from `lib/pkp/classes` and `classes`.

classes/statistics/MetricsDAO.inc.php
```diff
@@ -94,9 +94,241 @@ protected function getAssocObjectInfo($submissionId, $contextId) {
...
+				ON 1 = 1
+			GROUP BY
+				role_id
+		';
+		return array_reduce($this->retrieve($sql, $params)->GetAll(), function ($data, $row) {
+			$data[$row['role_id']] = $row;
+			return $data;
+		}, []);
+	}
+}
```
NateWr: Phew! SQL is not my strong suit and this looks like some challenging work. Nice job!

When it comes time to port this work to `master`, lets have a conversation about migrating these queries to a `QueryBuilder` class. Our main goals will be to make it easier to adapt the queries for OJS/OMP, and also to create a more reusable query structure, so that plugins (or us in different use cases) can use the QueryBuilder to pull out just the information they want.

locale/en_US/emailTemplates.xml
```diff
@@ -789,4 +789,24 @@ Our decision is to: Decline Submission<br />
...
+	<li>Accepted submissions this month: {$acceptedSubmissions}</li>
+	<li>Total submissions in the system: {$totalSubmissions}</li>
+</ul>
+In addition, you may want to download and review the following reports:<br />
+<ul>
+	<li><a href="{$articlesReportLink}">Articles Report</a></li>
+	<li><a href="{$editorialReportLink}">Review Report</a></li>
+</ul>
+Sincerely,<br />
+{$principalContactSignature}]]></body>
```
NateWr: What do you think about this @jmacgreg?

```
The <a href="{$editorialReportLink}">Journal Health Report</a> can be used to keep track of submissions and editorial activity for {$journalName}. Your key stats for {$month} {$year} are as follows:<br />
<ul>
	<li>New submissions: {$newSubmissions}</li>
	<li>Declined submissions: {$declinedSubmissions}</li>
	<li>Accepted submissions: {$acceptedSubmissions}</li>
	<li>Total submissions in the system: {$totalSubmissions}</li>
</ul>

You can view more stats, such as acceptance and decline rates, in the <a href="{$editorialReportLink}">Journal Health Report</a>. You may also want to download and review the <a href="{$articlesReportLink}">Articles Report</a> and <a href="{$editorialReportLink}">Review Report</a>.

Sincerely,<br />
{$principalContactSignature}]]></body>
```

locale/pt_BR/emailTemplates.xml
NateWr: Typically, we only modify `en_US` during development. That's because most languages have dedicated translators who are approved and managed by @mtub. I'm not sure who that is for `pt_BR`, but in general I'd suggest leaving it off.

If you want, you can create a separate PR, but it would be approved by @mtub.

registry/emailTemplates.xml
```diff
@@ -62,4 +62,5 @@
 	<email key="SUBSCRIPTION_RENEW_INSTL" can_edit="1" can_disable="0" to_role_id="2097152"/>
 	<email key="CITATION_EDITOR_AUTHOR_QUERY" can_edit="1" can_disable="0"/>
 	<email key="REVISED_VERSION_NOTIFY" can_edit="1" can_disable="0" to_role_id="512"/>
+	<email key="STATISTICS_REPORT_NOTIFICATION" can_edit="1" can_disable="0" from_role_id="16" to_role_id="17"/>
```
NateWr: If you change who the email is sent to don't forget to change this value too. :+1:

classes/statistics/MetricsDAO.inc.php
```diff
@@ -94,9 +94,241 @@ protected function getAssocObjectInfo($submissionId, $contextId) {
...
+								ed.decision NOT IN (' . SUBMISSION_EDITOR_DECISION_DECLINE . ', ' . SUBMISSION_EDITOR_DECISION_INITIAL_DECLINE . ')
+								OR ed.decision IS NULL
+							)
+							THEN 0
+						END
+					) / COUNT(0) * 100 AS submission_declined_other_rate
+				FROM submissions s
+				LEFT JOIN edit_decisions first_ed
+					ON first_ed.edit_decision_id = (
+						SELECT ed.edit_decision_id
```
asmecher: You can optimize this to remove the subselect by adding another LEFT JOIN. Try:
```
LEFT JOIN edit_decisions first_ed ON (first_ed.submission_id = s.submission_id)
LEFT JOIN edit_decisions before_first_d ON (before_first_ed.submission_id = s.submission_id AND before_first_id.edit_decision_id < first_ed.edit_decision_id)
...
WHERE before_first_d.edit_decision_id IS null;
```
(In pseudocode form, think "find an edit decision for which an earlier edit decision doesn't exist -- therefore it'll be the first one.")

That ought to be a lot quicker, I think.

classes/statistics/MetricsDAO.inc.php
```diff
@@ -94,9 +94,241 @@ protected function getAssocObjectInfo($submissionId, $contextId) {
...
+								SELECT s.date_submitted
+								FROM submissions s
+								WHERE s.date_submitted IS NOT NULL
+								ORDER BY s.date_submitted
+								LIMIT 1
+							),
+							CURRENT_TIMESTAMP
+						)
+					) + 1 AS count
+			) AS years
```
asmecher: Does this work on PostgreSQL?

locale/pt_BR/emailTemplates.xml
jonasraoni: Thanks, so I'll leave only the changes connected with the task, the rest I'll re-check against /master and create another PR.

classes/statistics/MetricsDAO.inc.php
jonasraoni: I'll take a look in the Illuminate later to see what it offers.

classes/statistics/MetricsDAO.inc.php
jonasraoni: I just checked against MySQL. To port specifically this part against PostgreSQL, replacing YEAR by EXTRACT will suffice.

classes/statistics/MetricsDAO.inc.php
jonasraoni: @asmecher I avoid using auto-increment IDs to find the first/last record when the order is controlled by another column (in this case a combination of stage_id > round > date_decided, right?). So, I'd rather keep it this way for the sake of rightness.

I checked against the database backup that I have and looks like some data was updated from outside (or maybe an update script), and even though the result will be the same, the rule of `smallest id = first record` was broken:
image.

jonasraoni: ps: I can't see much difference here because my `edit_decisions` table has only ~600 records and I'm using an SSD, but the idea is good! I've used it in the past, in another situation, but wouldn't remember to apply it here :D

NateWr:
Replaced by https://github.com/pkp/ojs/pull/2530

PR #2466, opened 2019-09-16
pkp/pkp-lib#4325 include affiliations for all authors

jnugent:
This was a customization for a hosted journal, but per @mgscano's recent comments:

"Being this a requirement for NIH submission, this change can be made public, i.e. included in the next release. This will be a good tool for all medical journals."


asmecher:
Gotcha -- and while I couldn't find confirmation that more than the first affiliation is desired, it does look like the first-affiliation-only requirement has been removed from where it used to appear. Merged and cherry-picked to `stable-3_1_2`.

mgscano:
Hosted client here :)
Well, Alec, actually I never found anything about the affiliations for all authors anywhere, but when I had to redo the pubmed application last year, Sarah Weis from the Data Provider Support Team at NLM had me add them for compliance.
There are a few other requirements, indeed, but I'd rather let them OT in another discussion.

asmecher:
Gotcha, thanks, @mgscano. (For reference -- the first-author-only check that has now been removed per this issue was first introduced here: https://pkp.sfu.ca/bugzilla/show_bug.cgi?id=7774)

PR #2472, opened 2019-09-24
pkp/pkp-lib#5089 fix variable test for section editor count

jnugent:
This trivial fix adjusts the variable in the template so the warning goes away if section editors are assigned.  Although the form is slightly different in master, this fix also works in stable-3_1_2


asmecher:
Thanks, @jnugent, merged and cherry-picked back to `stable-3_1_2`.

PR #2476, opened 2019-09-26
pkp/pkp-lib#5099 add escaping to fix Postgres SQL error

jnugent:
Adds trivial escaping to make this query work on Postgres.


asmecher:
Merged, thanks! Fixed separately in `master` as part of https://github.com/pkp/ojs/commit/b999d3c61e4a8ceaababd30defdbb4e627192e5a.

PR #2485, opened 2019-10-01
Update ru_RU locale after PRs #2478, #2481

ppv1979:


asmecher:
Thanks, @ppv1979! We've started converting locales over to `.po` files -- have a look at the English locale files if you want to get an early look. The change is minor, you'll see, but will permit us to offer Weblate to translators. Watch the PKP blog for details, particularly around our schedule for converting locale files over.

PR #2494, opened 2019-10-11
fr_CA translation

mhvezina:
fr_CA translation reviewed and updated

These are the main changes:
I have :
- reordered message keys based on the en_US equivalent files
- added feminisation of titles/roles (e.g. : "utilisateur-trice", "auteur-e") (but I have kept common terms ("droit d'auteur", "compte utilisateur", "données utilisateur"))
- standardized the use of "téléversement" vs "téléchargement" (and related verb "téléverser" vs "télécharger") for "upload" and "download"
- standardized the use of « some text » instead of "some text"
- standardized the spacing (that is one blank before "?", ":", "!" (better yet would be non-breaking space (&nbsp;) but the translator plugin seems to systematically remove them...)
- removed a few tupples
- removed obsolete message keys (i.e. those keys that didn't have any equivalent in the en_US files)
- fixed a few typos, fixed what seemed to be like wrong translations (sometimes difficult to get the context though...)
- completed what was missing (either new message keys since last translation and/or modified message keys)
- I had to undo the change made for locale/fr_CA/locale.xml (https://github.com/pkp/ojs/commit/a3dd8cd4ebbf83f788fcf51cd51d0bb0bc164058) : In French ("No" stands for Numero. Since it's the first and last letter of the word, and not an abbrevation like, for example, "vol." (that stands for "volume"), it doesn't end with a period (.). Hence correct form in French is : "vol. 4, no 5". Same should be done for locale/fr_FR/locale.xml...


asmecher:
Merged and forward-ported to `master` (https://github.com/pkp/ojs/commit/e142d7aafc69668e53f32f591be8a2613fae0d42) -- thanks, @mhvezina!

PR #2501, opened 2019-10-20
insert "dear" before user name

fgnievinski:


asmecher:
Thnks, @fgnievinski -- as a matter of convention we opted not to include salutations by default, so I'll close this for now, but the journal manager can always add these using the "Prepared Emails" interface. (This is one of those things were there's no one approach that everyone will agree on!)

PR #2504, opened 2019-10-22
two new email templates

fgnievinski:


asmecher:
Merged -- thanks, @fgnievinski! Just for next time, this kind of change is fine to send in a single PR, rather than splitting each language into a different one.

PR #2517, opened 2019-10-24
updated fr_CA translation

mhvezina:
Mainly consist of a few fixes + translation of new keys + removing of obsolete ones (comparison made to the en_US corresponding files)

I wasn't sure if I should have changed the PO-Revision-Dates or not....


asmecher:
I don't think the po-revision-dates are important; those will be automatically updated by 3rd party tools but we aren't using them for anything. Thanks!

PR #2526, opened 2019-11-05
CrossRef Plugin: Add »Select/Unselect All« Capabilities

TobiasWantzen:
To reduce click rate there should be a »Select All« and a »Unselect All« button.
I can’t see, where the `name` attribute (»selectedSubmissions[]«) is generated. This should be changed so that every name change is respected automatically.

### DIFF
--- a/plugins/importexport/crossref/templates/index.tpl
+++ b/plugins/importexport/crossref/templates/index.tpl
@@ -87,8 +87,8 @@
 				}			
 			</script>
 			<p>
-				<input class="pkp_button" type="button" onclick='selectAll()' value="Select All"/>
-				<input class="pkp_button" type="button" onclick='UnSelectAll()' value="Unselect All"/>
+				<input class="pkp_button" type="button" onclick='selectAll()' value="{translate key="common.selectAll"}"/>
+				<input class="pkp_button" type="button" onclick='UnSelectAll()' value="{translate key="common.selectNone"}"/>
 			</p>
 			<form id="exportSubmissionXmlForm" class="pkp_form" action="{plugin_url path="exportSubmissions"}" method="post">
 				{csrf}
--- a/plugins/importexport/crossref/templates/index.tpl
+++ b/plugins/importexport/crossref/templates/index.tpl
@@ -70,7 +70,26 @@
 					// Attach the form handler.
 					$('#exportSubmissionXmlForm').pkpHandler('$.pkp.controllers.form.FormHandler');
 				{rdelim});
+				function selectAll() {
+					var items = document.getElementsByName('selectedSubmissions[]');
+					for (var i = 0; i < items.length; i++) {
+						if (items[i].type == 'checkbox')
+							items[i].checked = true;
+					}
+				}
+
+				function UnSelectAll() {
+					var items = document.getElementsByName('selectedSubmissions[]');
+					for (var i = 0; i < items.length; i++) {
+						if (items[i].type == 'checkbox')
+							items[i].checked = false;
+					}
+				}			
 			</script>
+			<p>
+				<input class="pkp_button" type="button" onclick='selectAll()' value="Select All"/>
+				<input class="pkp_button" type="button" onclick='UnSelectAll()' value="Unselect All"/>
+			</p>
 			<form id="exportSubmissionXmlForm" class="pkp_form" action="{plugin_url path="exportSubmissions"}" method="post">
 				{csrf}
 				<input type="hidden" name="tab" value="exportSubmissions-tab" />
--- a/plugins/importexport/crossref/templates/index.tpl
+++ b/plugins/importexport/crossref/templates/index.tpl
@@ -70,25 +70,17 @@
 					// Attach the form handler.
 					$('#exportSubmissionXmlForm').pkpHandler('$.pkp.controllers.form.FormHandler');
 				{rdelim});
-				function selectAll() {
+				function toggleSelectAll(state) {
 					var items = document.getElementsByName('selectedSubmissions[]');
 					for (var i = 0; i < items.length; i++) {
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks, @TobiasWantzen! One change we'll definitely need is for the English-language text that's currently in the template in your PR to use localized language instead. There are already locale keys called `common.selectAll` and `common.selectNone` (in `lib/pkp/locale/en_US/common.po` for English, for example) so you can use:
```
{translate key="common.selectAll"}
```
...in place of `Select All`.

@NateWr, do you have an opinion/pattern for select all/none buttons?

TobiasWantzen:
> use localized language instead

You’re totally right! I changed that in my code.

NateWr:
Rather than buttons to select or deselect all, the form should have a second submit button which allows someone to export all in one action. That's because we don't show every single submission for export in the list at one time, so a select all button won't actually select all submissions for deposit, just those that are visible.

TobiasWantzen:
@NateWr You’re right, the buttons only select the visible items. But sometimes this is more than enough, because sometimes e.g. CrossRef’s response times are long leading to a timeout. And I could integrate them easily.
To provide a certain amount of flexibility IMHO both use cases make sense: select all visible items AND select all (filtered) items.

NateWr:
The problem is that outside of you and I no one knows that's how it behaves. I want to avoid a situation where an editor thinks they have deposited everything but they haven't.

NateWr:
Maybe we can get around this by using the phrase "Select All Visible". Ideally this would be incorporated into the grid, so that the relationship is clear. But from experience I know that that will be very difficult to do. That's one reason we're transitioning away from the grids.

In the meantime, let's try to put these options into the buttons at the bottom along with the others.

Screenshot_2020-04-23 CrossRef XML Export Plugin

I'll do a quick code review now.

plugins/importexport/crossref/templates/index.tpl
```diff
@@ -70,7 +70,26 @@
...
 				{rdelim});
+				function selectAll() {
+					var items = document.getElementsByName('selectedSubmissions[]');
+					for (var i = 0; i < items.length; i++) {
+						if (items[i].type == 'checkbox')
+							items[i].checked = true;
+					}
+				}
+
+				function UnSelectAll() {
```
NateWr, line 81: `UnSelectAll` should be `unselectAll` here and everywhere else.

plugins/importexport/crossref/templates/index.tpl
```diff
@@ -70,7 +70,26 @@
...
+					for (var i = 0; i < items.length; i++) {
+						if (items[i].type == 'checkbox')
+							items[i].checked = false;
+					}
+				}			
 			</script>
+			<p>
+				<input class="pkp_button" type="button" onclick='selectAll()' value="{translate key="common.selectAll"}"/>
+				<input class="pkp_button" type="button" onclick='UnSelectAll()' value="{translate key="common.selectNone"}"/>
+			</p>
```
NateWr, line 92: Let's combine these methods into one:

```
function toggleSelectAll(state) {
  ...
  items[i].checked = state;
}
```

Then:

```
<button ... onclick="toggleSelectAll(true)">...</button>
<button ... onclick="toggleSelectAll(false)">...</button>
```

plugins/importexport/crossref/templates/index.tpl
```diff
@@ -70,7 +70,26 @@
...
+					var items = document.getElementsByName('selectedSubmissions[]');
+					for (var i = 0; i < items.length; i++) {
+						if (items[i].type == 'checkbox')
+							items[i].checked = false;
+					}
+				}			
 			</script>
+			<p>
+				<input class="pkp_button" type="button" onclick='selectAll()' value="{translate key="common.selectAll"}"/>
+				<input class="pkp_button" type="button" onclick='UnSelectAll()' value="{translate key="common.selectNone"}"/>
```
NateWr, line 91: We prefer to use `<button>` over `<input type="button">`.

TobiasWantzen:
> I'll do a quick code review now.

Thanks, @NateWr, I updated the code according to your review comments.

TobiasWantzen:
> Maybe we can get around this by using the phrase "Select All Visible". Ideally this would be incorporated into the grid, so that the relationship is clear. But from experience I know that that will be very difficult to do. That's one reason we're transitioning away from the grids.

> In the meantime, let's try to put these options into the buttons at the bottom along with the others.

I think, that I’m at the end of my possibilities with these two comments of yours.
- What is hiding behind »grid«, which you want to transitioning away from?
- Where can I insert the buttons into the button line? This is obviously not possible inside this template.

NateWr:
> What is hiding behind »grid«, which you want to transitioning away from?

It's very hard to introduce something to one grid without impacting every other grid. We are migrating, slowly, to ListPanels in our new [UI Library](https://docs.pkp.sfu.ca/dev/ui-library/dev/#/component/ListPanel), but this requires REST API endpoints to be set up for the data that needs to be managed.

> Where can I insert the buttons into the button line? This is obviously not possible inside this template.

The buttons can be added as `<li>` elements to the end of this list: https://github.com/pkp/ojs/blob/fc21fa2a6a944c6f7821432bc220cb461141e8e4/plugins/importexport/crossref/templates/index.tpl#L97-L101

> And how do I add a new localization item? I can only localize »Select all visible« to English an German ...

In the plugin you will find a `locales` directory with `en_US`. You can create a new locale key there and then reference it in a template like this:

```
<button ...>{translate key="generic.plugin.crossref.my.locale.key"}</button>
```

Please only provide the `en_US`, even if you're a native German speaker. Changes for any other language should be done through our [translation tool](https://translate.pkp.sfu.ca/).

PR #2529, opened 2019-11-11
pkp/pkp-lib#5264: Correct Crossref temporary filename construction

ctgraham:
Resolves pkp/pkp-lib#5264


ctgraham:
Suitable to cherry-pick to stable.

PR #2534, opened 2019-11-21
Fix an error when returning contextId in representation.inc.php 

ajnyga:


ajnyga:
ping @NateWr can you check that I fixed this correctly

asmecher:
Merged manually, thanks!

PR #2537, opened 2019-11-25
Move EXCLUDE variable outside buildpkg.sh

marcbria:
Dockerfile is going to be based on buildpkg.sh so moving EXCLUDE variable ouside the file will help in readability of the buildingpkg.sh and Dockerfile but also will let Dockerfile call this remote exclude.list file instead of keeping a replica of the list (reducing sync errors).


asmecher:
Good idea! I would suggest removing the escaping and variable setting from the list -- just making it a list of files/paths -- and then using something like this from the script where it's needed:
```
EXCLUDE=`cat path/to/exclude.list`
```

marcbria:
@asmecher better now?

Disclaimer: Although I have merge rights -that makes translation work much more easier- I won't never use them for code. ;-)

asmecher:
Almost there! I think that'll only work if the script is run from within the `tools` directory, or else the list won't be in the CWD, but I normally don't `cd` there first. You can find out where to look with something like...

```
TOOLDIR=`dirname "${BASH_SOURCE[0]}"`
```
...then...
```
EXCLUDE=`cat ${TOOLDIR}/exclude.list`
```
(This is off the top of my head -- I haven't tested it.)

marcbria:
Added your proposal... but not tested yet.

PR #2540, opened 2019-11-27
Fix mapping of role ids to role i18n keys.

ptrckb:
The function `Application::getRoleNames()` introduced with https://github.com/pkp/ojs/commit/35cb328da2d84723a40168fc1b543c054bc08ec7 can erroneously return the subscription manager role key.

As a result a lot of the email templates on the 'Prepared Email Templates' grid of OJS 3.1.2 have the subscription manager role as sender or recipient (also see [this comment](https://github.com/pkp/pkp-lib/issues/2724#issuecomment-347220421)).

It also applies to `master` even after the role ids of template emails have been fixed with pkp/pkp-lib#2906.


asmecher:
Thanks, @ptrckb -- merged and will cherry-pick forward to `master`.

PR #2545, opened 2019-12-04
resolve function signature warnings

bricas:
Declaration of OpenAIREDAO::setOAI(&$oai) should be compatible with PKPOAIDAO::setOAI($oai)
Declaration of DRIVERDAO::setOAI(&$oai) should be compatible with PKPOAIDAO::setOAI($oai)


asmecher:
Thanks, @bricas! Will also cherry-pick to `stable-3_1_2`.

PR #2547, opened 2019-12-11
pkp/pkp-lib#5177, 5179

thinkbulecount2:
https://github.com/pkp/pkp-lib/issues/5177
https://github.com/pkp/pkp-lib/issues/5179

### DIFF
--- a/plugins/themes/default/styles/objects/article_details.less
+++ b/plugins/themes/default/styles/objects/article_details.less
@@ -55,7 +55,6 @@
 				font-family: @font-heading;
 				font-size: @font-bump;
 				font-weight: @bold;
-				font-weight: @normal;
 			}
 		}
 
--- a/plugins/themes/default/styles/objects/article_details.less
+++ b/plugins/themes/default/styles/objects/article_details.less
@@ -10,17 +10,13 @@
  */
 .obj_article_details {
 
-	h2 {
-		margin: 0;
-	}
-
 	> .page_title {
 		margin: 0;
 	}
 
 	> .subtitle {
 		margin: 0;
-		font-size: @font-lead;
+		font-size: @font-base;
 		line-height: @line-lead;
 		font-weight: @normal;
 	}
@@ -43,15 +39,7 @@
 	}
 
 	.sub_item {
-
 		margin-bottom: @double;
-
-		.label {
-			margin: 0;
-			font-size: @font-sml;
-			font-weight: @normal;
-			color: @text-light;
-		}
 
 		&:last-child {
 			margin-bottom: 0;
@@ -169,7 +157,6 @@
 
 			h2.label {
 				margin: 0;
-				padding: 0;
 				font-family: @font-heading;
 				font-size: @font-sml;
 				font-weight: @normal;
@@ -300,6 +287,19 @@
 
 		.item {
 			padding: @double;
+
+			.label {
+				margin: 0 0 @double;
+				font-family: @font-heading;
+				font-size: @font-bump;
+				font-weight: @bold;
+
+				&.doi .label,
+				&.keywords .label {
+					display: inline;
+					font-size: @font-base;
+				}
+			}
 		}
 
 		.entry_details {
--- a/plugins/themes/default/styles/objects/article_details.less
+++ b/plugins/themes/default/styles/objects/article_details.less
@@ -55,6 +55,7 @@
 				font-family: @font-heading;
 				font-size: @font-bump;
 				font-weight: @bold;
+				font-weight: @normal;
 			}
 		}
 
@@ -155,7 +156,7 @@
 				border-bottom: none;
 			}
 
-			h2.label {
+			.label {
 				margin: 0;
 				font-family: @font-heading;
 				font-size: @font-sml;
@@ -293,12 +294,12 @@
 				font-family: @font-heading;
 				font-size: @font-bump;
 				font-weight: @bold;
+			}
 
-				&.doi .label,
-				&.keywords .label {
-					display: inline;
-					font-size: @font-base;
-				}
+			&.doi .label,
+			&.keywords .label {
+				display: inline;
+				font-size: @font-base;
 			}
 		}
 
--- a/plugins/themes/default/styles/objects/article_details.less
+++ b/plugins/themes/default/styles/objects/article_details.less
@@ -10,6 +10,10 @@
  */
 .obj_article_details {
 
+	h2 {
+		margin: 0;
+	}
+
 	> .page_title {
 		margin: 0;
 	}
@@ -39,7 +43,15 @@
 	}
 
 	.sub_item {
+
 		margin-bottom: @double;
+
+		.label {
+			margin: 0;
+			font-size: @font-sml;
+			font-weight: @normal;
+			color: @text-light;
+		}
 
 		&:last-child {
 			margin-bottom: 0;
@@ -155,7 +167,9 @@
 				border-bottom: none;
 			}
 
-			.label {
+			h2.label {
+				margin: 0;
+				padding: 0;
 				font-family: @font-heading;
 				font-size: @font-sml;
 				font-weight: @normal;
--- a/plugins/themes/default/styles/objects/issue_summary.less
+++ b/plugins/themes/default/styles/objects/issue_summary.less
@@ -14,6 +14,7 @@
 		margin: 0;
 		font-size: @font-base;
 		line-height: @line-base;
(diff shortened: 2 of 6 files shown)


plugins/themes/default/styles/objects/article_details.less
```diff
@@ -10,6 +10,10 @@
  */
 .obj_article_details {
 
+	h2 {
+		margin: 0;
+	}
+
```
NateWr: I think you can remove this because every `<h2>` should have a class of `label` or `subtitle` with its margins set there.

plugins/themes/default/styles/objects/article_details.less
```diff
@@ -39,8 +43,16 @@
 	}
 
 	.sub_item {
+
```
NateWr: Extra line can be removed.

plugins/themes/default/styles/objects/article_details.less
```diff
@@ -39,8 +43,16 @@
...
 		margin-bottom: @double;
 
+		.label {
+			margin: 0;
+			font-size: @font-sml;
+			font-weight: @normal;
+			color: @text-light;
+		}
+
```
NateWr: This block can be removed because these styles are overridden by the other `.label` styles.

plugins/themes/default/styles/objects/article_details.less
```diff
@@ -155,7 +167,9 @@
 				border-bottom: none;
 			}
 
-			.label {
+			h2.label {
+				margin: 0;
+				padding: 0;
```
NateWr: The `padding` here can be removed. The `margin` should stay. :+1:

plugins/themes/default/styles/objects/issue_summary.less
```diff
@@ -10,6 +10,11 @@
  */
 .obj_issue_summary {
 
+	h2 {
+		margin: 0;
+		font-size: @font-lead;
+	}
```
NateWr: It looks like before these changes the issue titles were smaller:

Selection_172

I like your font sizing here, but let's try to keep it the same as it was to minimise disruption for people who may be using a custom child theme.

So change `font-size` to `@font-base`. Also, the `h2` will effect the `line-height` as well, so add a `line-height` prop and set it to `@line-base`.

templates/frontend/components/skipLinks.tpl
```diff
@@ -7,19 +7,18 @@
...
-	<a href="#pkp_content_footer">{translate key="navigation.skip.footer"}</a>
-</div>
+	{if $numAnnouncementsHomepage && $announcements|@count}
+		<a href="#homepageAnnouncements">{translate key="navigation.skip.announcements"}</a>
+	{/if}
+	{if $issue}
+		<a href="#homepageIssue">{translate key="navigation.skip.issue"}</a>
+	{/if}
+{/if}
+<a href="#pkp_content_footer">{translate key="navigation.skip.footer"}</a>
```
NateWr: The `<nav>` wrapper element should be placed here inside of the `skipLinks.tpl` file. See comment here: https://github.com/pkp/pkp-lib/pull/5343/files#r356657148

templates/frontend/objects/article_details.tpl
```diff
@@ -141,25 +141,25 @@
...
 					{translate key="semicolon" label=$translatedKeywords}
-				</span>
+				</h2>
 				<span class="value">
 					{foreach name="keywords" from=$publication->getLocalizedData('keywords') item="keyword"}
 						{$keyword|escape}{if !$smarty.foreach.keywords.last}{translate key="common.commaListSeparator"}{/if}
 					{/foreach}
 				</span>
-			</div>
+			</section>
```
NateWr: We should add `<section>` and `<h2>` to the DOI section that's just above keywords as well.

Once you've done that, you'll need to add some CSS code to bring the size of the DOI/Keywords labels back down to `14px` and make them inline, so that they look like this:

Selection_173

To do that, you'll need to add some CSS that just modify's the `.label` styles for the `doi` and `keywords` sections. Find this block in `article_details.less`:

```css
	.main_entry {

		.item {

			.label {
				margin: 0 0 @double;
				font-family: @font-heading;
				font-size: @font-bump;
				font-weight: @bold;
			}
		}
```

You can extend the `.item` definition like this:

```css
.item {
	
	.label {
		margin: 0 0 @double;
		font-family: @font-heading;
		font-size: @font-bump;
		font-weight: @bold;
	}

	&.doi .label,
	&.keywords .label {
		display: inline;
		font-size: @font-base;
	}
}
```

That will bring just those labels down to size.

templates/frontend/objects/article_details.tpl
```diff
@@ -172,41 +172,41 @@
...
 							{translate key="submission.authorBiography"}
 						{/if}
-					</h3>
+					</h2>
 					{foreach from=$publication->getData('authors') item=author}
 						{if $author->getLocalizedData('biography')}
-							<div class="sub_item">
-								<div class="label">
+							<section class="sub_item">
+								<h2 class="label">
```
NateWr: Because this is nested within another `<h2>`, the heading should be `<h3>`.

templates/frontend/pages/indexJournal.tpl
```diff
@@ -66,7 +68,8 @@
 
 	{* Latest issue *}
 	{if $issue}
-		<div id="homepageIssue" class="current_issue">
+		<div class="current_issue" role="contentinfo">
```
NateWr: The `role="contentinfo" should be removed from this div as well as the announcements and about divs above. This landmark role should only be used once for the page footer.

It probably makes sense to make each of these a `<section>`.

NateWr:
Good work! I've noticed one more thing that wasn't part of the issue but should be. On the article landing page, the authors should be wrapped in a `<section>` and ought to have a `<h2>` for screen readers that labels the section as relating to authors.

Currently the code in `article_details.tpl` looks like this:

```
{if $publication->getData('authors')}
	<ul class="item authors">
		...
	</ul>
{/if}
```

This should look like this:

```
{if $publication->getData('authors')}
	<section class="item authors">
		<h2 class="pkp_screen_reader">{translate key="..."}</h2>
		<ul>
			...
		</ul>
	</section>
{/if}
```

The h2 should say "Authors".

plugins/themes/default/styles/objects/article_details.less
```diff
@@ -155,7 +155,8 @@
 				border-bottom: none;
 			}
 
-			.label {
+			h2.label {
```
NateWr: Sorry, I missed this in the last pass through. The `h2` should be removed here so that the selector is just `.label`. That way, if we have to change the heading level the styles will still apply.

plugins/themes/default/styles/objects/article_details.less
```diff
@@ -286,6 +287,19 @@
...
+				margin: 0 0 @double;
+				font-family: @font-heading;
+				font-size: @font-bump;
+				font-weight: @bold;
+
+				&.doi .label,
+				&.keywords .label {
+					display: inline;
+					font-size: @font-base;
+				}
```
NateWr: These need to be placed up a level so that the `&.doi` attaches itself to the `.item` instead of the `.label`. So this:

```
.item {
  .label {
    &.doi .label {
       ..
    }
  }
}
```

Will produce a CSS selector like this:

```
.item .label.doi .label { ... }
```

But we actually want it to be:

```
.item.doi .label { ... }
```

So this block needs to come up a level, like this:

```
.item {
  ...
  
  .label {
    ...
  }

  &.doi .label,
  &.keywords .label {
    ...
  }
}
```

plugins/themes/default/styles/objects/article_details.less
thinkbulecount2: Just fixed this!

plugins/themes/default/styles/objects/article_details.less
thinkbulecount2: This as well!

thinkbulecount2:
@NateWr just pushed the last (hopefully) few changes! Let me know if anything else looks wrong.

plugins/themes/default/styles/objects/article_details.less
```diff
@@ -55,6 +55,7 @@
 				font-family: @font-heading;
 				font-size: @font-bump;
 				font-weight: @bold;
+				font-weight: @normal;
```
NateWr: This line should be removed. The labels should remain bold.

PR #2560, opened 2019-12-18
Change ORCID plugin setting name during upgrade

ajnyga:
See https://github.com/pkp/orcidProfile/pull/83#pullrequestreview-333658402

### DIFF
--- a/dbscripts/xml/upgrade/3.2.0_update.xml
+++ b/dbscripts/xml/upgrade/3.2.0_update.xml
@@ -11,15 +11,15 @@
   * 3.2.0.0 general updates.
   -->
 <data>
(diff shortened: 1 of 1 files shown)


ctgraham:
The Plugin's `register` method hooks a bunch of "Installer::postInstall" hooks for this purpose:
https://github.com/pkp/pkp-lib/blob/68ae864c9be4c824aaf57b6ca1ec3969a9ea358f/classes/plugins/Plugin.inc.php#L90-L120

Leveraging one of these (probably `$this->getInstallSchemaFile()` and `$this->updateSchema()`) , or just hooking "Installer::postInstall" directly is probably the best method.

PR #2562, opened 2019-12-19
Fully updating PT-BR translation

Lu-Brito:
Fully updating PT-BR translation


asmecher:
@Lu-Brito, there appears to be a conflict with `locale/pt_BR/manager.po` in the master branch; can you see if you can resolve it?

PR #2589, opened 2020-01-25
pkp/pkp-lib#5439 fix calls to new for PHP7 compatibility

jnugent:
tweaks to calls to new in the SimplePie class.


asmecher:
@jnugent, is there an upstream version that fixes this without too many other implications?

jnugent:
@asmecher The version in stable-3_1_2 might, but I haven't tried it.

asmecher:
All right, normally I wouldn't merge a code change to a third-party library, but this is 2.x we're talking about...

PR #2611, opened 2020-02-06
fix reassign DOIs

ajnyga:
https://github.com/pkp/pkp-lib/issues/5473


ajnyga:
@asmecher can you check the tests and maybe restart?

asmecher:
Sorry for the wait -- restarted!

ajnyga:
thanks, tests worked!

ajnyga:
Also @asmecher, I went with option B here: https://github.com/pkp/pkp-lib/issues/5473#issuecomment-582939890

Even if in OJS the DOI does not change with versions, this is still safe, because only the DOI from the latest version is registered and the Crossref plugin there, I think, always just registers one DOI per submission

In OPS the Crossref plugin will register all DOIs within a submission, because versions are supported.

asmecher:
Excellent, thanks, @ajnyga!

PR #2621, opened 2020-02-12
pkp/pkp-lib#5482 Permit newly-assigned editors to edit metadata (tests only)

asmecher:


asmecher:
(Will manually update submodules.)

PR #2628, opened 2020-02-19
add CLI flags for opting out of embedded files in xml

quoideneuf:


plugins/importexport/native/NativeImportExportPlugin.inc.php
```diff
@@ -445,6 +448,38 @@ function executeCLI($scriptName, &$args) {
...
+		$newArgs = [];
+		$opts = [];
+		$sticky = null;
+		foreach ($args as $arg) {
+			if ($sticky) {
+				$opts[$sticky] = $arg;
+				$sticky = null;
+				continue;
+			}
+			if (!starts_with($arg, '--')) {
```
asmecher: Is `starts_with` an actual function, or is that pseudocode?

plugins/importexport/native/NativeImportExportPlugin.inc.php
```diff
@@ -445,6 +448,38 @@ function executeCLI($scriptName, &$args) {
 		$this->usage($scriptName);
 	}
 
+	/**
+	 * Pull out getopt style long options.
```
asmecher: What about actually using PHP's `getopt` function?

plugins/importexport/native/NativeImportExportPlugin.inc.php
quoideneuf: It is a helper provided by the laravel framework. I could take it out...

plugins/importexport/native/NativeImportExportPlugin.inc.php
quoideneuf: I didn't think it could work without changing the existing CLI api. It seems like getopt only works if all the arguments conform to the getopt syntax.

asmecher: Gotcha, just a thought. It might be nice to move this toolset up to the parent class but that can wait for another use case. :+1:

plugins/importexport/native/NativeImportExportPlugin.inc.php
asmecher: Ah, OK, I just hadn't run across it. That's much more readable than the `substr` calls I've been relying on for the same thing.

PR #2634, opened 2020-02-20
#fixes 5345

quoideneuf:


asmecher:
I extended the length of that field to 40 characters to match the UI max-length constraint a few weeks ago (15411bf3d1fd943e3b97dc93adc639457a94d3ac), so this should be changed to match that too; also, I suspect that people are using the number field for not-strictly-numeric data (e.g. roman numerals or ranges) for the moment I'd suggest leaving it to accept non-numeric data.

In fact, since the UI constraint should already prevent data with more than 40 characters from going in, and since anything more than that will be chopped by the DB (e.g. in case someone uses a web dev tool to remove the `max-length`, I'm not sure any further work is needed.

asmecher:
(It might be good to check the original issue's description of the steps to replicate the issue to make sure we're [I'm] not missing something!)

PR #2635, opened 2020-02-22

vormia:
Adding some missing keys to Arabic locale files


vormia:
I have to update this repo

PR #2636, opened 2020-02-22

vormia:
Adding some new keys to Arabic locale files


asmecher:
Merged, thanks, @vormia!

PR #2655, opened 2020-02-28
Translation updates

asmecher:


asmecher:
@kant, just a heads-up that some of your work came in as es_AR rather than es_ES. Was that intentional?

kant:
@asmecher I tried to fill the gaps (since weblate or the system that controls the translations points that some blanks were present), so, the goal was to fill all the strings unstranslated. If this a burden in the path, please, wipe out, move or... your scrutiny will be welcome.

PR #2668, opened 2020-03-07
Consider upgrade to current scheme

klausru:
Consider upgrade to current scheme

### DIFF
--- a/plugins/importexport/crossref/CrossrefExportDeployment.inc.php
+++ b/plugins/importexport/crossref/CrossrefExportDeployment.inc.php
@@ -18,10 +18,10 @@
  */
 
 // XML attributes
-define('CROSSREF_XMLNS' , 'http://www.crossref.org/schema/4.3.6');
+define('CROSSREF_XMLNS' , 'http://www.crossref.org/schema/4.4.2');
 define('CROSSREF_XMLNS_XSI' , 'http://www.w3.org/2001/XMLSchema-instance');
-define('CROSSREF_XSI_SCHEMAVERSION' , '4.3.6');
-define('CROSSREF_XSI_SCHEMALOCATION' , 'https://www.crossref.org/schemas/crossref4.3.6.xsd');
+define('CROSSREF_XSI_SCHEMAVERSION' , '4.4.2');
+define('CROSSREF_XSI_SCHEMALOCATION' , 'https://www.crossref.org/schemas/crossref4.4.2.xsd');
 define('CROSSREF_XMLNS_JATS' , 'http://www.ncbi.nlm.nih.gov/JATS1');
 define('CROSSREF_XMLNS_AI' , 'http://www.crossref.org/AccessIndicators.xsd');
 


asmecher:
Tagging @AhemNason (on behalf of the metadata WG) as they might know whether this has other implications that would need tweaking...

AhemNason:
I'll look into this. I sort of assume this wouldn't have changed anything too dramatic, but better safe than sorry.

AhemNason:
Ok, so, just checking the schema at [https://gitlab.com/crossref/schema/-/blob/master/schemas/crossref4.4.2.xsd](https://gitlab.com/crossref/schema/-/blob/master/schemas/crossref4.4.2.xsd)

> 4.3.0 (PDF) 7/24/2012 added namespace and import for fundref.xsd

This would have implications, I'd think... although only in that we wouldn't be supporting this in our plugin(s) atm. 

> 4.3.0 (PDF) 	9/20/11 Added <noisbn> to book_series_metadata, book_set_metadata, and proceedings_series_metadata

This would, ideally, have implications for OMP if the plugin were in _good shape_.  This includes: 

>4.2.0 (BDR) 10/25/07
> Changed structure of book metadata deposits so that
>      a) book series are handled seperately from books that are not series
>      b) book sets are properly supported
>     This is accomplished by changing the top book structure to be a choice
>      of book_metadata, book_series_metadata, or book_set_metadata
>     book_metadata is backwards compatible provided you are depositing books
>      that are not part of a series or a set
>     A minimum of 1 ISBN is required for each book volume. Previously ISBN was
>      optional
> 
>     In conjunction with the book changes above, volume and issn were removed
>      from proceedings_metadata, report-paper_metadata, and standards_metatdata
>      because they are now in series_metadata.

However, almost everything here is additions instead of changing anything we'd be using currently. I think it's likely safe to update but I'd recommend us running some local deposit tests just to make sure. Happy to do that with @jmacgreg if there's a will.

AhemNason:
Oh, I'm a dummy... most of these are pre 4.3.6 anyway... I'd still recommend a test. And, I think, a bit of an overhaul of metadata assignments in OMP, apparently. I apologise. If it's any consolation, I just turned my coffee maker on but forgot to put my mug under it. Very alert.

asmecher:
See also https://github.com/pkp/ojs/pull/2669#issuecomment-596689544, which would be dependent on this update.

klausru:
My uploads from OJS as XML to https://doi.crossref.org, containing the `journal_issue` -> `titles` -> `title` tags worked quite fine.
However, special attention should be given to the `maxlength `of `volume` (aka `issue`) see here: https://forum.pkp.sfu.ca/t/raising-an-issue-char-length-of-volume/58437/10

klausru:
I added another commit in order to export number and volume regardless of the setting to show or not in the issue settings: https://github.com/pkp/ojs/pull/2669/commits/879cb729d30efb64b3ff0d6dc60a1d37ae2372df

asmecher:
(@AhemNason, I could also use your feedback on https://github.com/pkp/ojs/pull/2668#issuecomment-596747855!)

PR #2678, opened 2020-03-12
pkp/pkp-lib#5623 Escape raw citations

Vitaliy-1:
Escape, leave allowed tags.


asmecher:
Used `escape` instead of `strip_unsafe_html`: https://github.com/pkp/ojs/commit/36c3502ee35d17a70e27f3e9c85f0629940cca77

PR #2679, opened 2020-03-12
[stable-3_1_2][OJS]#5623 escape raw citations

Vitaliy-1:


asmecher:
My mistake, @Vitaliy-1 -- this should actually be passed through `escape` rather than `strip_unsafe_html`. Corrected and will forward-port to `master`. Do you know if any themes need this adapted?

PR #2694, opened 2020-03-25
Add save for review forms, initial work 

ajnyga:
Initial work for https://github.com/pkp/pkp-lib/issues/3698, not ready for merge yet

### DIFF
--- a/pages/reviewer/ReviewerHandler.inc.php
+++ b/pages/reviewer/ReviewerHandler.inc.php
@@ -101,16 +101,11 @@ class ReviewerHandler extends PKPReviewerHandler {
 	 */
 	public function getReviewForm($step, $request, $reviewerSubmission, $reviewAssignment) {
 	    switch ($step) {
-	        case 1:
-	        	import("lib.pkp.classes.submission.reviewer.form.PKPReviewerReviewStep1Form"); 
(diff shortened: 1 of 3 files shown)


pages/reviewer/ReviewerHandler.inc.php
```diff
@@ -95,6 +95,24 @@ function _validateAccessKey($request) {
 			$this->user = $user;
 		}
 	}
+
+	/**
+	 * @copydoc PKPReviewerHandler::getReviewForm()
+	 */
+	public function getReviewForm($step, $request, $reviewerSubmission, $reviewAssignment) {
```
asmecher: This should only contain the `case 3`, I think, and the class should be `ReviewerReviewStep3Form`, not `PKPReviewerReviewStep3Form`.

ajnyga: ah yes.

If I just add case 3 then I should probably call for the parent function as well. I figured that almost the same amount of code if I just add all three cases?

asmecher: I agree, not a big deal either way; I'm slightly in favour of just overriding the one we care about in the application.

ajnyga: how about now?

asmecher:
Merged manually (merge conflict)

PR #2699, opened 2020-03-30
Initial work for assigning sub editors to categories

ajnyga:


ajnyga:
ping @asmecher  and @NateWr 

All basic functions are now ready.
1. You can assign subeditors to categories
2. Choose the category in step 1 of the submission form
3. When submission is ready, subeditors are assigned to the submission based on the selected categories.

There are a couple of duplicate functions in the CategoriesForm and SectionForm, so we could add already now a base SubmissionGroupForm. What do you think? If yes, what should be the location?

Moving the selection of categories to step 1 had one problem. 
The Categories were handled in PKPSubmissionMetadataFormImplementation.inc.php this is called from step 3. However, also quicksubmit uses this, so I can not just remove it from PKPSubmissionMetadataFormImplementation.
So I changed the conditional here a little: https://github.com/pkp/pkp-lib/pull/5684/files#diff-46c7561032ef6c2981126d705c6f66a1R231
I think this works, because in Quicksubmit, you can not go back to the form later like you can in the actual submission form. We could of course remove it totally from PKPSubmissionMetadataFormImplementation and add handling for Categories separately to Quicksubmit. What do you think?

ajnyga:
tests passed

ajnyga:
If everything looks good above, I can do a pr for OMP and OPS.

asmecher:
Thanks, @ajnyga! Is there an issue open for this yet? If not, please open one, and prefix the commits with the issue number; then add links for the PRs to the issue. Then we can hash over technical details there; otherwise it can be easy to lose those discussions between several PRs.

> There are a couple of duplicate functions in the CategoriesForm and SectionForm, so we could add already now a base SubmissionGroupForm. What do you think? If yes, what should be the location?

Hmm, I suppose `lib/pkp/classes/context`, which is where `Category` and `PKPSection` live. Down the road, I could see moving all three of those classes to `lib/pkp/classes/submissionGroup`, but that's a change more suitable for 3.3 than 3.2.1.

Don't forget to consider the upgrade process -- the table rename, column rename, etc!

> I think this works, because in Quicksubmit, you can not go back to the form later like you can in the actual submission form. We could of course remove it totally from PKPSubmissionMetadataFormImplementation and add handling for Categories separately to Quicksubmit. What do you think?

I think it's best to keep the main codebase as clean as possible and move that consideration into the QuickSubmit plugin. It needs some UI/UX work already, and it would be easy to refactor it and forget that there was a work-around in the main codebase to support it.

dbscripts/xml/upgrade/3.2.1_preupdate_subeditor_categories.xml
```diff
@@ -0,0 +1,26 @@
...
+<data>
+	<sql>
+		<dropindex table="section_editors" index="section_editors_pkey" />
+		<dropindex table="section_editors" index="section_editors_context_id" />
+		<dropindex table="section_editors" index="section_editors_section_id" />
+		<dropindex table="section_editors" index="section_editors_user_id" />
+		<rename table="section_editors" to="subeditor_submission_group" />
+		<rename table="subeditor_submission_group" column="section_id" to="assoc_id" />
+		<query>ALTER TABLE subeditor_submission_group ADD assoc_type BIGINT<query>
+		<query>UPDATE subeditor_submission_group SET assoc_type = 530</query>
```
asmecher, line 24: Add a comment to the end of the line: `<!-- ASSOC_TYPE_SECTION -->` (so this can be grepped)

dbscripts/xml/upgrade/3.2.1_preupdate_subeditor_categories.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE data SYSTEM "../../../lib/pkp/dtd/xmlData.dtd">
+
+<!--
+	* dbscripts/xml/upgrade/3.2.1_preupdate_subeditor_categories.xml
```
asmecher, line 5: Looks good, but needs adding to `dbscripts/xml/upgrade.xml`.

PR #2720, opened 2020-04-23
add two lines to the config file to allow suppressing the SMTP authentication in PHPMailer

HongPong:
This is one side of the modification. The other side is in pkp-lib.

### DIFF
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -318,7 +318,8 @@ allowed_html = "a[href|target|title],em,strong,cite,code,ul,ol,li[class],dl,dt,d
 ; smtp_password = password
 
 ; Enable suppressing verification of SMTP certificate in PHPMailer
-: smtp_suppress_cert_check = On
+; Note: this is not recommended per PHPMailer documentation
+; smtp_suppress_cert_check = On
 
 ; Allow envelope sender to be specified
 ; (may not be possible with some server configurations)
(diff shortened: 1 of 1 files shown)


HongPong:
See issue https://github.com/pkp/pkp-lib/issues/5781

config.TEMPLATE.inc.php
```diff
@@ -317,6 +317,9 @@
 ; smtp_username = username
 ; smtp_password = password
 
+; Enable suppressing verification of SMTP certificate in PHPMailer
+: smtp_suppress_cert_check = On
```
asmecher, line 321: Minor typo: that colon at the start of the line should be a semicolon.

config.TEMPLATE.inc.php
```diff
@@ -317,6 +317,9 @@
 ; smtp_username = username
 ; smtp_password = password
 
+; Enable suppressing verification of SMTP certificate in PHPMailer
```
asmecher, line 320: I'd suggest adding a note that this is not recommended as per the PHPMailer documentation (see under "Updating CA certificates" at https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting).

HongPong:
Ahh thank you sorry i missed the correct character, good eye - 👁️ 
I added a better notification that this is not the recommended config, let me know if you think the phrase should change.

asmecher:
Both good, thanks!

PR #2737, opened 2020-05-08
pkp/pkp-lib#5843 User object in payment records for past users null

ewhanson:


asmecher:
@ewhanson, could you move the pkp-lib submodule update into its own commit? Generally we keep the submodule updates in their own commits so that the PR can be cherry-picked in case of a merge conflict (which is not infrequent).

ewhanson:
Hi @asmecher, I've separated out the pkp-lib update into its own submodule commit. Let me know if any other changes are needed.

asmecher:
Merged manually!

PR #2739, opened 2020-05-08
Citations doesn't show

hsh01:
Citations doesn't show if citationRaw is empty, but existing parsed citations. I think here is mistake.

### DIFF
--- a/pages/article/ArticleHandler.inc.php
+++ b/pages/article/ArticleHandler.inc.php
@@ -214,7 +214,7 @@ class ArticleHandler extends Handler {
 		));
 
 		// Citations
-		if ($publication->getData('citationsRaw')) {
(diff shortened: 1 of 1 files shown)


asmecher:
@hsh01, are you working with content that was migrated from a previous version of OJS? What version are you currently using?

hsh01:
Yes, I'm constantly updating to the latest version, and I'm migrating all the content. I'm currently using version 3.2.0-3. I usually upload ready-made issues using the import plugin. I used to add an additional citationRaw field for importing bibliography, but now that you have added the ability to import citations, there is no need to add citationsRaw, and it remains empty. Therefore, the bibliography is not displayed on the article page, although it is in the database. Here, the $parsedCitations variable remains empty if citationsRaw is empty.

PR #2745, opened 2020-05-13
[OJS] #5801 remove duplicates and referenced submission from rec…

Vitaliy-1:
…ommend by author results

1. Not sure why array_unique was used on submissions; can this array contain duplicate values?
2. On line 89 array_filter is used. According to the [example](https://www.php.net/manual/en/function.array-filter.php#example-6248), I can't figure out situation when it may be useful. Left it as is but can remove if necessary.

### DIFF
--- a/plugins/generic/recommendByAuthor/RecommendByAuthorPlugin.inc.php
+++ b/plugins/generic/recommendByAuthor/RecommendByAuthorPlugin.inc.php
@@ -84,9 +84,15 @@ class RecommendByAuthorPlugin extends GenericPlugin {
 				$publication = Services::get('publication')->get($publicationId);
 				return $publication->getData('status') == STATUS_PUBLISHED ? $publication->getData('submissionId') : null;
 			}, array_unique($publicationIds));
-			$foundArticles = array_merge($foundArticles, array_unique($submissionIds));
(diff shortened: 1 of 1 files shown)


asmecher:
@Vitaliy-1, I think `array_unique` is necessary so that submissions with multiple publications don't get fetched twice.

The `array_filter` call was added a part of a fix for pkp/pkp-lib#5683; it removes entries from the array that correspond to publications that aren't published.

asmecher:
Just points for style, but I'd suggest using a lambda function in the existing `array_filter` call to kill two birds with one stone -- remove `null` entries, and the current submission ID.

Vitaliy-1:
Thanks!

I suspected that it may be related to another issue, although I can't reproduce it. Did you mean something like in the additional commit? Also, I don't understand why some checks are failing.

asmecher:
@Vitaliy-1, yes, the additional commit looks good. Don't worry about the tests -- they've been a little less reliable lately; I'll try to take a look. But that's not related to this PR.

asmecher:
In other words, if you're ready, please go ahead and merge!

Vitaliy-1:
@asmecher, I'm afraid I don't have write access to ojs repo, only to shared library.

asmecher:
@Vitaliy-1, I've just sent you an invite!

PR #2750, opened 2020-05-14
[OJS] #5061 Fix default locale priority for getGivenName and getFamilyName

defstat:


plugins/importexport/crossref/filter/ArticleCrossrefXmlFilter.inc.php
```diff
@@ -327,7 +358,26 @@ function createComponentListNode($doc, $submission, $componentGalleys) {
...
 
+	/**
+	 * Get's the Iso639 representation of the language code that 
+	 * localised parameters are using. For example if en_US is entered
+	 * then en is returned
+	 * 
+	 * @var string $languageCode It is the code that localised parameters are using
+	 * @return string|null the ISO 639 code
+	 */
+	function getIso639LanguageFromLanguageCode($languageCode) {
```
asmecher, line 369: I think there are already support functions to do this in `PKPLocale`.

PR #2766, opened 2020-05-22
Remove empty file

Mandroide:
I don't get why this file was created in the first place.


asmecher:
@Mandroide, I think that's just to prevent a directory index from being displayed on servers with directory indexing enabled (although directory indexes should definitely be disabled) -- is it causing problems for you?

Mandroide:
Not at all. Then this PR is closed

asmecher:
Thanks, @Mandroide! (Always happy to have scrutiny on the codebase.)

PR #2807, opened 2020-06-12
Fix path

Mandroide:
I noticed that those path were different from the ones like the English version.

I don't think that `${indexUrl}{$contextPath}` = `${indexUrl}/{$journalPath}`


locale/el_GR/default.po
```diff
@@ -145,10 +145,10 @@ msgstr ""
...
+msgstr "Εάν ενδιαφέρεστε να υποβάλετε την εργασία σας στο συγκεκριμένο περιοδικό, ανατρέξετε στη σελίδα &lt;a href=\"{$indexUrl}{$contextPath}}/about\"&gt;Σχετικά με το περιοδικό&lt;/a&gt; για τις πολιτικές κάθε ενότητας του περιοδικού, καθώς και στη σελίδα &lt;a href=\"{$indexUrl}{$contextPath}}/about/submissions#authorGuidelines\"&gt;Οδηγίες προς τους Συγγραφείς&lt;/a&gt;. Οι Συγγραφείς πρέπει να κάνουν &lt;a href=\"{$indexUrl}{$contextPath}}/user/register\"&gt;εγγραφή&lt;/a&gt; στο περιοδικό πριν από την υποβολή ή, σε περίπτωση που έχουν ήδη εγγραφεί, μπορούν να κάνουν &lt;a href=\"{$indexUrl}{$contextPath}}/login\"&gt;Σύνδεση&lt;/a&gt; και να ξεκινήσουν τη διαδικασία υποβολής σε 5 βήματα."
```
asmecher, line 151: I think this introduces a typo (see the `}}` close).

Mandroide: Fixed. Thank!

Mandroide: Is it `{$indexUrl}{$contextPath}` outdated? I don't know if `{$indexUrl}/{$journalPath}` is the true way, because I noticed that in the Spanish version the webpage states that it doesn't have a context when I tried to access the default information for authors or readers.

asmecher: Thanks, @Mandroide! It should definitely be `$contextPath` rather than `$journalPath`, and the `$indexUrl` should include a trailing slash (at least on my dev system, which is pretty typical). Does the URL it's generating appear to include a typo?

Mandroide: Neither {$indexUrl}{$contextPath} nor {$indexUrl}/{$journalPath} seem to include a typo.

asmecher: I'm not sure I'm following what you said earlier...
> In the Spanish version the webpage states that it doesn't have a context when I tried to access the default information for authors or readers.

Does the link to the Spanish URL look different from the link to the English URL?

Mandroide: Oh I see. In the English Version `{$indexUrl}{$contextPath}` shows a Not Found on the server error, while in the Canada French Version `{$indexUrl}/{$journalPath}` the generated URL is working. 

I don't understand why in the Spanish version appears that it cannot find a context, if the way it generates the URL is the same as Canade French Version.

Mandroide:
I guess what I should really replace is `{$indexUrl}{$contextPath}` with `{$indexUrl}/{$journalPath}`

asmecher:
> I guess what I should really replace is `{$indexUrl}{$contextPath}` with `{$indexUrl}/{$journalPath}`

No, I think `{$indexUrl}{$contextPath}` is right (at least speaking about 3.2.0 or newer). The variables are provided in https://github.com/pkp/pkp-lib/blob/3_2_0-0/classes/services/PKPContextService.inc.php#L314..L325.

Mandroide:
> > I guess what I should really replace is `{$indexUrl}{$contextPath}` with `{$indexUrl}/{$journalPath}`
> 
> No, I think `{$indexUrl}{$contextPath}` is right (at least speaking about 3.2.0 or newer). The variables are provided in https://github.com/pkp/pkp-lib/blob/3_2_0-0/classes/services/PKPContextService.inc.php#L314..L325.

OK

asmecher:
Merged, and I'll cherry-pick to other branches as necessary. Thanks, @Mandroide!

PR #2834, opened 2020-07-09
Add timezone %Z to all time_formats

maifeld:
[https://forum.pkp.sfu.ca/t/ojs3-timestamps-fixed-in-utc-and-display-no-timezone/60808/7?u=maifeld](url)

We have users across all timezones in the United States.  Seeing an unlabelled UTC time while sitting in their own timezone has caused some confusion.

### DIFF
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -58,9 +58,9 @@ time_zone = "UTC"
 date_format_trunc = "%m-%d"
 date_format_short = "%Y-%m-%d"
 date_format_long = "%B %e, %Y"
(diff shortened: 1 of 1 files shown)


asmecher:
@maifeld, I'm ambivalent about adding this as a default -- and noting your previous forum discussions, I think this would be a better general resolution: https://github.com/pkp/pkp-lib/issues/6082

fradeve:
Also discussed here: https://forum.pkp.sfu.ca/t/ojs3-timestamps-fixed-in-utc-and-display-no-timezone/60808

PR #2843, opened 2020-07-31
Fix page not found

Mandroide:
Since the file  _README-GIT.md_ not longer exists, I updated it to the main _README.md_


asmecher:
Thanks, @Mandroide!

PR #2844, opened 2020-07-31
Add link to permissions explanation to README

bondjimbond:
The README's "change permissions" section assumes quite a bit of technical knowledge that the user installing OJS might not have. There is a decent outline of the recommended permissions configuration in the Forum; this change simply adds a link there for further information.


asmecher:
@bondjimbond, thanks -- I'll hold off merging this because we're in the process of moving the forum-based FAQ to the PKP Documentation Hub (https://docs.pkp.sfu.ca/), but I'll adapt this change when that's ready.

PR #2862, opened 2020-08-19
Fix links

Mandroide:
Fix links for [doxygen](https://pkp.sfu.ca/ojs/doxygen/master/html/) documentation


asmecher:
Merged -- thanks, @Mandroide! I'll make sure this gets cherry-picked/ported to other repos.

PR #2873, opened 2020-09-22
pkp/pkp-lib#5772 Move from PHP to JSON serialization

Vitaliy-1:


classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
```diff
@@ -41,4 +43,114 @@ public function down() {
...
+						Capsule::table($tableName)->where('setting_name', $propName)->get()->each(function ($row) use ($tableName, $propName) {
+							$this->_toJSON($row, $tableName, ['setting_name', 'locale'], 'setting_value');
+						});
+					}
+				}
+			}
+		}
+
+		// Convert settings where only setting_type column is available
+		$tables = Capsule::connection()->select('SHOW TABLES');
```
asmecher, line 74: Have you tried this with PostgreSQL? I don't think PostgreSQL supports `SHOW TABLES`.

asmecher: (I don't think there's any harm in just listing the affected tables, as you've done above for other cases; it makes things more greppable.)

classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
```diff
@@ -41,4 +43,114 @@ public function down() {
...
+	private function _settingsAsJSON() {
+
+		// Convert settings where type can be retrieved from schema.json
+		$schemaDAOs = ['SiteDAO', 'AnnouncementDAO', 'AuthorDAO', 'ArticleGalleyDAO', 'JournalDAO', 'EmailTemplateDAO', 'PublicationDAO', 'SubmissionDAO'];
+		$processedTables = [];
+		foreach ($schemaDAOs as $daoName) {
+			$dao = DAORegistry::getDAO($daoName);
+			$schemaService = Services::get('schema');
+			$schema = $schemaService->get($dao->schemaName);
+			$tableName = $dao->settingsTableName;
```
asmecher, line 60: There's no common base class or interface specifying that `settingsTableName` needs to exist (same with `schemaName`); I'd almost rather fully code the details out here rather than adding properties that are not fully used in other tables (e.g. `SiteDAO` has `settingsTableName` per the other PR, but it's never used anywhere but in this bit of ephemeral upgrade code).

asmecher: This code will come and go eventually, but there's a tendency to accumulate details in long-standing DAOs etc. that will eventually be misunderstood and cause trouble.

classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
Vitaliy-1: Indeed, doesn't support it.

classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
```diff
@@ -41,4 +43,121 @@ public function down() {
...
+	 * @param $searchBy array additional parameters to the where clause that should be combined with AND operator
+	 * @param $valueToConvert string column name for values to convert to JSON
+	 * @return void
+	 */
+	private function _toJSON($row, $tableName, $searchBy, $valueToConvert)
+	{
+		$oldValue = unserialize($row->{$valueToConvert});
+		// Reset arrays to avoid keys being mixed up
+		if (is_array($oldValue) && $this->_isNumerical($oldValue)) $oldValue = array_values($oldValue);
+		if (!$oldValue && !is_array($oldValue)) return; // don't continue if value cannot be unserialized
```
Vitaliy-1, line 127: Changed this line from `if (!$oldValue)` as empty arrays also can fall under this condition. I've noticed that in the test PostgreSQL dump in `publication_settings`, the value of `categoryIds` is empty serialized array. It will pass this check and `categoryIds` will be unset at the end.

Vitaliy-1: Hmm, interesting that the behavior with empty arrays is different in MySQL, at least on my test instance; probably related to Laravel's Query Builder.

asmecher: Hmm, yes, interesting -- maybe see if you can track this down as an expected/known behaviour? If not, and you can duplicate it with a small test script, maybe worth filing a new issue with Laravel. It would be interesting to see how responsive they are to cross-DB issues.

classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
```diff
@@ -41,4 +43,121 @@ public function down() {
...
+		// Remove empty filters
+		$searchBy = array_filter($searchBy, function ($item) use ($row) {
+			if (empty($row->{$item})) return false;
+			return true;
+		});
+
+		// Build raw where query
+		$whereRaw = '';
+		$lastKey = array_key_last($searchBy);
+		foreach ($searchBy as $key => $column) {
```
asmecher, line 141: I'd suggest using the `where` function to construct the conditions rather than `whereRaw`; it's much easier to read (and won't involve string concatenation!)

asmecher:
Looks good, @Vitaliy-1, I just made a couple of comments but when you've tinkered with those it'll be ready to go!

classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
Vitaliy-1: Strange, and now the behavior is the same. Something else interacted with those empty arrays, not the current update script. Maybe the script had differences when I tested it with MySQL.

PR #2874, opened 2020-09-24
Remove the version number from OJS

Mandroide:
Since currently the development is 3.x, I thought that this might help.


asmecher:
@Mandroide, better still, I'd suggest removing the version number from there entirely.

Mandroide:
> @Mandroide, better still, I'd suggest removing the version number from there entirely.

Ready

asmecher:
Merged, thanks!

PR #2886, opened 2020-10-18
doc: Replace OpenOffice with LibreOffice

Mandroide:
Since LibreOffice is currently more active than Apache OpenOffice, I see it as a good replacement.


asmecher:
Merged -- thanks, @Mandroide!

PR #2900, opened 2020-11-11
pkp/pkp-lib#6330 add lens galley hook name to download finished hooks

jnugent:
one of two pull requests for #6330


asmecher:
Merged and back-ported to `stable-3_2_1`.

PR #2953, opened 2020-12-15
Update README.md

inittab:
Directories under `plugins`should be writeable by the web server in order to allow plugin installation.


asmecher:
Noted, thanks! I'll adapt this to the other application repos.

PR #2954, opened 2020-12-16
Add coverImage to publication schema to avoid unserialize() error during v3 migration (setFileName / DAO.inc.php)

davidjb:
When upgrading from OJS v2.4.8-5 to OJS v3.2.1 and when running `php tools/upgrade.php upgrade`, I hit an issue during the `Installer::setFileName` process which looked like so:

```
[...]
[code: Installer Installer::setFileName]
PHP Notice:  unserialize(): Error at offset 0 of 28 bytes in /usr/share/nginx/html/ojs/lib/pkp/classes/db/DAO.inc.php on line 357
PHP Notice:  unserialize(): Error at offset 0 of 28 bytes in /usr/share/nginx/html/ojs/lib/pkp/classes/db/DAO.inc.php on line 357
PHP Notice:  unserialize(): Error at offset 0 of 28 bytes in /usr/share/nginx/html/ojs/lib/pkp/classes/db/DAO.inc.php on line 357
PHP Notice:  unserialize(): Error at offset 0 of 28 bytes in /usr/share/nginx/html/ojs/lib/pkp/classes/db/DAO.inc.php on line 357
[...]
```

Delving into why, the underlying error stems from the `coverImage` setting field on a publication being treated as an `object` when it's actually a string - dumping the value passed to unserialize at that point in the code shows it's like so:  `string(28) "cover_article_1234_en_US.png"`. 

Access happens in `lib/pkp/classes/db/DAO.inc.php:357`, which gets called by `lib/pkp/classes/db/SchemaDAO.inc.php:250`, which loads the conversion type from the schema.  In this case, because of the lack of a `coverImage` field, hence why the `unserialize()` call is erroring. 

This PR add this field into `schemas/publication.json` which allows the migration from v2 to complete to v3 without error.  This mightn't be _the_ overall solution as I'm not familiar with the internals of OJS so changes/improvements to this PR are very welcome.

### DIFF
--- a/schemas/publication.json
+++ b/schemas/publication.json
@@ -33,6 +33,12 @@
 				"nullable"
 			]
 		},
+		"coverImage": {
+			"type": "string",
+			"validation": [
+				"nullable"
+			]
+		},
 		"pages": {
 			"type": "string",
 			"description": "The pages of the issue in which this article appears if it is published in a physical format.",


davidjb:
Actually, whilst this PR avoids the database error during the migration, it isn't sufficient to correctly migrate data because it (or some other part of the migration) appears to have left the original database rows in place (e.g. where `setting_value` is the string of `cover_article_1234_en_US.jpg`):

```sql
+----------------+--------+--------------+--------------------------------------------------------+
| publication_id | locale | setting_name | setting_value                                          |
+----------------+--------+--------------+--------------------------------------------------------+
|           1234 |        | coverImage   | c:1:{s:5:"en_US";s:28:"cover_article_2345_en_US.jpg";} |
|           1234 | en_US  | coverImage   | cover_article_2345_en_US.jpg                           |
```

In my OJS install's case, there were only a handful of publications with coverImages applied so I've gone through each and fixed them to prevent conflicts between a string being stored and the saving of the new object format, presumably since there are two rows for `coverImage` for each publication.

PR #2955, opened 2020-12-16
[OJS] #6490 Native Import/Export Plugin Improvements

defstat:


classes/issue/Issue.inc.php
```diff
@@ -559,6 +559,10 @@ function hasDescription() {
 	function getDAO() {
 		return DAORegistry::getDAO('IssueDAO');
 	}
+
+	function getUIDisplayString() {
```
asmecher, line 563: Add code self-documentation.

plugins/importexport/native/templates/resultsExport.tpl
```diff
@@ -0,0 +1,76 @@
...
+							<li>{$relatedItemMessage|escape}</li>
+						{/foreach}
+					{/foreach}
+				</ul>
+			{/if}
+		{/foreach}
+	{/foreach}
+{/if}
+
+{if array_key_exists('errors', $errorsAndWarnings) && $errorsAndWarnings.errors|@count > 0}
```
asmecher, line 51: (as elsewhere)

plugins/importexport/native/templates/resultsImport.tpl
```diff
@@ -0,0 +1,71 @@
...
+					<li>
+						{$contentItem->getUIDisplayString()}
+					</li>
+				{/foreach}
+			{/foreach}
+		{/foreach}
+	</ul>
+{/if}
+
+{if array_key_exists('warnings', $errorsAndWarnings) && $errorsAndWarnings.warnings|@count > 0}
```
asmecher, line 28: (as elsewhere)

asmecher (changes requested):
Good to see so much removed code! Several small-ish comments, thanks.

classes/issue/Issue.inc.php
defstat: Done.

plugins/importexport/native/templates/resultsExport.tpl
defstat: Do you mean a comment/documentation block here?

defstat: In fact those files are the same as in lib-pkp. I had trouble getting them from lib-pkp though. I think the best way is to find a way to get the templates from pkp-lib if possible

asmecher: Yes, I agree -- put those in a template of their own and call them wherever they're needed. I think there were 6 copies in the PRs, each with two almost-duplicate blocks.

plugins/importexport/native/templates/resultsImport.tpl
defstat: Moved everything to pkp-lib and added a common `innerResults.tpl` template for the specific block

plugins/importexport/native/templates/resultsExport.tpl
defstat: Moved everything to pkp-lib and added a common `innerResults.tpl` template for the specific block

classes/issue/Issue.inc.php
```diff
@@ -559,6 +559,14 @@ function hasDescription() {
 	function getDAO() {
 		return DAORegistry::getDAO('IssueDAO');
 	}
+
+	/**
+	 * Display the object in Import/Export results
+	 * @return string A string that Identifies the object
+	 */
+	function getUIDisplayString() {
+		return $this->getId() . ' - ' . $this->getIssueIdentification();
```
asmecher, line 568: (As in the pkp-lib PR: this should probably be a locale key.)

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
```diff
@@ -214,15 +210,13 @@ function addArticles($doc, $issueNode, $issue) {
...
-		$exportFilter = array_shift($nativeExportFilters);
-		$exportFilter->setDeployment($this->getDeployment());
+		$currentFilter = PKPImportExportFilter::getFilter('issuegalley=>native-xml', $this->getDeployment());
 
 		$issueGalleyDao = DAORegistry::getDAO('IssueGalleyDAO'); /* @var $issueGalleyDao IssueGalleyDAO */
 		$issue = $issueGalleyDao->getByIssueId($issue->getId());
-		$issueGalleysDoc = $exportFilter->execute($issue);
+		$issueGalleysDoc = $currentFilter->execute($issue);
+
+		// throw new Exception('Ass Issue Galleys Exception');
```
asmecher, line 219: Dead code

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
```diff
@@ -214,15 +210,13 @@ function addArticles($doc, $issueNode, $issue) {
...
 		if ($issueGalleysDoc->documentElement instanceof DOMElement) {
```
asmecher, line 220: What condition would cause this not to be true? Off the top of my head it would be an `issue_galley` node that didn't conform to the schema. In that case, rather than continuing without including it, it would be better for the export to fail.

asmecher (changes requested):

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
defstat: Removed

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
defstat: Because the initial code "https://github.com/pkp/pkp-lib/blob/master/classes/filter/Filter.inc.php#L442" returned the `$this->_output`, that check there indented not to try to add the `$issueGalleysDoc`. But you have a point there - Perhaps any such failure should result in a total process failure. 

If so, I will have to check in other parts too for such a behavior (I found it in 4-5 more places)

classes/issue/Issue.inc.php
asmecher: This one still needs attention.

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
asmecher: Is this resolved?

asmecher:
Thanks, @defstat, just a couple of outstanding issues! It looks like this needs a rebase -- I think you had changes to some code that has since been removed.

classes/issue/Issue.inc.php
defstat: Fixed

plugins/importexport/native/filter/IssueNativeXmlFilter.inc.php
defstat: Resolved

PR #2957, opened 2020-12-19
set urlPath when importing galleys via NativeXML

eocarragain:
This pull request aims to fix bug reported https://github.com/pkp/pkp-lib/issues/6501 ("Galley url_path not imported in Native XML Import/Export")

I have tested it on OJS 3.2.1-1 and it produces the desired result, i.e. imported galleys now have the url_path specified in the article_galley element in the xml.

Feedback very welcome. Thanks

### DIFF
--- a/plugins/importexport/native/filter/NativeXmlArticleGalleyFilter.inc.php
+++ b/plugins/importexport/native/filter/NativeXmlArticleGalleyFilter.inc.php
@@ -74,6 +74,7 @@ class NativeXmlArticleGalleyFilter extends NativeXmlRepresentationFilter {
 			if ($newSubmissionFileId) $addSubmissionFile = true;
 		}
 		$representation = parent::handleElement($node);
+		$representation->setData('urlPath', $node->getAttribute('url_path'));
 
 		for ($n = $node->firstChild; $n !== null; $n=$n->nextSibling) if (is_a($n, 'DOMElement')) switch($n->tagName) {
 			case 'name':


eocarragain:
1 of the 5 CI jobs failed but, from the log, it seems to have been an unrelated issue with starting Chrome?

plugins/importexport/native/filter/NativeXmlArticleGalleyFilter.inc.php
```diff
@@ -74,6 +74,7 @@ function handleElement($node) {
 			if ($newSubmissionFileId) $addSubmissionFile = true;
 		}
 		$representation = parent::handleElement($node);
+		$representation->setData('urlPath', $node->getAttribute('url_path'));
```
defstat, line 77: Try adding this line at `NativeXmlRepresentationFilter`, which is the parent class. That way your change will go into the PKP-LIB project, effecting all applications (OJS, OMP, OPS). Moreover, the `url_path` attribute is declared here (https://github.com/pkp/pkp-lib/blob/master/plugins/importexport/native/pkp-native.xsd#L224) which resides at PKP-LIB project.

defstat (changes requested):
Along with the review change, please check whether the exported XML file you are using, actually has the `url_path` attribute in its `article_galley` node. If it does not, we should consider adding the export line, like `$representationNode->setAttribute('url_path', $representation->getData('urlPath'));` in the `createRepresentationNode` function of the `RepresentationNativeXmlFilter` export filter.

eocarragain:
thanks @defstat 

That all makes sense. See new pull request against pkp-lib here: https://github.com/pkp/pkp-lib/pull/6521

Closing this pull request in favour of https://github.com/pkp/pkp-lib/pull/6521

PR #2958, opened 2020-12-19
pkp/pkp-lib#6411 add datacite test account settings and consider datacite test api 

bozana:
s. https://github.com/pkp/pkp-lib/issues/6411

there is `catch (GuzzleHttp\Exception\RequestException $e)`. should there be `catch (Exception $e)` too?


asmecher:
Merged, thanks! (I'm not sure other kinds of exceptions will happen from Guzzle -- if they do, they'll propagate up to the application and turn into a fatal error -- I'm OK with this for now, though at some point would like to add a better application-wide exception handler.)

PR #2959, opened 2020-12-21
Use two queries in _getRecordsRecordSet

ajnyga:


classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -211,8 +211,19 @@ function setOAIData($record, $row, $isRecord = true) {
 	function _getRecordsRecordSet($setIds, $from, $until, $set, $submissionId = null, $orderBy = 'journal_id, submission_id') {
 		$journalId = array_shift($setIds);
 		$sectionId = array_shift($setIds);
+		
+		$result = $this->retrieve('
+			SELECT journal_id FROM journal_settings WHERE setting_name = ? AND setting_value = ?',
```
asmecher, line 216: I'd suggest doing a LEFT JOIN from journals onto journal_settings, so as to include journals that don't have `enableOai` settings. Otherwise these will be assumed enabled by the code you've proposed.

ajnyga: the query will get all journals that have oai specifically disabled. I was thinking along the logic that if the setting is completely missing, the OAI should be on. Note that the journa_id values that we get are used in a NOT IN statement.

asmecher: Whether to treat ambiguous installations as enabled or disabled in this case should be consistent with the same issue handled elsewhere. See e.g. https://github.com/pkp/ojs/blob/master/classes/oai/ojs/OAIDAO.inc.php#L238; here we require a specific `enableOai` setting to be set to `1`.

ajnyga: Ok, I will change this to the pr on master. The 3.1.2 stable pr is not needed imo.

PR #2961, opened 2020-12-22
pkp/ojs#6320

majaroenne:
Change cast to char to cast to char(20) for postgres compatibility

According the postgres problem described in issue #6320 I found one place in IssueDAO there the sql cast was not changed
Otherwise the cast change i issue #6320 seems to work perfect with  postgres on a big dataset


asmecher:
Merged and forward-ported -- thanks!

PR #2964, opened 2020-12-22
pkp/pkp-lib#6503 divide _getRecordsRecordSet to two queries to make OAI faster

ajnyga:


ajnyga:
ping @asmecher also not this pr for current master using querybuilder.

ajnyga:
ping @asmecher 
Here is the pr for master branch, the other one is not needed I think.

I was not sure how to use left join in this case, so I used a NOT IN instead. Basically all journals mentioned in the journals table are excluded from OAI expect for the ones that have OAI specifically turned on in journal_settings.

I do think that we should not allow a situation where the OAI setting is neither on or off.

asmecher:
Looks good -- thanks, @ajnyga!

Can you describe your thinking about the OAI setting being neither off nor on a little further?

PR #2597, opened 2020-01-29
Stable 3 1 2: Enhancement in mEDRA plugin

njikonga-med:
Modification of plugin mEDRA to allow user to deposit metadata not only
on mEDRA, but also in Crossref. implementation of logic that allows
mEDRA plugin user to choose either to deposit only on mEDRA, or to
deposit also in Crossref. This new implementation allows also the
display of some validations before the deposit in Crossref in case of
errors. The errors are displayed in a table. Some other enhancements:
insert as contributor the keyNames. Some bugfixes: remotion of the
control that remove on issn the value that is NOT in 0-9.

PR #2965, opened 2020-12-28
Medra test

bozana:
Here are the changes of the medra plugin to support another mEDRA service, the DOI Crossref deposit.
The PR contains the contribution of @njikonga-med provided in the PR https://github.com/pkp/ojs/pull/2597, as well as some further changes and fixes.

PR #2966, opened 2020-12-28
mEDRA plugin modification to enable also the Crossref deposit

bozana:
Here are the changes of the medra plugin to support another mEDRA service, the DOI Crossref deposit.
The PR contains the contribution of @njikonga-med provided in the PR #2597, as well as some further changes and fixes.

### DIFF
--- a/plugins/importexport/medra/MedraExportPlugin.inc.php
+++ b/plugins/importexport/medra/MedraExportPlugin.inc.php
@@ -90,10 +90,13 @@ class MedraExportPlugin extends DOIPubIdExportPlugin {
 	 * @copydoc PubObjectsExportPlugin::depositXML()
 	 */
 	function depositXML($objects, $context, $filename) {
-		// Use a different endpoint for testing and
-		// production.
 		$this->import('classes.MedraWebservice');
-		$endpoint = ($this->isTestMode($context) ? MEDRA_WS_ENDPOINT_DEV : MEDRA_WS_ENDPOINT);
+		// Use a different endpoint for testing and production.
+		// New endpoint: use a different endpoint if the user selects the checkbox to deposit also in Crossref.
+		$crEnabled = false;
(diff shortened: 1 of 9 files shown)


bozana:
@asmecher, these are the changes for stable-3_2_1. Could you take a look?

asmecher:
Is this change not required for the 3.2.1 port? https://github.com/pkp/ojs/pull/2965/files#diff-d4826263ff6f79bd01bc8ff7a50fe9ea6a76f6f77e19500597dacb9bfbcd3b83R103

bozana:
No, it seems somebody fixed it, s. https://github.com/pkp/ojs/blob/stable-3_2_1/plugins/importexport/medra/filter/IssueMedraXmlFilter.inc.php#L103.

njikonga-med:
Hi @bozana and @asmecher , I would like to thank so much for your support in this enhancement introduced on mEDRA plugin. Now that the merged has been performed, how then, end user can upgrade only the mEDRA plugin without the CMS? Is there any procedure that explain how user can do it for their own? I saw that in the Web site settings of an OJS CMS installation, on the tab installed plugin, it's possible to upgrade the mEDRA plugin, but the user have to first compress the directory in format medra.tar.gz? It's right? Could you please let me have more information about it?

Regards,
Jehu Njikonga

bozana:
Hi @njikonga-med, maybe the easiest way would be for the web admin to get the changes as `diff` and apply them as a `patch`.
One can get the `diff` by adding `.diff` to a github (merge) commit, e.g. 
for the changes in stable_3_1_2: https://github.com/pkp/ojs/commit/e9b7f8b254a3637f4439170a04f6de67754bba7d.diff
or for the changes here, in stable_3_2_1: https://github.com/pkp/ojs/commit/5ddf905b8faec100832daddb5f438d9179eef026.diff

PR #3026, opened 2021-01-29
pkp/pkp-lib#5948 define the current issue back again at the test end

bozana:
s. https://github.com/pkp/pkp-lib/issues/5948
forgotten to define the current issue back again at the end of the test


bozana:
@NateWr, would you like to take a look?
I would then merge when the tests pass...

PR #3046, opened 2021-02-19
[OJS] #5048 Move notification type settings in a more core place

defstat:


classes/notification/NotificationManager.inc.php
```diff
@@ -158,6 +158,19 @@ protected function getMgrDelegate($notificationType, $assocType, $assocId) {
 		return parent::getMgrDelegate($notificationType, $assocType, $assocId);
 	}
 
+	/**
+	 * @copydoc PKPNotificationManager::getNotificationSettingsMap()
+	 */
+	public function getNotificationSettingsMap() {
```
asmecher, line 164: This overrides a protected function, which will cause a warning/error, I think.

dbscripts/xml/upgrade/3.2.0_preupdate_email_templates.xml
```diff
@@ -41,6 +41,13 @@
 		<!-- 4097 = ROLE_ID_ASSISTANT, 768 and 8192 = unused role ids -->
 		<query>UPDATE email_templates_default SET from_role_id=4097 WHERE from_role_id=768 OR from_role_id=8192</query>
 		<!-- 4097 = ROLE_ID_ASSISTANT, 768 and 8192 = unused role ids -->
-		<query>UPDATE email_templates_default SET to_role_id=4097 WHERE to_role_id=768 OR to_role_id=8192</query>
+		<query>
+			UPDATE email_templates_default SET to_role_id=4097 WHERE to_role_id=768 OR to_role_id=8192
```
asmecher, line 45: Does this need to move into a newer upgrade?

locale/en_US/emails.po
```diff
@@ -25,7 +25,7 @@ msgstr ""
 "<br />\n"
 "Link: {$url}<br />\n"
 "<br />\n"
-"{$principalContactSignature}"
+"{$principalContactSignature} {$unsubscribeLink}"
```
asmecher, line 28: Should the new link be placed on a new line?

asmecher (changes requested):

classes/notification/NotificationManager.inc.php
defstat: Although it does not produce a warning it definitely needs addressing. (I think the reason it does not complain is that it thinks I overload the function rather than overriding it, and then, inside, I am using the parent function (which is OK). 

The exposed function is, in fact the child's `getNotificationSettingsMap` function. So here we have 2 solutions I think (except the one already implemented).
1. Change the visibility of parent's `getNotificationSettingsMap` to public.
2. Implementing a "getter" for the SettingsMap at the child class level which will expose the array, calling only the `getNotificationSettingsMap` inside, and changing the `getNotificationSettingsMap` to `protected`. I will go for this one.

locale/en_US/emails.po
defstat: I have put an `<hr />` in between.

dbscripts/xml/upgrade/3.2.0_preupdate_email_templates.xml
defstat: Yes. I changed it to 3.4

classes/notification/NotificationManager.inc.php
asmecher: Agreed, I think making the parent class implementation `public` is the best option!

PR #3062, opened 2021-03-03
[TESTS ONLY] pkp/pkp-lib#6801 Replace name with original_file_name

Vitaliy-1:


Vitaliy-1:
Looks like tests are failing because of commits made in a shared library earlier

PR #3086, opened 2021-03-28
pkp/pkp-lib#6887 fix crossref errors handling

bozana:
s. https://github.com/pkp/pkp-lib/issues/6887

PR #3087, opened 2021-03-28
Fix querying Crossfef API by updating status URL to doi.crossref.org

davidjb:
Up until a few weeks ago, querying of statuses of deposits made to Crossref were working - being made to the URL of https://api.crossref.org/servlet/submissionDownload (as far as I saw, this URL has been in OJS for almost 5 years).  However - and I'm still discussing this with CrossRef support as to what changed recently - the above URL is now returning a 404 (`{"status":"error","message-type":"route-not-found","message-version":"1.0.0","message":"Route not found"}`) and no longer functioning.

This PR updates the status URL for Crossref to point at doi.crossref.org rather than api.crossref.org, as per the official docs at https://www.crossref.org/education/content-registration/verify-your-registration/submission-queue-and-log/#00145; making this switch allows the plugin to continue working as expected.

Quoting from the support replies I've had from Crossref:

>  api.crossref.org/servlet/submissionDownload isn't a deposit endpoint.  It's not used for anything, actually. 
>
> doi.crossref.org/servlet/submissionDownload is a query endpoint that lets you poll for submission logs that confirm the success or failure of your deposits, given a username, password, and file name, as shown here. 
>
> **There is no "api.crossref.org" equivalent.**
>
> [...]
>
> https://api.crossref.org/v2/deposits is a special route that was created just for the OJS plugin, and specifically just for the automated deposits made via the OJS plugin.  It shouldn't be used in any other situation. 

I'll comment back with details as to what caused this change once I hear back from Crossref.  I suspect that the api.crossref.org endpoint was either working unintentionally on both URLs, or perhaps Crossref changed their API and documentation without notice at some point.  That said, the [docs](https://www.crossref.org/education/content-registration/verify-your-registration/submission-queue-and-log/#00145) were last updated on 2020-April-08 - assuming that's true, that points to no official recent changes.

### DIFF
--- a/plugins/importexport/crossref/CrossRefExportPlugin.inc.php
+++ b/plugins/importexport/crossref/CrossRefExportPlugin.inc.php
@@ -26,7 +26,7 @@ define('CROSSREF_API_URL', 'https://api.crossref.org/v2/deposits');
 //TESTING
 define('CROSSREF_API_URL_DEV', 'https://test.crossref.org/v2/deposits');
 
-define('CROSSREF_API_STATUS_URL', 'https://api.crossref.org/servlet/submissionDownload');
+define('CROSSREF_API_STATUS_URL', 'https://doi.crossref.org/servlet/submissionDownload');
 //TESTING
 define('CROSSREF_API_STATUS_URL_DEV', 'https://test.crossref.org/servlet/submissionDownload');
 


davidjb:
Closing in lieu of #3086 which features the same fix

PR #3103, opened 2021-04-12
pkp/pkp-lib#6915 Laravel Facades instead of Capsule Manager calls

Vitaliy-1:


Vitaliy-1:
Missed the part of initializing Facades during the installation

PR #3126, opened 2021-05-09
XML locale order

ruienf:
Hi!

I'm not sure if a pull request is the best option for this...

I'm using OjS 3.3.0.4 and I'm using DOAJ I/E plugin.

I've exported all of our articles in .xml and upload it on DOAJ.

However, our journal primary language is pt_PT, making "por" language apperar first in the .xml file (our other languages are en_US and es_ES, and we have the title, abstract and keywords in the three languages).

DOAJ only shows one language after upload, and it considers the first language in the .xml file order.

Although our primary language is pt_PT, we'd like to have the english version of our titles, abstracts and keywords in DOAJ (for better dissemination of our articles). 

We've done that by manually editing the .xml file and removing the "por" entries. Then, we've dig a little bit in the plugin code and we're able to get a proper .xml file order by adding ksort ($articleTitles), ksort ($articleAbstracts) and ksort ($articleKeywords) in  filter/DOAJXmlFilter.inc.php. 
[DOAJXmlFilter.inc (sort).zip](https://github.com/pkp/ojs/files/6448074/DOAJXmlFilter.inc.sort.zip)

But this only worked because en_US is the first key in an alphabetical order...

It would be great to have an option to define the locale/language order in the settings page of the plugin, so that the user can decide which language will be considered in the DOAJ upload.

Can you help me getting that done?

Thank you!

Kind regards,

Rui Pedro Silva


asmecher:
@ruienf, I think this is the same as https://github.com/pkp/pkp-lib/issues/3119 -- could you add your feedback there? Thanks!

ruienf:
Yes @asmecher , it's the same issue, sorry I miss it!...

I'll do as you sugest.

Thank you!!!

PR #3134, opened 2021-05-21
pkp/pkp-lib#6963 Improve OAI performance

asmecher:


classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -252,72 +242,64 @@ public function _getRecordsRecordSet($setIds, $from, $until, $set, $submissionId
...
+            ->where('i.published', '=', 1)
+            ->where('j.enabled', '=', 1)
+            ->where('a.status', '=', PKPSubmission::STATUS_PUBLISHED);
+        if ($excludeJournals) $submissionsQuery->whereNotIn('j.journal_id', $excludeJournals);
+        if (isset($journalId)) $submissionsQuery->where('j.journal_id', '=', (int) $journalId);
+        if (isset($sectionId)) $submissionsQuery->where('p.section_id', '=', (int) $sectionId);
+        if ($from) $submissionsQuery->where('GREATEST(a.last_modified, i.last_modified)', '>=', $from);
+        if ($until) $submissionsQuery->where('GREATEST(a.last_modified, i.last_modified)', '<=', $until);
+        if ($submissionId) $submissionsQuery->where('a.submission_id', '=', (int) $submissionId);
+
```
bozana, line 274: According to the documentation we could maybe use the conditional clauses (s. https://laravel.com/docs/7.x/queries#conditional-clauses) instead of `if ...`, or?
If so, the result is surely the same, so I will let you decide what to use :-)

classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -252,72 +242,64 @@ public function _getRecordsRecordSet($setIds, $from, $until, $set, $submissionId
...
+                                NULL AS oai_identifier'))
+            ->join('publications AS p', 'a.current_publication_id', '=', 'p.publication_id')
+            ->join('publication_settings AS psissue', function($join) {
+                $join->on('psissue.publication_id', '=', 'p.publication_id');
+                $join->where('psissue.setting_name', '=', DB::raw('\'issueId\''));
+                $join->where('psissue.locale', '=', DB::raw('\'\''));
+            })
+            ->join('issues AS i', function($join) {
+                $join->on(DB::raw('CAST(i.issue_id AS CHAR(20))'), '=', 'psissue.setting_value');
+            })
```
bozana, line 262: Could this be used without `function($join)...`?
If so the result is the same, so you can decide and leave it as it is... :-)

classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -223,19 +223,9 @@ public function setOAIData($record, $row, $isRecord = true)
...
-     * @param $until int/string *nix timestamp or ISO datetime string
-     * @param $set string
-     * @param $submissionId int optional
-     * @param $orderBy string UNFILTERED
-     *
-     * @return Iterable
+     * @copydoc PKPOAIDAO::_getRecordsRecordSet
      */
-    public function _getRecordsRecordSet($setIds, $from, $until, $set, $submissionId = null, $orderBy = 'journal_id, submission_id')
+    public function _getRecordsRecordSetQuery($setIds, $from, $until, $set, $submissionId = null, $orderBy = 'journal_id, submission_id', $offset = null, $limit = null)
```
bozana, line 228: As commented in the pkp-lib, do we need the `$offset` and `$limit` parameter?

bozana (commented):
@asmecher, just a few comments, else all fine...

classes/oai/ojs/OAIDAO.inc.php
asmecher: The conditional clauses are a little more verbose, but using them permits me to build the query in one (long) statement, and I wonder if eventually we'll be able to benefit from this in other ways. In any case, I made the change.

classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -252,72 +242,90 @@ public function _getRecordsRecordSet($setIds, $from, $until, $set, $submissionId
...
+                ])
+                ->when(isset($journalId), function($query, $journalId) {
+                    return $query->join('data_object_tombstone_oai_set_objects AS tsoj', function($join) use ($journalId) {
+                          $join->on('tsoj.tombstone_id', '=', 'dot.tombstone_id');
+                          $join->where('tsoj.assoc_type', '=', ASSOC_TYPE_JOURNAL);
+                          $join->where('tsoj.assoc_id', '=', (int) $journalId);
+                        })->addSelect(['tsoj.assoc_id']);
+                    }, function($query) {
+                        return $query->addSelect([DB::raw('NULL AS assoc_id')]);
+                })
```
bozana, line 304: Why are the 4 last lines (addSelects) needed?

classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -252,72 +242,90 @@ public function _getRecordsRecordSet($setIds, $from, $until, $set, $submissionId
...
+            ->join('issues AS i', DB::raw('CAST(i.issue_id AS CHAR(20))'), '=', 'psissue.setting_value')
+            ->join('sections AS s', 's.section_id', '=', 'p.section_id')
+            ->join('journals AS j', 'j.journal_id', '=', 'a.context_id')
+            ->where('i.published', '=', 1)
+            ->where('j.enabled', '=', 1)
+            ->where('a.status', '=', PKPSubmission::STATUS_PUBLISHED)
+            ->when($excludeJournals, function($query, $excludeJournals) {
+                return $query->whereNotIn('j.journal_id', $excludeJournals);
+            })
+            ->when(isset($journalId), function($query) use ($journalId) {
```
bozana, line 271: Why sometimes with and sometimes without 'use'?

bozana (commented):
@asmecher, I have only a few questions, I think rather because I do not understand them 100%

classes/oai/ojs/OAIDAO.inc.php
asmecher: This keeps the new code consistent with the prior tests for `if ($myVariable) ...` vs. `if (isset($myVariable)) ...`
I'm not sure if the difference was originally there for a reason and didn't want to introduce regressions.

classes/oai/ojs/OAIDAO.inc.php
asmecher: The column sets in the two halves of the UNION statement need to be consistent and ordered the same, i.e. the `a, b, c, d` and `c, d, e, f` in
```
SELECT
  a, b, c, d
  FROM ...
UNION SELECT
  c, d, e, f
  FROM ...
```
This keeps the column set consistent with the other half of the `UNION`.

PR #3143, opened 2021-05-31
pkp/pkp-lib#7096 OJS 2 Supp Files Missing in OJS 3.2

ewhanson:


dbscripts/xml/upgrade/3.2.0_versioning.xml
```diff
@@ -143,6 +143,19 @@
...
 		</query>
 
+		<!--
+			Copy over updated submission_galleys table with supplementary files as galleys.
+			These are otherwise lost as these changes are added to submission_galleys but never
+			copied over to temp_submission_galleys.
+		-->
+
+		<!-- First remove temp table created in 3.2.0_preupdate_versioning -->
+		<drop table="temp_submission_galleys" />
```
asmecher, line 153: Are you sure this table won't be necessary for the 3.1.x to 3.2.x upgrade path? Temporary tables are generally created no earlier than necessary and removed no later.

ewhanson: I had wondered about that as well, but I didn't see it being used anywhere else. It was created in `3.2.0_preupdate_versioning.xml`, moved from the temp table to the publication_galleys table and subsequently dropped in `3.2.0_versioning.xml`. Are there any other ways it might affect the 3.1.x to 3.2.x upgrade process other than writing to `temp_submission_galleys`?

asmecher: Sorry for the delay in responding --

In summary, the `temp_submission_galleys` table is used as follows as part of the 3.1.x to 3.2.x introduction of metadata versioning:

1. Backup the `submission_galleys` table contents in `temp_submission_galleys`.
2. (do other 2.x upgrade things)
3. (do other 3.x upgrade things)
4. Adapt the content in `temp_submission_galleys` into `publication_galleys`.

The problem you're encountering with the 2.x to 3.x upgrade, I think, is that step 2 includes population of the `submission_galleys` rather than `temp_submission_galleys`, and that new content isn't handled in step 4.

I think what your PR is doing changes it to...

1. Backup the `submission_galleys` table contents in `temp_submission_galleys`.
2. (do other 2.x upgrade things affecting `submission_galleys`)
3. (do other 3.x upgrade things affecting `temp_submission_galleys`)
4. Throw away and re-create `temp_submission_galleys`
5. Adapt the content in `temp_submission_galleys` into `publication_galleys`.

This will suffer the reverse problem -- the 2.x upgrade might be improved but the 3.1.x to 3.2.x upgrade will suffer.

I think the solution you're looking for might be to find cases where the 2.x to 3.x upgrade process refers to `submission_galleys` after `temp_submission_galleys` has been created, and change them to use `temp_submission_galleys` instead.

ewhanson: Thanks! Yes, that's exactly right. I'll take a look at re-addressing the 2.x to 3.x upgrade issue in the way you mentioned.

ewhanson:
Hi @asmecher, I've adapted reference `temp_submission_galleys` as part of the 2.x to 3.2 upgrade process and removed the dropping and recreating of the table. I ran a test upgrade from 2.4 to 3.2 and it successfully brought the support files along with it. Could you have a look and let me know what you think? Thanks!

dbscripts/xml/upgrade/3.2.0_versioning.xml
suprianto12: ye

dbscripts/xml/upgrade/3.2.0_preupdate_versioning.xml
```diff
@@ -47,4 +47,13 @@
...
 		<rename table="citations" column="submission_id" to="publication_id" />
 	</sql>
+
+	<!--
+	Necessary to allow longer supplementary files titles to be added as a label when
+	they are converted to public galleys. This happens with previous upgrades from 2.4.x
+	to other versions of OJS 3, but does not otherwise happen with the 2.4.x to 3.2.x upgrade.
+ -->
+	<sql>
+		<query>ALTER TABLE temp_submission_galleys MODIFY COLUMN label VARCHAR(255);</query>
```
asmecher, line 57: The `label` column will already be varchar(255) for OJS 3.x installations, but it'll be varchar(32) for OJS 2.x installations.

I think this needs two changes:
1. Move it to `dbscripts/xml/upgrade/3.0.0_preupdate.xml`, just after `article_galleys` is renamed to `submission_galleys`, so it's only applied when needed
2. I don't think this SQL will work for PostgreSQL, so you'd need to add a clause for each DBMS and adapt.

ewhanson:
Hi @asmecher, I've gotten this updated with the additions you suggested as well as addressed a few other things that came up in testing. @jnugent and I have run this through a few rounds of testing with journals on OJS 2.4, and everything is working as expected. Could you have another look? Thanks!

asmecher:
That looks good from a 2.x to 3.x perspective, thanks! But could you run one last smoke test of a 3.1.x to 3.2.x upgrade? This would help confirm that there aren't any unintended regressions. (Did you run tests with a PostgreSQL database?)

ewhanson:
Thanks. No I hadn't yet, a PostgreSQL test was going to be my next step. I'll run it through that and a 3.1.x to 3.2.x upgrade to make sure everything's working correctly.

ewhanson:
Hi @asmecher, could you have a final look at this? It should be good to go now. I've run tests with both mysql and Postgres and upgrades completed successfully. Thanks!

dbscripts/xml/upgrade/3.0.0_preupdate.xml
```diff
@@ -58,6 +58,15 @@
 		<!-- article_galleys to submission_galleys -->
 		<rename table="article_galleys" column="article_id" to="submission_id" />
 		<rename table="article_galleys" to="submission_galleys" />
+
+		<!--
+			Necessary to allow longer supplementary files titles to be added as a label when
+			they are converted to public galleys. This happens with previous upgrades from 2.4.x
+			to other versions of OJS 3, but does not otherwise happen with the 2.4.x to 3.2.x upgrade.
+		-->
+		<query driver="mysql,mysqli">ALTER TABLE submission_galleys MODIFY COLUMN label VARCHAR(255);</query>
```
asmecher, line 67: (Not a show-stopper, but the `;` at the end of each statement isn't needed.)

asmecher:
Looks good, @ewhanson. I presume you've checked the schema and constraints on the table after upgrading with a freshly-generated OJS 3.2.1 install? From the looks of it (particularly PostgreSQL) you've done that but just to confirm.

dbscripts/xml/upgrade/3.0.0_preupdate.xml
ewhanson: Thanks, missed these ones. `;`s removed.

ewhanson:
> Looks good, @ewhanson. I presume you've checked the schema and constraints on the table after upgrading with a freshly-generated OJS 3.2.1 install? From the looks of it (particularly PostgreSQL) you've done that but just to confirm.

Yes, they match with both PostgreSQL and MySQL.

asmecher:
OK, thanks -- looks good. Please go ahead and merge! I don't know if we'll release any more 3.2.1-x packages but it'll be helpful to have at least on the stable branch.

ewhanson:
Thanks, @asmecher! I don't think I can merge it, though, as I don't have write access to the repo.

asmecher:
Done!

jnugent:
Hi everyone, unfortunately there is still a problem with the way this works.  I'm upgrading a client from 2.4.8 to stable-3_2_1 in preparation to upgrade to 3.3 and the Installer::convertSupplementaryFiles method is failing with:

`Uncaught Exception: DB Error: Field 'file_id' doesn't have a default value Query: INSERT INTO temp_submission_galleys (locale, submission_id, remote_url, label) VALUES (?, ?, ?, ?)`

I've worked through this.  When you run the upgrade, the first step that runs is:

`[data: dbscripts/xml/upgrade/3.0.0_preupdate.xml]`

which renames `article_galleys` to `submission_galleys`.  This table has a `file_id` column, as defined in the 2.4.8 schema here:

https://github.com/pkp/ojs/blob/ojs-stable-2_4_8/dbscripts/xml/ojs_schema.xml#L848

The `submission_galleys` table is used as a blueprint for the `temp_submission_galleys` table in the `[data: dbscripts/xml/upgrade/3.2.0_preupdate_versioning.xml]` step, so the `temp_submission_galleys` ends up with a `file_id` column that cannot be null.  

Unfortunately, the Installer::convertSupplementaryFiles method does an explicit insert into this table and does not provide a file_id, so the error occurs right away.

PR #3132, opened 2021-05-17
pkp/pkp-lib#4622 Enabling Laravel's queue worker at register_shutdown_function

henriqueramos:

PR #3144, opened 2021-06-07
pkp/pkp-lib#7105 Allow queued jobs to be processed by a cron job

henriqueramos:
This is a `sub pull request` for the pkp/pkp-lib#4622 [PR 3132](https://github.com/pkp/ojs/pull/3132).


henriqueramos:
Keeping in mind this should only be merged after [PR 3132](https://github.com/pkp/ojs/pull/3132)

asmecher:
I'd suggest interactive-rebasing your branch so that there is only one submodule commit update, and it's the very last. (If the tests are passing locally but it seems that the Travis checkout is using the wrong pkp-lib, this is probably why.)

PR #3151, opened 2021-06-09
pkp/pkp-lib#7108 load pubIds plugins for executeCLI in ImportExportPl…

withanage:
This is a fix @bozana  did for the pkp/pkp-lib#7108

was only backported to OJS 3.2.1


bozana:
Thanks  lot @withanage!!!

PR #3157, opened 2021-06-23
pkp/pkp-lib#6828 Fixed the section ordering when a new section is added later into an already sorted TOC.

jonasraoni:
I just applied a fast fix, as I believe the grid classes are going to be replaced later.


jonasraoni:
Alternatives/considerations:
- Remove the `custom_section_orders` record when unpublishing/changing issue if there are no other submissions "using" the section (just to remove garbage).
- After publishing a submission or changing its issue (I didn't inspect when both are wired) we could already insert a record into the `custom_section_orders`, in case the issue is already sorted. One possible problem is inserting such record at the same time someone is updating the TOC (I didn't inspect if the ordering breaks when N records end up having the same `seq`). While doing such "auto-ordering", I think it would make sense to respect the order from the `sections.seq`, but trying to find the right spot in a chaotic sorted list becomes confuse 😁

asmecher:
That looks fine, please go ahead and merge!

PR #3162, opened 2021-07-20
pkp/pkp-lib#7129 Issue EntityDAO refactor

ewhanson:


classes/controllers/grid/issues/IssueGridHandler.inc.php
```diff
@@ -370,16 +369,15 @@ public function deleteIssue($args, $request)
...
-            if ($issue = $issues->next()) {
-                $issue->setCurrent(1);
-                $issueDao->updateObject($issue);
+            $issues = Repo::issue()->getPublishedIssues($journal->getId());
+            if ($issue = $issues->first()) {
+                Repo::issue()->setCurrent($issue);
             }
         }
 
+        // TODO: Should this be updated in any way?
```
ewhanson, line 380: This was to a deprecated method. Should this be updated, and if so, what to?

classes/core/Application.inc.php
```diff
@@ -118,6 +118,7 @@ public function getDAOMap()
             'AuthorDAO' => 'APP\article\AuthorDAO',
             'IndividualSubscriptionDAO' => 'APP\subscription\IndividualSubscriptionDAO',
             'InstitutionalSubscriptionDAO' => 'APP\subscription\InstitutionalSubscriptionDAO',
+            // TODO: What should be done with this?
             'IssueDAO' => 'APP\issue\IssueDAO',
```
ewhanson, line 122: I'm not sure what the usage here was. Can this be remove or should it point to the new DAO?

classes/journal/JournalDAO.inc.php
```diff
@@ -116,7 +117,8 @@ public function anyPubIdExists(
         $forSameType = false
     ) {
         $pubObjectDaos = [
-            ASSOC_TYPE_ISSUE => DAORegistry::getDAO('IssueDAO'),
+            // TODO: See if this needs anything else or just updating like publication
+            ASSOC_TYPE_ISSUE => Repo::issue()->dao,
```
ewhanson, line 121: Does anything else need to be considered here when listing the new DAO or is this alright as-is?

tests/classes/search/ArticleSearchTest.php
```diff
@@ -250,6 +250,7 @@ private function registerMockArticleSearchDAO()
     private function registerMockIssueDAO($published = true)
     {
         // Mock an IssueDAO.
+        // TODO: Will this work with Repo/EntityDAO and/or are these tests still used (as they're skipped)?
```
ewhanson, line 253: Two questions here:

1. Does this need to be updated as all the tests here are skipped?
2. What's the best approach for mocking this as the functionality would require the Issue Repository rather than just the DAO.

classes/issue/maps/Schema.inc.php
```diff
@@ -0,0 +1,163 @@
...
+    /** @var Enumerable */
+    public $collection;
+
+    /** @var string */
+    public $schema = PKPSchemaService::SCHEMA_ISSUE;
+
+    public function __construct(PKPRequest $request, Context $context, PKPSchemaService $schemaService)
+    {
+        parent::__construct($request, $context, $schemaService);
+    }
```
NateWr, line 29: You can remove this constructor. The parent will be called automatically if it is missing.

classes/issue/maps/Schema.inc.php
```diff
@@ -0,0 +1,163 @@
...
+                    $submissions = Repo::submission()->getMany(
+                        Repo::submission()
+                            ->getCollector()
+                            ->filterByContextIds([$issue->getJournalId()])
+                            ->filterByIssueIds([$issue->getId()])
+                    );
+
+                    /** @var UserGroupDAO $userGroupsDao */
+                    $userGroupsDao = DAORegistry::getDAO('UserGroupDAO');
+                    $userGroups = $userGroupsDao->getByContextId($this->context->getId());
```
NateWr, line 106: If this is run on a collection of 20 issues, it will retrieve the user groups 20 times. Extract this to a private method that stores the value and only runs it once:

```
private function getUserGroups()
{
    if (!isset($this->userGroups)) {
      $userGroupsDao = DAORegistry::getDAO('UserGroupDAO');
      $userGroups = $userGroupsDao->getByContextId($this->context->getId());
      $this->userGroups = $userGroups->toArray();
    }
    return $this->userGroups;
}
```

classes/journal/JournalDAO.inc.php
NateWr: I think it's alright as-is.

classes/services/IssueService.inc.php
```diff
@@ -19,143 +19,15 @@
 use APP\facades\Repo;
```
NateWr, line 19: When done it should be possible to remove this file completely.

classes/submission/Submission.inc.php
```diff
@@ -93,8 +93,7 @@ public function _getContextLicenseFieldValue($locale, $field, $publication = nul
                         case 'issue':
                             if ($publication->getData('issueId')) {
                                 // override to the issue's year if published as issue-based
-                                $issueDao = & DAORegistry::getDAO('IssueDAO');
-                                $issue = $issueDao->getBySubmissionId($this->getId());
+                                $issue = Repo::issue()->getBySubmissionId($this->getId());
```
NateWr, line 96: I can see you're reproducing the way it was done before, but it doesn't make a lot of sense to go back to the submission via `getBySubmissionId(...)`. We've got the issue ID already in `$publication->getData('issueId')` so we can pull it out with `Repo::issue()->get(...)` directly.

controllers/grid/issues/ExportableIssuesListGridHandler.inc.php
```diff
@@ -42,8 +44,18 @@ public function getSelectName()
...
-        $issueDao = DAORegistry::getDAO('IssueDAO'); /* @var $issueDao IssueDAO */
-        return $issueDao->getIssues($journal->getId(), $this->getGridRangeInfo($request, $this->getId()));
+
+        // Handle grid paging (deprecated style)
+        $rangeInfo = $this->getGridRangeInfo($request, $this->getId());
+        $collector = Repo::issue()->getCollector()
+            ->filterByContext($journal->getId());
+        $totalCount = Repo::issue()->getCount($collector);
+        $collector->limit($rangeInfo->getCount());
+        $collector->offset($rangeInfo->getoffset() + max(0, $rangeInfo->getPage() - 1) * $rangeInfo->getCount());
```
NateWr, line 54: Is `$rangeInfo->getoffset()` supposed to be all lower case?

pages/issue/IssueHandler.inc.php
```diff
@@ -147,16 +146,16 @@ public function archive($args, $request)
...
+        $collector = Repo::issue()->getCollector()
+            ->limit($count)
+            ->offset($offset)
+            ->filterByContext($context->getId())
+            ->orderBy(Collector::ORDERBY_SEQUENCE)
+            ->orderDirection(Collector::ORDER_DIR_ASC)
+            ->filterByPublished(true);
+
+        $issues = Repo::issue()->getMany($collector)->toArray();
+        $total = Repo::issue()->getCount($collector);
```
NateWr, line 158: You'll need to rmeove the limit and offset before calling getCount. I often do this:

```
$collector->filterByX(...);
$total = ...->getCount($collector);
$issues = ...->getMany($collector->limit(...)->offset(...);
```

schemas/issue.json
```diff
@@ -93,6 +112,21 @@
...
+        },
+        "showVolume": {
+            "type": "boolean"
+        },
+        "showYear": {
+          "type": "boolean"
+        },
+        "styleFileName": {
+            "type": "string"
+        },
```
NateWr, line 129: There are some whitespace issues here that need formatting. We use tabs not spaces in the JSON files, I think.

tests/classes/search/ArticleSearchTest.php
NateWr: cc @asmecher

NateWr (changes requested):
Looks good @ewhanson! I've left a few comments here and there and flagged a few things for Alec to follow up on.

tests/classes/search/ArticleSearchTest.php
asmecher: In general I think it should be possible to set up testbeds for database calls by mocking the DAO alone; it's the only place where queries are actually executed.

In this specific case, the `main` branch's `registerMockIssueDAO` just mocks `IssueDAO::getById`; the new equivalent would be `DAO::get` (and `Repository` calls would be passed through to this). I think this can just be adapted as is without much work being required.

classes/issue/DAO.inc.php
```diff
@@ -0,0 +1,512 @@
...
+    /**
+     * Import the current issue orders into the specified journal as custom issue orderings.
+     *
+     */
+    public function setDefaultCustomIssueOrders(int $contextId)
+    {
+        $publishedIssues = Repo::issue()->getPublishedIssues($contextId);
+        $publishedIssues->each(function ($item, $key) {
+        });
+    }
```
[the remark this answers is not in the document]
ewhanson: `setDefaultCustomIssueOrders()` was never called so I've removed it.

controllers/grid/issues/ExportableIssuesListGridHandler.inc.php
ewhanson: Nope. Should be `getOffset()`. Updated.

api/v1/issues/IssueHandler.inc.php
```diff
@@ -152,47 +157,51 @@ public function getMany($slimRequest, $response, $args)
...
+                    $collector->searchPhrase($val);
                     break;
             }
         }
 
-        $params['contextId'] = $context->getId();
+        $collector->filterByContext($context->getId());
 
-        \HookRegistry::call('API::issues::params', [&$params, $slimRequest]);
+        \HookRegistry::call('API::issues::params', [&$collector, $slimRequest]);
```
asmecher, line 188: (Since you're touching on this -- there should be  `use PKP\plugins\HookRegistry;` at the head, and the `\` here should be removed)

classes/controllers/grid/issues/IssueGridHandler.inc.php
```diff
@@ -239,8 +239,7 @@ public function deleteCoverImage($args, $request)
 
         // Check if the passed filename matches the filename for this issue's
         // cover page.
-        $issueDao = DAORegistry::getDAO('IssueDAO'); /* @var $issueDao IssueDAO */
-        $issue = $issueDao->getById((int) $args['issueId']);
+        $issue = Repo::issue()->get((int) $args['issueId']);
```
asmecher, line 242: This affected the old code already, but there's no check here to ensure that the user-specified issue ID is in the current journal.

classes/install/Upgrade.inc.php
```diff
@@ -309,7 +304,7 @@ public function localizeIssueCoverImages()
                 break;
             default: fatalError('Unknown database type!');
         }
-        $issueDao->flushCache();
+        $deprecatedIssueDao->flushCache();
```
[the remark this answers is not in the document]
asmecher: The upgrade scripts generally avoid calling on DAO functions (because these will evolve with new code and break older upgrade steps), so we often worked around this by instantiating any old DAO then calling `::update` et al with SQL coded directly. You can adapt this to Laravel tools without problems (and without having to rewrite everything using a query builder) by using `DB::select`, `DB::update`, `DB::delete` instead of `DAO::retrieve` and `DAO::update`.

classes/issue/Repository.inc.php
```diff
@@ -0,0 +1,368 @@
...
+            $this->getCollector()->filterByContext($contextId)->filterByCurrent()
+        )->first();
+        if (!is_null($previousCurrentIssue)) {
+            $this->edit($previousCurrentIssue, ['current' => 0]);
+        }
+
+        if ($newCurrentIssue) {
+            $this->edit($newCurrentIssue, []);
+        }
+    }
```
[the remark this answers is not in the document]
asmecher: Yes, that's better database design.

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,460 @@
...
+     * @inheritDoc
+     */
+    public function getQueryBuilder(): Builder
+    {
+        $this->columns[] = 'i.*';
+        $q = DB::table($this->dao->table, 'i')
+            ->leftJoin('issue_settings as iss', 'i.issue_id', '=', 'iss.issue_id')
+            ->leftJoin('custom_issue_orders as o', 'o.issue_id', '=', 'i.issue_id')
+            // TODO: Is this still necessary (not just for search phrase, but also others)
+            ->groupBy('i.issue_id', $this->orderColumn);
```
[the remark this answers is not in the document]
asmecher: The way it's currently coded, yes; if there are multiple `issue_settings` rows for a single `issues` entry (which is likely), the `LEFT JOIN` will result in several rows for each issue, thus the `GROUP BY` is being used to collapse the result set into a single row per issue.

To resolve this, you'll need to make sure you are only making joins that do not result in multiple rows matching. For example, there will never be more than one more row per issue for...
```
SELECT i.* FROM issues i JOIN journals j ON (i.journal_id = j.journal_id);
```
However, this will potentially return several rows per issue:
```
SELECT i.* FROM issues i JOIN issue_settings iss ON (i.issue_id = iss.issue_id);
```
It looks like the `iss` join has already been converted below to a `WHERE i.issue_id IN (some subquery)`, so I think the `iss` join above is actually not used.

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,460 @@
...
+                    }
+                });
+            }
+
+            $words = array_unique(explode(' ', $searchPhrase));
+            if (count($words)) {
+                foreach ($words as $word) {
+                    $word = strtolower(addcslashes($word, '%_'));
+                    $q->whereIn('i.issue_id', function (Builder $q) use ($word) {
+                        $q->
```
asmecher, line 416: Typo!

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,460 @@
...
+                        $q->select('iss.issue_id')
+                            ->from($this->dao->settingsTable, 'iss_t')
+                            ->where('iss_t.setting_name', '=', 'title')
+                            ->where(DB::raw('lower(iss_t.setting_value'), 'LIKE', "%${word}%");
```
asmecher, line 420: Need to call `addcslashes` to escape `_` and `%` special characters that might appear in the supplied keywords. See e.g.: https://github.com/asmecher/pkp-lib/blob/7b06a009eb7db9e0d7df80d19a93b91e73462025/classes/user/Collector.inc.php#L219

asmecher: Oops, I see you've already done that above!

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,460 @@
...
+        $q = DB::table($this->dao->table, 'i')
+            ->leftJoin('issue_settings as iss', 'i.issue_id', '=', 'iss.issue_id')
+            ->leftJoin('custom_issue_orders as o', 'o.issue_id', '=', 'i.issue_id')
+            // TODO: Is this still necessary (not just for search phrase, but also others)
+            ->groupBy('i.issue_id', $this->orderColumn);
+
+        // Issue titles (exact matches)
+        foreach ($this->titles as $key => $title) {
+            $tableId = 'iss' . $key;
+            $q->join('issue_settings as ' . $tableId, function (JoinClause $join) use ($tableId, $title) {
```
asmecher, line 316: This should probably be converted to a `WHERE i.issue_id IN (subquery)` form; otherwise multiple matches here will result in multiple rows, which need to then be resolved using `GROUP BY`.

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,460 @@
...
+            $q->whereIn('i.number', $this->numbers);
+        }
+
+        // Years
+        if (!is_null($this->years)) {
+            $q->whereIn('i.year', $this->years);
+        }
+
+        // Issue ids
+        if (!empty($this->issueIds)) {
```
asmecher, line 360: I'd like to firm up our handling of "falsy" variables as we start to make more use of lists in place of single IDs. I documented this in Slack but I'll put it here for when that disappears:

For conditional SQL, we've historically used PHP `if` statements and built a `$params` array...
```
$params = [];
if ($contextId) $params[] = (int) $contextId;
return new DAOResultFactory($this->retrieve(
    'SELECT ... WHERE ...' .
    ($contextId ? ' AND context_id = ?' : ''),
    $params));
```

Some problems with this:

- It means dealing with $params twice, once when building the $params array and again when building the SQL query. Lots of conditions = messy!
- It's messy about null vs false vs 0 etc.

Laravel query builders improve on this:

```
$query = DB::table('...');
if ($contextId) $query->where('context_id', '=', $contextId);
```

I propose two additional patterns:

- Use Laravel's conditional SQL tool (->when)
- Get strict about "falsy" values. Use null to distinguish between "we don't care" and "we do but no acceptable values are listed". (See below.)

An example using the proposed patterns:

```
$query = DB::table('...')
    ->when($this->contextId !== null, function($query) {
        $query->where('context_id', '=', $this->contextId);
    });
```

This avoids a gotcha, as empty lists are falsy could be misinterpreted as "we don't care". Consider calling code:

```
// We use two conditions to build up a list of allowed user groups:
// 1. Get a list of the current user's groups from the database
$allowedUserGroupsA = [1, 2, 3];
// 2. Get a list of groups the user has specified via the UI (e.g. checkboxes)
$allowedUserGroupsB = [4, 5, 6];
// Combine the two conditions using array_intersect.
$allowedUserGroups = array_intersect($allowedUserGroupsA, $allowedUserGroupsB);
// This will result in an empty list, meaning none of the specified user groups were allowed!
$collector->filterByUserGroupIds($allowedUserGroups);
// We would expect the DAO to return no results.
return $dao->getMany($collector);
```

However, if the collector follows this pattern...

```
// BAD PATTERN BELOW: Empty lists are skipped
$query = DB::table('...')
    ->when($this->userGroupIds, function($query, $userGroupIds) {
        $query->whereIn('user_group_ids', $userGroupIds);
    });
```

...then the `[]` value of `$userGroupIds` means the condition will not be added at all and all user groups will be returned! (Note that this will often mean avoiding the use of the second parameter in lambda functions to `$query->when`. Use `$this` or a `use` clause instead.)

As we start to rely more on arrays when specifying conditions, this kind of gotcha will become more prominent!

So to summarize...

1. In your collector class, initialize constraints to null. Think of this as a reserved value for "we don't care".
   ```
   /** ?array A list of context IDs to allow */
   var $contextIds = null;
   ```
2. When calling the filter function, an array is required, but it could be empty:
   ```
   function filterByContextIds(array $contextIds) {
       $this->contextIds = $contextIds;
   }
   ```
3. When building the query, make sure to treat null values carefully.
   ```
   $query = DB::table(...)
      ->when($this->contextIds !== null, function($query) {
           $query->whereIn('context_id', $this->contextIds
       });
   ```

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,460 @@
...
+    protected $dao;
+
+    /** @var int */
+    protected $count = 30;
+
+    /** @var int */
+    protected $offset = 0;
+
+    /** @var array List of columns to select with query */
+    protected $columns = [];
```
asmecher, line 31: @NateWr, the goal of this is to make the column list extensible by outside code, right? I think this is duplicating some query builder functionality we can use. At the top of the `getQueryBuilder` function, we can write...
```
$q = DB::table($this->dao->table, 'i')
    ->select('i.*')
    ...
```
...then any extending code wanting to add a column can write...
```
$q->addSelect('some_other_table.some_other_column AS some_extra_column');
```
(I'm using this successfully in the reviewer extensions to the user stuff.)

NateWr: Yes, I think that we can remove the use of `$columns`. I think it was probably related to some conditional joins or something, but shouldn't be needed.

classes/issue/DAO.inc.php
```diff
@@ -0,0 +1,512 @@
...
+    {
+        $q = DB::table('custom_issue_orders')
+            ->where('journal_id', '=', $contextId);
+
+        if (!is_null($issueId)) {
+            $q->where('issue_id', '=', (int) $issueId);
+        }
+
+        $q->delete();
+    }
```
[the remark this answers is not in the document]
ewhanson: Yes, this is used in the `Repo::issue()->delete()`.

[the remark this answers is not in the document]
NateWr: In that case, the function should be `protected` (since it's only used internally) and only the `$issueId` is needed.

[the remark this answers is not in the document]
ewhanson: Thanks, I removed `$contextId` and made `$issueId` required. I've kept it public for now as it's a DAO function called by the repository. I could move it to the repository, but all of the other custom issue ordering logic is in the DAO.

ewhanson:
> Because we're working with the `issues` row, and every issue will always be associated with a journal, then it's best to just use a `join` rather than a `leftJoin`. (`leftJoin` will only be necessary when the table you're joining on may or may not have a match, but you still want results anyway.) You could do e.g. [untested but should be OK]...
> 
> ```
> SELECT * FROM issues i JOIN journals j ON (i.issue_id = j.current_issue_id) ORDER BY CASE WHEN j.current_issue_id IS NOT NULL THEN 1 ELSE 0 END DESC;
> ```

Thanks, @asmecher! I ended up needing to use a `leftJoin` as ordering by current issue wasn't always the top item in the list and doing a `join` would only returned the current issue. How does this implementation here https://github.com/pkp/ojs/pull/3162/commits/0fc51ed4178c2532a4553229e72321399a3c4450 look?

asmecher:
That looks good, @ewhanson; I'd suggest moving the join down adjacent to where it's used, but otherwise good to go!

ewhanson:
> That looks good, @ewhanson; I'd suggest moving the join down adjacent to where it's used, but otherwise good to go!

Thanks, @asmecher. Done!

That was the last item on my todo list for this. The only outstanding question left was around the caching. If everything else looks good and is ready for merge, I'm wondering if it would be better to get this in and address the caching on a separate PR. What are your thoughts on this?

pages/article/ArticleHandler.inc.php
```diff
@@ -187,8 +187,10 @@ public function initialize($request, $args = [])
         }
 
         if ($this->publication->getData('issueId')) {
-            $issueDao = DAORegistry::getDAO('IssueDAO'); /* @var $issueDao IssueDAO */
-            $this->issue = $issueDao->getById($this->publication->getData('issueId'), $submission->getData('contextId'), true);
+            // TODO: Previously--useContext = true
```
asmecher, line 190: (not sure if you missed this TODO)

pages/gateway/GatewayHandler.inc.php
```diff
@@ -76,12 +78,12 @@ public function lockss($args, $request)
 
             $year = $request->getUserVar('year');
 
-            $issueDao = DAORegistry::getDAO('IssueDAO'); /* @var $issueDao IssueDAO */
+            $deprecatedDao = Repo::issue()->dao->deprecatedDao;
```
asmecher, line 81: I think this code still needs fixing.

asmecher (changes requested):
Thanks, @ewhanson! This is very close, but I found a few more things needing attention. Beyond that, I'm OK to work on the caching stuff outside of this PR.

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,429 @@
...
+    public const ORDERBY_PUBLISHED_ISSUES = 'publishedIssues';
+    public const ORDERBY_UNPUBLISHED_ISSUES = 'unpublishedIssues';
+    public const ORDERBY_GATEWAY_LOCKSS = 'gatewayLockss';
+    public const ORDER_DIR_ASC = 'ASC';
+    public const ORDER_DIR_DESC = 'DESC';
+
+    /**  */
+    public DAO $dao;
+
+    public ?int $count = 30;
```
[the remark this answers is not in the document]
ewhanson: @NateWr, I know we discussed this, but I don't recall where we landed with this.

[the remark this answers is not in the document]
NateWr: Yeah, initialize this to `null`. (The old QueryBuilders did it differently, but we're changing our approach.)

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,429 @@
...
+                        'orderBy' => 'i.volume',
+                        'direction' => self::ORDER_DIR_ASC
+                    ],
+                    [
+                        'orderBy' => 'i.number',
+                        'direction' => self::ORDER_DIR_ASC
+                    ]
+                ];
+                break;
+            case self::ORDERBY_GATEWAY_LOCKSS:
```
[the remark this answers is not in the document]
NateWr: This isn't necessary for the PR, but I wanted to add a comment about how plugins can augment the collectors without modifying their code. I suspect the code that calls this sits in the LOCKSS plugin and looks like this:

```php
$collector = Repo::issue()->getCollector();
$collector->orderBy($collector::ORDERBY_SHELF);
$issues = Repo::issue()->getMany($collector);
```

If a plugin needs to do some custom handling here, there are two options. The first is to use the hook to extend the query. That would look like:

```php
$collector = Repo::issue()->getCollector();
$collector->orderBy = 'customOrder';
```

With a hook like:

```php
HookRegistry::register('Issue::Collector', function($hookName, $args) {
    $q = $args[0];
    $collector = $args[1];

    if ($collector->orderBy === 'customOrder') {
        $q->orderBy(...); // Whatever you want.
    }
});
```

However, this is only necessary when you want the custom ordering to be available for for any other code -- for example if another plugin depends on this one, or if this plugin calls this code a lot and you want to standardize it. 

When a plugin only needs to change the query builder for a one-off use case, it can get the Laravel query builder instance and use it's methods directly:

```php
$collector = Repo::issue()->getCollector();
$q = $collector->getQueryBuilder();

// Use Laravel's query builder to set an order and get rows
$q->orderBy(...);
$rows = $q->get();

// Use Laravel's collection methods with the DAO to get objects
$issues = $rows->map(function($row) {
    return Repo::issue()->dao->fromRow($row);
});
```

I think that this kind of one-off usage is ok, especially in plugins, when we don't necessarily want this custom result to be available to everyone.

ewhanson:
Thanks @asmecher! I've taken these collector best practices into account and rebased onto the User DAO changes.

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,408 @@
+<?php
+
+namespace APP\issue;
+
+use HookRegistry;
```
asmecher, line 5: Should be `use PKP\plugins\HookRegistry;`

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,408 @@
...
+        $this->contextIds = $contextIds;
+        return $this;
+    }
+
+    /**
+     * Set result order column
+     *
+     * @return $this
+     */
+    public function orderBy(string $column): self
```
asmecher, line 77: `$column` no longer refers to a column name -- it's a set of defined constants -- so I'd suggest changing its name. Also the self-documentation.

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,408 @@
...
+                        'orderBy' => 'i.volume',
+                        'direction' => self::ORDER_DIR_ASC
+                    ],
+                    [
+                        'orderBy' => 'i.number',
+                        'direction' => self::ORDER_DIR_ASC
+                    ]
+                ];
+                break;
+            default:
```
asmecher, line 148: Rather than having a default that's always accepted, I'd suggest making all the valid calls to this function defined constants, and having `default: throw new \InvalidArgumentException(...)`

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    /**
+     * Set volumes filter
+     *
+     * @param int[] $volumes
+     *
+     * @return $this
+     */
+    public function filterByVolumes(array $volumes): self
```
asmecher, line 178: This should be `?array`, allowing calling code to reset the filter entirely (also below)

classes/issue/Collector.inc.php
```diff
@@ -0,0 +1,408 @@
...
+     * @inheritDoc
+     */
+    public function getQueryBuilder(): Builder
+    {
+        $q = DB::table($this->dao->table, 'i')
+            ->select('i.*')
+            ->leftJoin('custom_issue_orders as o', 'o.issue_id', '=', 'i.issue_id');
+
+        // Issue titles (exact matches)
+        foreach ($this->titles as $title) {
```
asmecher, line 276: I think this could be done with a single join, rather than one join per `title`:
```SQL
WHERE i.issue_id IN (SELECT issue_id FROM issue_settings WHERE setting_name='title' AND setting_value IN (?, ?, ?)
```

asmecher (changes requested):
Sorry, @ewhanson, they are so minor I feel like I'm just meddling -- please have a quick look through them, and once they're resolved, I'm content to merge this as is. Thanks, it's a lot of very clean-looking work!

ewhanson:
Thanks @asmecher! All taken care of now.

classes/issue/DAO.inc.php
```diff
@@ -0,0 +1,484 @@
...
+
+    /**
+     * Get a collection of issues matching the configured query
+     *
+     */
+    public function getMany(Collector $query): LazyCollection
+    {
+        $rows = $query
+            ->getQueryBuilder()
+            ->get();
```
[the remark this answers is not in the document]
NateWr: :+1: That's great! I'm hoping that the new structure provides room for these kinds of performance improvements.

The `LazyCollection` here is still useful because this `get()` only returns the row data. The `fromRow` method, which instantiates the `Issue` objects and often runs additional queries), won't be called until the collection is looped over.

PR #3174, opened 2021-09-01
pkp/pkp-lib#7247 Fix bugs introduced in Issue EntityDAO refactor

ewhanson:
Hi @asmecher, thanks for reviewing these!

This addresses a typo in the collector, removing the "current issue" status from an issue when its unpublished, and updating the CLOCKSS refactor to match that of LOCKSS.

### DIFF
--- a/pages/gateway/GatewayHandler.inc.php
+++ b/pages/gateway/GatewayHandler.inc.php
@@ -75,7 +75,7 @@ class GatewayHandler extends Handler
             if (!$journal->getData('enableLockss')) {
                 $request->redirect(null, 'index');
             }
-            $yearsIssuesPublished = Repo::issue()->getYearsIssuesPublished($journal->getId());
+            $yearsIssuesPublished = Repo::issue()->getYearsIssuesPublished($journal->getId())->values();
 
             // FIXME Should probably go in IssueDAO or a subclass
             $year = $yearsIssuesPublished->contains((int) $request->getUserVar('year'))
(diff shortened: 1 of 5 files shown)


pages/gateway/GatewayHandler.inc.php
```diff
@@ -175,19 +161,13 @@ public function clockss($args, $request)
...
-                    'SELECT MIN(year) AS min_year FROM issues WHERE journal_id = ? AND published = 1 AND year > ?',
-                    [$journal->getId(), $year]
-                );
-                $row = $result->current();
-                $nextYear = $row ? $row->min_year : null;
+                $key = $yearsIssuesPublished->search(function ($i) use ($year) {
+                    return $i === $year;
+                });
+                if (isset($key)) {
+                    $prevYear = $yearsIssuesPublished->get($key - 1);
```
asmecher, line 168: It looks like this is counting on database IDs being sequential, right? I'd rather not assume this -- e.g. an issue could be deleted.

ewhanson: Yes, thanks for flagging this. That was the case. I'm switching it to use `Repo::issue()->getYearsIssuesPublished($journal->getId())->values()` when getting the collection at the start to reset the keys so this won't be a problem. Doing this for LOCKSS and for CLOCKSS.

ewhanson:
Hey @asmecher, could you have another look at this? I've updated it as described above. Thanks!

asmecher:
Thanks, @ewhanson -- I went ahead and merged it!

PR #3180, opened 2021-09-13
pkp/pkp-lib#6093 Add foreign key constraints

asmecher:


classes/migration/upgrade/OJSv3_3_1_i6872_UrlPathMigration.inc.php
```diff
@@ -15,23 +15,22 @@
...
+    public function up() : void
     {
         DB::table('publications')->whereNull('url_path')->update(['url_path' => null]);
     }
 
     /**
      * Reverse the downgrades
      */
-    public function down()
+    public function down() : ?bool
```
Vitaliy-1, line 33: The method doesn't return anything, will throw a fatal error.

PR #3194, opened 2021-09-28
pkp/pkp-lib#4235 Don't output invalid chars in OAI setSpec

olovy:
Issue: https://github.com/pkp/pkp-lib/issues/4235
Depends on corresponding PR in pkp-lib https://github.com/pkp/pkp-lib/pull/7345

Valid characters in setSpec [1] are URI unreserved characters [2]
(`a-z`, `A-Z`, `0-9`, `-_.!~*'()`) Currently, OJS creates the setSpec
string by running `urlencode` on journal path and section abbreviation.
This produces invalid character, e.g. `%`, and the OAI-PMH feed fails
to be harvested by clients with strict validation.

This change assumes that journal path is validated to be ASCII
alphanumeric, '-' and '_' elsewhere.

Section abbreviation can be a user supplied containing any Unicode character.
Make it into a valid setSpec string by
- transliterating all non-Latin characters
- stripping all diacritics
- dropping any remaining invalid characters

This will of course not produce a meaningful result for many languages
but it will be a valid setSpec string.

Examples:
"innehåll" -> "innehall"
"Ελληνικά" -> "Ellenika"
"ひらがな" -> "hiragana"

Depends on php-intl extension (which I had to install to get OJS running so should be available?)

[1] https://www.openarchives.org/OAI/2.0/openarchivesprotocol.htm#Set
[2] http://www.ietf.org/rfc/rfc2396.txt

### DIFF
--- a/classes/migration/upgrade/v3_4_0/I4235_OAISetSpec.inc.php
+++ b/classes/migration/upgrade/v3_4_0/I4235_OAISetSpec.inc.php
@@ -43,7 +43,3 @@ class I4235_OAISetSpec extends \PKP\migration\Migration
         // The old format is not recoverable since some characters might have been stripped
     }
 }
-
-if (!PKP_STRICT_MODE) {
-    class_alias('\APP\migration\upgrade\v3_4_0\I4235_OAISetSpec', '\I4235_OAISetSpec');
-}
--- a/classes/migration/upgrade/v3_4_0/I4235_OAISetSpec.inc.php
+++ b/classes/migration/upgrade/v3_4_0/I4235_OAISetSpec.inc.php
@@ -1,0 +1,49 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_4_0/I4235_OAISetSpec.inc.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
+ * Copyright (c) 2000-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class I4235_OAISetSpec
+ */
+
+namespace APP\migration\upgrade\v3_4_0;
+
+use Illuminate\Support\Facades\DB;
+use PKP\oai\OAIUtils;
+
+class I4235_OAISetSpec extends \PKP\migration\Migration
+{
+    /**
+     * Run the migration.
+     */
+    public function up(): void
+    {
+        // pkp/pkp-lib/issues/4235 Improve OAI-PMH set spec compliance
+        // Convert stored setSpec strings to valid format
+        $setSpecs = DB::table('data_object_tombstones')->select('set_spec')->get()->toArray();
+        foreach ($setSpecs as $row) {
+            $a = preg_split('/:/', $row->set_spec);
+            if (count($a) == 2) {
+                [$journalSpec, $sectionSpec] = $a;
+                $new = OAIUtils::toValidSetSpec(urldecode($journalSpec)) . ':' . OAIUtils::toValidSetSpec(urldecode($sectionSpec));
+                DB::table('data_object_tombstones')->where('set_spec', $row->set_spec)->update(['set_spec' => $new]);
+            }
+        }
+    }
+
+    /**
+     * Reverse the downgrades
+     */
+    public function down(): void
+    {
+        // The old format is not recoverable since some characters might have been stripped
+    }
+}
+
+if (!PKP_STRICT_MODE) {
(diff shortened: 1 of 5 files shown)


dbscripts/xml/upgrade.xml
```diff
@@ -216,6 +216,10 @@
 		<data file="dbscripts/xml/upgrade/3.4.0_preupdate_email_templates.xml" />
 		<note file="docs/release-notes/README-3.4.0" />
 	</upgrade>
+        
+	<upgrade minversion="3.0.0.0" maxversion="3.4.9.9">
```
olovy, line 220: Not sure about this

classes/migration/upgrade/v3_4_0/I4235_OAISetSpec.inc.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Reverse the downgrades
+     */
+    public function down(): void
+    {
+        // The old format is not recoverable since some characters might have been stripped
+    }
+}
+
+if (!PKP_STRICT_MODE) {
```
asmecher, line 47: This is strictly for compatibility for old code from the 3.3 to 3.4 upgrade -- you don't need to add it for new code.

classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -172,20 +173,30 @@ public function getSetJournalSectionId($journalSpec, $sectionSpec, $restrictJour
...
         if (isset($sectionSpec)) {
-            $section = $this->sectionDao->getByAbbrev($sectionSpec, $journal->getId());
-            if (isset($section)) {
-                $sectionId = $section->getId();
-            } else {
-                $sectionId = 0;
+            $sectionIterator = $this->sectionDao->getByJournalId($journalId);
+
+            while ($section = $sectionIterator->next()) {
+                if ($sectionSpec == $this->toValidSetSpec($section->getLocalizedAbbrev())) {
```
asmecher, line 182: Was this meant to be a call to `OAIUtils::toValidSetSpec`?

asmecher (changes requested):
Thanks, @olovy! I've added a few comments; also this change would need to be ported to OMP and OPS. Do you have experience with those applications?

classes/oai/ojs/OAIDAO.inc.php
olovy: Oops, missed this one when moving the method

classes/oai/ojs/OAIDAO.inc.php
```diff
@@ -172,20 +173,30 @@ public function getSetJournalSectionId($journalSpec, $sectionSpec, $restrictJour
         }
 
         $journalId = $journal->getId();
-        $sectionId = null;
+        $sectionId = 0;
```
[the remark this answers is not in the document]
olovy: Yes, of course. You are right.

olovy:
> Thanks, @olovy! I've added a few comments; also this change would need to be ported to OMP and OPS. Do you have experience with those applications?

Hi!
I have zero experience with OMP and OPS. But I also had almost zero experience with OJS (and PHP...) before doing this.
It wasn't part of our original planning but I will check if I have time to look at it.

olovy:
> > Thanks, @olovy! I've added a few comments; also this change would need to be ported to OMP and OPS. Do you have experience with those applications?
> 
> Hi! I have zero experience with OMP and OPS. But I also had almost zero experience with OJS (and PHP...) before doing this. It wasn't part of our original planning but I will check if I have time to look at it.

For OMP I think we can just remove `urlencode`/`urldecode` if we can depend on both paths always
being validated by `validator.alpha_dash_period` like I do for journal paths here? Both strings should always be ok.
(Just guessing, didn't check if they are validated in the same way)
https://github.com/pkp/omp/blob/3a3d2d342644f93ded511c15ed99a80868a2a4ed/classes/oai/omp/OAIDAO.inc.php#L188

OPS looks like a straightforward port of the fix for OJS.

olovy:
I guess it should be noted in the release notes that the setSpec might change. 
In case there are any harvesters configured to fetch sections that used to have a bad setSpec.

We could make the lookup backwards compatible but I'm not sure it is necessary.
i..e `/oai...&setSpec=foo:inneh%C3%A5ll` would give the same result as `/oai...&setSpec=foo:innehall`

asmecher:
Yes, you're right about the comparison with OMP and OPS -- if you're able, could you open PRs for each of them too?

olovy:
> Yes, you're right about the comparison with OMP and OPS -- if you're able, could you open PRs for each of them too?

Done

asmecher:
Merged manually (cherry-picking). Thanks, @olovy!

PR #3197, opened 2021-09-29
Add p.last_modified to update the last datestamp.

doortegaa:
In an OJS 3.3.0-7 from where a harvest server, harvest the metadata of published articles, the harvest server doesn’t identify the updated articles in OAI-PMH because the datestamp isn’t updated to the date the article was updated.

My proposal to solve it, is to modify in the file classes/oai/ojs/OAI DAO.inc.php, adding p.last modified on the querys. With this change, the OAI already shows the datestamp with the date of the last_modified field of the publications table and vufind already identifies it as an updated article to re-harvest it.

There are more information in: https://forum.pkp.sfu.ca/t/ojs-3-3-0-7-oai-doesnt-update-the-datestamp-to-the-date-of-the-last-update-to-allow-metadata-harvest/69677

Regards

### DIFF
--- a/classes/oai/ojs/OAIDAO.inc.php
+++ b/classes/oai/ojs/OAIDAO.inc.php
@@ -240,7 +240,7 @@ class OAIDAO extends PKPOAIDAO
 
         return DB::table('submissions AS a')
             ->select([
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks, @doortegaa! I've also ported this back to `stable-3_3_0` for the next build.

PR #3199, opened 2021-10-01
Remove mention of precompiled js for stable branches

OyvindLGjesdal:
Also change from master to main
https://forum.pkp.sfu.ca/t/ojs-build-js-is-not-present-in-the-stable-branches/70034/2

### DIFF
--- a/docs/UPGRADE.md
+++ b/docs/UPGRADE.md
@@ -158,12 +158,11 @@ web-based script. To do so:
 
 ### Update Javascript libraries and build.js
 
-The official .tar.gz releases, and the stable branches in git (e.g.
-`ojs-stable-3_1_1`), contain precompiled javascript. If you are installing
-OJS using either of those and have not modified your Javascript, you do not
-need to compile Javascript.
+The official .tar.gz releases contain precompiled javascript. If you are installing
+OJS using this method, you do not need to compile Javascript.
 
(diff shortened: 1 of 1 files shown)


docs/UPGRADE.md
```diff
@@ -158,12 +158,11 @@ web-based script. To do so:
...
-The official .tar.gz releases, and the stable branches in git (e.g.
-`ojs-stable-3_1_1`), contain precompiled javascript. If you are installing
-OJS using either of those and have not modified your Javascript, you do not
-need to compile Javascript.
+The official .tar.gz releases contain precompiled javascript. If you are installing
+OJS using this method, you do not need to compile Javascript.
 
-If you are using the git `master` branch, or have made changes to your
+If you are using the git `main` branch, or any of the stable branches (e.g.
+`ojs-stable-3_3_0`), or have made changes to your
```
asmecher, line 165: This should be `stable-3_3_0`, not `ojs-stable-3_3_0`, I think.

OyvindLGjesdal: Updated to `stable-3_3_0`

asmecher:
Thanks, @OyvindLGjesdal! We may go back to adding compiled JS to the stable branches (that was our intention but we didn't practice it), but until then, this makes the documentation accurate.

PR #3202, opened 2021-10-06
pkp/pkp-lib#7595 Allow valid special characters in doi suffix

dennmuel:
Hi!

The regex for the DOI suffix currently removes characters that are [rendered valid by Crossref](https://www.crossref.org/documentation/member-setup/constructing-your-dois/#afewrules), in particular: "hyphen, period, underscore, semicolon, parentheses, forward slash". This was introduced in / is related to pkp/pkp-lib/issues/5505 and #2620

This PR suggests an (untested!) regex based on [this post](https://www.crossref.org/blog/dois-and-matching-regular-expressions/).

Best regards!


asmecher:
@ewhanson, no, I'm not aware of other PID types that would be broken by this relaxation.

@dennmuel, have you looked at the additional case @ewhanson identified above?

dennmuel:
Sorry for the late reply!
I added the regex to DOIPubIdPlugin.inc.php as well.

PR #3205, opened 2021-10-18
pkp/pkp-lib#7282 Forcing IsoCodesFactory to use PKPLocale::getLocale()

henriqueramos:


asmecher:
(I added https://github.com/pkp/ojs/commit/63e7d93d5f20af87476812375c7617318a7ecdce to fix a missing `use` statement.)

PR #3243, opened 2021-12-06
pkp/pkp-lib#7266 Assigned articles missing from future issues

ewhanson:


dbscripts/xml/upgrade/3.2.0_versioning.xml
```diff
@@ -221,6 +221,46 @@
 				AND plugin_name='urnpubidplugin'
 		</query>
 
+		<!-- pkp/pkp-lib#7266 Assigned articles missing from future issues after upgrade to OJS 3.3.0-8 -->
```
NateWr, line 224: Can you add a reference to the constants that are used into the comment here? That helps us track where a constant is used across the code base. So something like:

ewhanson: Thanks, I've updated this here and in the `stable_3_3_0` PR.

PR #3275, opened 2022-01-13
pkp/pkp-lib#7016 Restore check for author include in browse list

asmecher:


asmecher:
Needs to be synced with changes in https://github.com/pkp/ojs/pull/3273/files!

PR #3297, opened 2022-02-11
Adding DOI field to MarcXML OAI output

GrazingScientist:
We realized in our team that the MARC-XML of the OAI-API did not return the article's DOI. Hence, here is the fix.

Also there were two wrongly named subfields that were fixed.

All kudos to @j4lib.


asmecher:
Thanks, @GrazingScientist and @j4lib!

@ewhanson, will this need adaptation for the new DOI code?

ewhanson:
@asmecher, nope! There's backwards compatibility with with `getStoredPubId('doi')` so it won't need any changes.

asmecher:
Thanks, all! Merged to `main` and cherry-picked back to `stable-3_3_0`.

PR #3299, opened 2022-02-14
[OJS] #5048 Change <data> upgrade script to Migrations

defstat:


asmecher:
@defstat, I think this is also obsolete, can you confirm?

PR #3303, opened 2022-02-16
Fix upgrade error

murilodbva:
It solves failt in citations TABLE on upgrading.
See: [https://forum.pkp.sfu.ca/t/upgrade-ojs-3-1-2-1-to-3-2-1-4-successfully-upgraded/69850](https://forum.pkp.sfu.ca/t/upgrade-ojs-3-1-2-1-to-3-2-1-4-successfully-upgraded/69850)

### DIFF
--- a/dbscripts/xml/upgrade/3.2.0_versioning.xml
+++ b/dbscripts/xml/upgrade/3.2.0_versioning.xml
@@ -183,6 +183,7 @@
 		<query driver="postgres,postgres64,postgres7,postgres8,postgres9"><![CDATA[
 			UPDATE citations SET publication_id = -publication_id WHERE publication_id < 0
 		]]></query>
+		<query driver="mysql,mysqli">ALTER TABLE citations MODIFY publication_id bigint(20) DEFAULT 0</query>
 		<query driver="mysql,mysqli">
 			UPDATE citations as c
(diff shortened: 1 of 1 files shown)


asmecher:
Thanks for contributing, @murilodbva! We need to maintain feature parity with PostgreSQL; are you able to adapt/test this so it also works with PostgreSQL? (If the same SQL won't work with both MySQL and PostgreSQL, you can specify different sets of queries for each.)

asmecher:
Rather than converting between DBMSs (which can be tricky), maybe try using one of the sample datasets published here? https://github.com/pkp/datasets

murilodbva:
> Thanks for contributing, @murilodbva! We need to maintain feature parity with PostgreSQL; are you able to adapt/test this so it also works with PostgreSQL? (If the same SQL won't work with both MySQL and PostgreSQL, you can specify different sets of queries for each.)

Hello @asmecher , I did what you requested me. I tested for PostgreSQL, so I kept what works for the 2, and included scripts for PostgreSQL and worked fine.
Tell me if you understand (or not) what I did. =)

asmecher:
Thanks, @murilodbva! I'll leave this to @Vitaliy-1 to review, since I think he was helping you on the forum with the same subject.

PR #3321, opened 2022-03-07
Allow journals to disable author biography

kmcarnes:
There were some journals that wanted to have the option to disable the Author Biography section.

We added the option to Website -> Appearance -> Setup.

It appears that there are others that would like this option (https://forum.pkp.sfu.ca/t/switch-off-author-biography-on-article-page/52724).

Please feel free to give feedback or edit this.

### DIFF
--- a/classes/components/forms/context/AppearanceSetupForm.inc.php
+++ b/classes/components/forms/context/AppearanceSetupForm.inc.php
@@ -14,6 +14,7 @@
  */
 namespace APP\components\forms\context;
 use \PKP\components\forms\context\PKPAppearanceSetupForm;
+use \PKP\components\forms\FieldOptions;
 use \PKP\components\forms\FieldUploadImage;
 
 class AppearanceSetupForm extends PKPAppearanceSetupForm {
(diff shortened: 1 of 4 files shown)


asmecher:
Hi @kmcarnes! Taken individually, options like this are not a bad idea -- but it is a pretty niche option and collectively this kind of feature leads to UI clutter (the OJS 2.x setup forms really got out of control in this way). I think this would be better implemented as a child theme or other form of plugin and maintained separately from the OJS codebase.

kovanovic:
We had an issue recently with learning-analytics.info/ journal.

We do not use author biographies, and if they are left empty, they do not show on the page. However, for one of the papers, even though the bio is empty, it is still showing up. In the end, I added a CSS rule to remove the autors_bios block so it works. Not sure why it didn't work correctly

asmecher:
@kmcarnes, you were probably looking at the author's user account, which is different than the author record. Edits to the author's user account don't affect author records in already-published content because it's part of the scholarly record. These records would need to be edited individually on each submission the author has made.

PR #3355, opened 2022-04-01
pkp/pkp-lib#7817 Add missing Datacite CSRF token

ewhanson:


asmecher:
Whoops, already cherry-picked this over from `main` before I saw the second PR. Thanks, @ewhanson!

PR #3369, opened 2022-04-10
pkp/pkp-lib/#7395 Delete OneClick email template varaibles

withanage:
#### Issue

https://github.com/pkp/pkp-lib/issues/7395


Vitaliy-1:
Here are localizations: https://github.com/pkp/ojs/blob/3eca01d5f798664928444f5c53fdd9bc20f4edbd/locale/en_US/emails.po#L446-L462

I left those templates in a migration script that renames email template variables as it's executed before another script, `I7592_RemoveUnusedEmailTemplates`, which actually removes them.

PR #3399, opened 2022-05-11
pkp/pkp-lib#7921: Make buildpkg.sh more flexible

ctgraham:
Resolves pkp/pkp-lib#7921 from stable-3_3_0.

### DIFF
--- a/tools/buildpkg.sh
+++ b/tools/buildpkg.sh
@@ -13,16 +13,16 @@
 #
 #
 
-GITREP=git://github.com/pkp/ojs.git
+GITREP=${GITREP:-'https://github.com/pkp/ojs.git'}
 
 if [ -z "$1" ]; then
-	echo "Usage: $0 <version> [<tag>-<branch>]";
+	echo "Usage: $0 <version> <tag-or-branch>";
 	exit 1;
 fi
 
 VERSION=$1
 TAG=$2
-PREFIX=ojs
+PREFIX=${PREFIX:-'ojs'}
 BUILD=$PREFIX-$VERSION
 TMPDIR=`mktemp -d $PREFIX.XXXXXX` || exit 1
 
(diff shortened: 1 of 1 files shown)


asmecher:
(see: https://github.com/pkp/ojs/pull/3400#issuecomment-1124274451)

tools/buildpkg.sh
```diff
@@ -108,17 +116,13 @@ git submodule -q update --init --recursive >/dev/null || exit 1
...
-echo "Done"
-
-echo -n " - plugins/paymethod/paypal ... "
-composer.phar --working-dir=plugins/paymethod/paypal install --no-dev
-echo "Done"
-
-echo -n " - plugins/generic/citationStyleLanguage ... "
-composer.phar --working-dir=plugins/generic/citationStyleLanguage install --no-dev
-echo "Done"
+for i in `find . -name composer.json`
```
asmecher, line 119: @ctgraham, there's a tool that'll do this (it's used in Travis testing) in `lib/pkp/tools/travis/install-composer-dependencies.sh`; do you feel like reconciling this to use that script (in which case it should be moved out of `travis`) or alternately adopting the same approach with `find -exec`? I think the reason for the `-maxdepth` is that this command as written will veer off into the dozens of `composer.json` files in `lib/pkp/lib/vendor/`...

ctgraham: In our case here, the `-maxdepth` should be unnecessary for excluding the vendor directory because the vendor directories should only be populated after the `find` has already executed.

I'll happily adopt the same approach with this script, though, as it removes the need for an extra environment variable.

ctgraham: Actually, using the one liner or reconciling the existing `install-composer-dependencies.sh` will alter the script's text output, which I would consider a backwards-incompatible change.

Since I'm currently proposing these PRs against both `main` and 3.3, can you confirm this request?

asmecher: I'm OK with merging this as is -- but it'll also need porting around to OMP and OPS.

PR #3400, opened 2022-05-11
pkp/pkp-lib#7921: Make buildpkg.sh more flexible

ctgraham:
Resolves pkp/pkp-lib#7921 for main.


asmecher:
@ctgraham, no objections here, but could you document the variables that can be taken from the outside environment (e.g. `GITREPO` and `PREFIX`) at the head of the file or in the usage text?

PR #3418, opened 2022-06-06
Stable 3 2 1

groetz:
we are running OJS with version 3.2.1.4. and migrated with intermediate steps from version 2.4.8.1 with the database. Since I do not know well the version 3.2.. My question, why submitting empty mails without content? They are sent with the address from the recipient and from the sender. Where is the error? The empty Mails are not wanted.
Sceenshot: The 2nd entry from top is the empty mail. The 3rd entry is the confirmation of receipt.
Aktivitaetenprotokoll_Wilhelm2
OJS_Forum_Leere_Mail_System2


asmecher:
@groetz, you've submitted a pull request, which is not what you want; see: https://github.com/pkp/pkp-lib/issues/7991#issuecomment-1146616342

PR #3429, opened 2022-06-15
Update locale.po - section.section

DiCaver:
Updated translation for **msgid "section.section"** from plural to singular.


asmecher:
Merged, thanks, @DiCaver! But in case you haven't seen it, we generally accept translation changes through Weblate (https://translate.pkp.sfu.ca).

PR #3445, opened 2022-06-30
Feature/main/8040 update tests

jonasraoni:


classes/core/Application.inc.php
```diff
@@ -105,7 +105,7 @@ public function getNameKey()
      */
     public function getVersionDescriptorUrl()
     {
-        return('http://pkp.sfu.ca/ojs/xml/ojs-version.xml');
+        return 'https://pkp.sfu.ca/ojs/xml/ojs-version.xml';
```
jonasraoni, line 108: Just updated some URLs to use HTTPS.

classes/search/ArticleSearch.inc.php
```diff
@@ -289,7 +289,7 @@ public function formatResults($results, $user = null)
 
                 // Get the issue, storing in cache if necessary.
                 $issueId = $publishedSubmission->getCurrentPublication()->getData('issueId');
-                if (!isset($issueCache[$issueId])) {
+                if ($issueId && !isset($issueCache[$issueId])) {
```
jonasraoni, line 292: The `issueId` might be empty, then the next line fails.

plugins/importexport/pubmed/tests/functional/FunctionalPubmedExportTest.php
```diff
@@ -13,12 +13,15 @@
  * @brief Test PubMed export.
  */
 
-import('lib.pkp.tests.functional.plugins.importexport.FunctionalImportExportBaseTestCase');
+namespace APP\plugins\importexport\pubmed\tests\functional;
 
-class FunctionalPubmedExportTest extends FunctionalImportExportBaseTestCase
+use PKP\tests\PKPTestCase;
```
jonasraoni, line 18: I was going to fix this test (and another very similar located at the `plugins/generic/crossref` submodule), but then I saw a similar test was moved to Cypress, and got in doubt.

asmecher: That's fine -- once the unit test stuff is cleaned up (and this is a big help) we should consider more broadly what do do with PHPUnit and what to do with Cypress.

jonasraoni: I think Cypress is the most important, it assures the software works as it should.
Unit tests are faster to run, better scoped and can test things that might not be accessible (or perhaps too complex to simulate) through the interface, such as the little details.
It would be great to have both 😁

asmecher (commented):
Sorry, I just noticed these comments were pending!

plugins/metadata/dc11/filter/Dc11SchemaArticleAdapter.inc.php
```diff
@@ -209,16 +218,20 @@ public function &extractMetadataFromDataObject(&$article)
         }
 
         // Public identifiers
-        $pubIdPlugins = (array) PluginRegistry::loadCategory('pubIds', true, $journal->getId());
-        foreach ($pubIdPlugins as $pubIdPlugin) {
-            if ($issue && $pubIssueId = $issue->getStoredPubId($pubIdPlugin->getPubIdType())) {
+        $publicIdentifiers = [
```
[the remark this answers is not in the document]
jonasraoni: Yeah, it's the same type of update I did on OPS.
As @bozana is away, I've asked @ewhanson to take a look, I've just updated the code to include the DOI identifier which has been moved out of the `pubIds` plugin's category.

plugins/metadata/dc11/filter/Dc11SchemaArticleAdapter.inc.php
```diff
@@ -209,16 +218,19 @@ public function &extractMetadataFromDataObject(&$article)
...
+            'doi',
```
ewhanson, line 222: This would always include DOIs even if they aren't enabled. If you want to first check if they're enabled before adding them to the array, `$journal->getData(Context::SETTING_ENABLE_DOIS)` would indicate whether they're enabled or not.

jonasraoni: I thought there could be something like that, as it was a plugin before (on/off switch), that's why I invited you :)
Thanks!

PR #3452, opened 2022-07-05
[OJS][main] #8073 Migration Added

defstat:


defstat:
Closing in favor of pkp/ojs#3481

PR #3482, opened 2022-08-09
Bugfix/main/8139 fix dc11 plugin

jonasraoni:


plugins/reports/views/locale/el_GR/locale.po
```diff
@@ -1,8 +0,0 @@
-# Manolis Saldaris <omanos@gmail.com>, 2022.
```
jonasraoni, line 1: Looks like the `plugins/reports/views/` has been removed, just this locale remained.

plugins/oaiMetadataFormats/dc/locale/ar_IQ/locale.po
```diff
@@ -1,18 +0,0 @@
-msgid ""
```
jonasraoni, line 1: These locale files already exist in the `lib/pkp/plugins/oaiMetadataFormats/dc/locale` folder.

PR #3500, opened 2022-08-17
pkp/pkp-lib#7165 Issue datePublished can be set when adding an issue

ajnyga:
https://github.com/pkp/pkp-lib/issues/7165

### DIFF
--- a/classes/publication/Repository.php
+++ b/classes/publication/Repository.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @file classes/publication/Repository.php
  *
@@ -151,24 +152,34 @@ class Repository extends \PKP\publication\Repository
         return $newId;
     }
 
-    /** @copydoc \PKP\publication\Repository::setStatusOnPublish() */
+    /**
+     * Set the publication status and datePublished on publish
+     *
+     * When an issue is published, any attached publications inherit the
+     * issue's `datePublished` in IssueGridHandler if they do not already
+     * have a specific publication date set.
+     *
+     * - If the issue is **not published**, the publication status is set to `STATUS_SCHEDULED`.
+     * - If the issue is **published** or there is no issue, the status is set to `STATUS_PUBLISHED`.
+     * - If the publication does not have a `datePublished`, it is set to the current date.
+     */
     protected function setStatusOnPublish(Publication $publication)
     {
-        // A publication may be scheduled in a future issue. In such cases,
-        // the `datePublished` should remain empty and the status should be set to
-        // scheduled.
-        //
-        // If there is no assigned issue, the journal may be using a continuous
-        // publishing model in which articles are published right away.
         $issue = Repo::issue()->get($publication->getData('issueId'));
+
+        // If issue is not published just set publication status to STATUS_SCHEDULED
         if ($issue && !$issue->getData('published')) {
-            $publication->setData('datePublished', null);
             $publication->setData('status', Submission::STATUS_SCHEDULED);
-        } else {
-            $publication->setData('status', Submission::STATUS_PUBLISHED);
-            if (!$publication->getData('datePublished')) {
-                $publication->setData('datePublished', Core::getCurrentDate());
-            }
+            return;
+        }
+
+        // If issue is published or no issue, set publication status to STATUS_PUBLISHED
+        $publication->setData('status', Submission::STATUS_PUBLISHED);
+
+        // If no predefined datePublished available for the publication, use current date
+        if (!$publication->getData('datePublished')) {
+            error_log('here4');
+            $publication->setData('datePublished', Core::getCurrentDate());
         }
     }
 
--- a/controllers/grid/issues/form/IssueForm.php
+++ b/controllers/grid/issues/form/IssueForm.php
@@ -30,6 +30,7 @@ use PKP\form\Form;
 use PKP\linkAction\LinkAction;
 use PKP\linkAction\request\RemoteActionConfirmationModal;
 use PKP\plugins\Hook;
+use PKP\validation\ValidatorFactory;
 
 class IssueForm extends Form
 {
@@ -60,6 +61,7 @@ class IssueForm extends Form
             return !$showTitle || implode('', $form->getData('title')) != '' ? true : false;
         }));
         $this->addCheck(new \PKP\form\validation\FormValidatorRegExp($this, 'urlPath', 'optional', 'validator.alpha_dash_period', '/^[a-zA-Z0-9]+([\\.\\-_][a-zA-Z0-9]+)*$/'));
+
         $this->addCheck(new \PKP\form\validation\FormValidatorPost($this));
         $this->addCheck(new \PKP\form\validation\FormValidatorCSRF($this));
         $this->issue = $issue;
@@ -141,6 +143,27 @@ class IssueForm extends Form
                     $this->addErrorField('urlPath');
                 }
             }
+        }
+
+        // Note! The following datePublished validation is not triggered
+        // when JQuery DatePicker is enabled in the datePublished field.
+        // The datePicker will convert any invalid date into a valid yyyy-mm-dd
+        // date before the form data is submitted.
+
+        if ($this->getData('datePublished')) {
+            $validator = ValidatorFactory::make(
+                ['value' => $this->getData('datePublished')],
+                ['value' => ['required', 'date_format:Y-m-d']]
+            );
+
+            if (!$validator->passes()) {
+                $this->addError('datePublished', __('editor.issues.datePublished.invalid'));
+                $this->addErrorField('datePublished');
+            }
+        } elseif ($this->issue?->getPublished()) {
+            // A published issue must have a published date
+            $this->addError('datePublished', __('editor.issues.datePublished.requiredWhenPublished'));
+            $this->addErrorField('datePublished');
         }
 
         return parent::validate($callHooks);
@@ -243,9 +266,10 @@ class IssueForm extends Form
         $issue->setVolume(empty($volume) ? null : $volume);
         $issue->setNumber(empty($number) ? null : $number);
         $issue->setYear(empty($year) ? null : $year);
-        if (!$isNewIssue) {
-            $issue->setDatePublished($this->getData('datePublished'));
-        }
+
+        // If issue is not published, allow empty datePublished
+        $issue->setDatePublished(!$this->getData('datePublished') && !$issue->getPublished() ? null : $this->getData('datePublished'));
+
         $issue->setDescription($this->getData('description'), null); // Localized
         $issue->setShowVolume((int) $this->getData('showVolume'));
         $issue->setShowNumber((int) $this->getData('showNumber'));
--- a/locale/en/editor.po
+++ b/locale/en/editor.po
@@ -120,6 +120,15 @@ msgstr "Unpublished"
 msgid "editor.issues.datePublished"
 msgstr "Date Published"
 
+msgid "editor.issues.datePublished.notPublished.description"
+msgstr "If left empty, the date will be set automatically when the issue is published."
+
+msgid "editor.issues.datePublished.invalid"
+msgstr "Date Published is not valid. The date must be in the format YYYY-MM-DD."
+
+msgid "editor.issues.datePublished.requiredWhenPublished"
+msgstr "Date Published is required when the issue is published."
+
 msgid "editor.issues.volumeRequired"
 msgstr "Volume is required and must be a positive, numeric value."
 
--- a/templates/controllers/grid/issues/form/issueForm.tpl
+++ b/templates/controllers/grid/issues/form/issueForm.tpl
@@ -43,16 +43,16 @@
 		{assign var=issuePublished value=false}
 	{/if}
 
-	{if $issuePublished}
 		{fbvFormArea id="datePublishedArea" title="editor.issues.datePublished"}
 			{fbvFormSection}
-				{if $issuePublished}
-					{fbvElement type="text" id="datePublished" value=$datePublished size=$fbvStyles.size.SMALL class="datepicker"}
-				{/if}
+				{if !$issuePublished}
+			        {assign var=notPublishedDescription value="editor.issues.datePublished.notPublished.description"}
+			    {else}
+			        {assign var=notPublishedDescription value=""}
+			    {/if}
+				{fbvElement type="text" id="datePublished" value=$datePublished size=$fbvStyles.size.SMALL class="datepicker" autocomplete="off" label=$notPublishedDescription}
 			{/fbvFormSection}
 		{/fbvFormArea}
-	{/if}
-
 
 	{fbvFormArea id="identificationArea" title="editor.issues.identification"}
 		{fbvFormSection}
--- a/classes/controllers/grid/issues/IssueGridHandler.php
+++ b/classes/controllers/grid/issues/IssueGridHandler.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @defgroup controllers_grid_issues Issues Grid
  * The Issues Grid implements the management interface allowing editors to
@@ -560,7 +561,11 @@ class IssueGridHandler extends GridHandler
         }
 
         $issue->setPublished(1);
-        $issue->setDatePublished(Core::getCurrentDate());
+
+        // If no datePublished was given, use current date
+        if (!$issue->getData('datePublished')) {
+            $issue->setDatePublished(Core::getCurrentDate());
+        }
 
         // If subscriptions with delayed open access are enabled then
         // update open access date according to open access delay policy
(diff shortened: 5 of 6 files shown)


controllers/grid/issues/form/IssueForm.php
```diff
@@ -241,7 +241,7 @@ public function execute(...$functionArgs)
         $issue->setVolume(empty($volume) ? null : $volume);
         $issue->setNumber(empty($number) ? null : $number);
         $issue->setYear(empty($year) ? null : $year);
-        if (!$isNewIssue) {
+        if ($this->getData('datePublished')) {
```
NateWr, line 244: Does this prevent the date published from being removed once it's been added through the form?

templates/controllers/grid/issues/form/issueForm.tpl
```diff
@@ -41,16 +41,15 @@
 		{assign var=issuePublished value=false}
 	{/if}
 
-	{if $issuePublished}
 		{fbvFormArea id="datePublishedArea" title="editor.issues.datePublished"}
 			{fbvFormSection}
 				{if $issuePublished}
 					{fbvElement type="text" id="datePublished" value=$datePublished size=$fbvStyles.size.SMALL class="datepicker"}
+				{else}
+					{fbvElement type="text" id="datePublished" value=$datePublished size=$fbvStyles.size.SMALL class="datepicker" label="editor.issues.datePublished.notPublished.description"}
```
NateWr, line 49: Does this exist already or needs to be added?

controllers/grid/issues/form/IssueForm.php
ajnyga: You are thinking of a scenario where a new issue is added and the date is set, but the issue not published and the user wants to use remove it altogether?

NateWr: Testing steps:

1. Open the form and put `2022-02-03` into the date published field.
2. Click save.
3. Close form and re-open form.
4. Delete `2022-02-03` from the date published field.
5. Click save.
6. Close form and re-open form.

The date published field should be empty.

ajnyga: So this is what I am thinking of adding

 ```
       // If issue is not published, allow empty datePublished
        if (!$this->getData('datePublished') && !$issue->getPublished()){
            $issue->setDatePublished(NULL);
        }

        // If datePublished is set
        if ($this->getData('datePublished')) {
            $issue->setDatePublished($this->getData('datePublished'));
        }
```

Comment (NateWr) [reply controllers/grid/issues/form/IssueForm.php:250]: If it works with the steps above, ok by me. What happens when an issue is published, though? Is there any error message when they try to edit the date published? Or does it just not change silently?

Comment (ajnyga) [reply controllers/grid/issues/form/IssueForm.php:250]: Note that this does not allow you to give an empty value for an issue that is published.

File: templates/controllers/grid/issues/form/issueForm.tpl (hunk 2 above)
Comment (ajnyga) [reply templates/controllers/grid/issues/form/issueForm.tpl:49]: yes, that was missing

File: classes/publication/Repository.php [hunk 3]
```diff
@@ -148,21 +148,33 @@ public function version(Publication $publication): int
...
             }
+            $publication->setData('status', Submission::STATUS_PUBLISHED);
+        } elseif ($issue && $issue->getData('published')) {
+            // If no publication datePublished is given, use the issue datePublished
+            if (!$publication->getData('datePublished')) {
+                $publication->setData('datePublished', $issue->getData('datePublished'));
+            }
+            $publication->setData('status', Submission::STATUS_PUBLISHED);
+        } else {
+            $publication->setData('status', Submission::STATUS_SCHEDULED);
```
NateWr, line 177: This looks good. However, I'd like to restructure the conditional block a little. Right now the `else` block corresponds to a very specific case: a future issue. But it's not that clear from reading the code. Can we structure this to be a little more explicit for reading?

```php
if ($issue) {
    if ($issue->getData('published')) {
        $publication->setData('status', Submission::STATUS_PUBLISHED);
        if (!$publication->getData('datePublished')) {
            $publication->setData('datePublished', $issue->getData('datePublished'));
        }
    } else {
        $publication->setData('status', Submission::STATUS_SCHEDULED);
    }
} else {
    $publication->setData('status', Submission::STATUS_PUBLISHED);
    if (!$publication->getData('datePublished')) {
        $publication->setData('datePublished', Core::getCurrentDate());
    }
}
```

It results in a little more conditional block nesting. But I think it's easier to interpret what is happening in each block.

classes/publication/Repository.php
```diff
@@ -148,21 +148,33 @@ public function version(Publication $publication): int
...
+        // A publication may be published without an issue. Use current date if no
+        // custom `datePublished` is set. Status is set to published.
+
+        // A publication may be published in a published issue.
+        // In such cases the `datePublished` should be what the user has entered
+        // or fall back to the issue published date. Status should be set to published.
+
+        // A publication may also be scheduled in a future issue. In such cases,
+        // the the status should be set to scheduled and `datePublished`
+        // should remain what it is.
```
NateWr, line 161: Can you move this comment to the docblock and make it the following:

```php
/**
 * Set the status and date of publication
 *
 * A publication's status and date of publication will be set depending
 * on the issue it is assigned to. Publications in future issues should
 * be set to `STATUS_SCHEDULED`. Otherwise, they should be set to
 * `STATUS_PUBLISHED` and should inherit the date of publication from the 
 * issue, unless a date of publication has already been set.
 *
 * Usually, the date of publication will be set when the issue is
 * published. In some unusual cases, a publication may be published 
 * without an issue. This may happen in a continuous publishing model 
 * where articles are published right away. In these cases, the date of 
 * publication should be set to the current date, unless a date of 
 * publication has already been set.
 */
```

controllers/grid/issues/form/IssueForm.php
```diff
@@ -241,9 +241,16 @@ public function execute(...$functionArgs)
         $issue->setVolume(empty($volume) ? null : $volume);
         $issue->setNumber(empty($number) ? null : $number);
         $issue->setYear(empty($year) ? null : $year);
-        if (!$isNewIssue) {
+
+        // If issue is not published, allow empty datePublished
+        if (!$this->getData('datePublished') && !$issue->getPublished()) {
+            $issue->setDatePublished(null);
+        }
```
NateWr, line 248: Do you need to add a check to the `validate()` method when `!$this->getData('datePublished') && $issue->getPublished()`?

NateWr (changes requested):

classes/publication/Repository.php
ajnyga: Looks good! Just want to note that again the last `else` is something that the current code base will never trigger. You will not get here without having an `issue` set. But it is definitely something we have been working towards. Just thinking that do we want to have this "sleeping code" here already now?

controllers/grid/issues/form/IssueForm.php
ajnyga: You mean if no `datePublished` is given, but `issue` is published, we should give a warning. I think at the moment you can't just remove the date and no warning is shown.

Maybe we need a check for a valid date as well? We did not have one formerly, but would probably make sense. I was just looking around and could not find if we have one somewhere already, I mean if there is an existing solution for validating date stamps.

classes/publication/Repository.php
NateWr: That's a fair point, but if we left the `else` block out, we'd have an "invisible" condition. Calling `setStatusOnPublish()` for such a publication would do nothing. I'd prefer to fall back to a sensible default behavior that does what the function says it does (changing the status when something is published).

controllers/grid/issues/form/IssueForm.php
NateWr: Yeah, so if someone removes the `datePublished` and clicks save, and the `datePublished` isn't removed, they need to see some message indicating that. Otherwise they will assume it's changed and then we have bad metadata.

I don't think we have one of the old form validators, but if you look at `ValidatorRegExp` you can see an example of using [Laravel's Validation Rules](https://laravel.com/docs/9.x/validation#available-validation-rules). I think we'd want [date_format:format](https://laravel.com/docs/9.x/validation#rule-date-format), something like this:

 ```php
$validator = ValidatorFactory::make(
    ['value' => $value],
    ['value' => ['required', 'date_format:Y-m-d']]
);

$result = $validator->passes();
```

You can also skip the `Validator` class and just add this to `IssueForm::validate()`.

File: classes/publication/Repository.php (hunk 3 above)
Comment (ajnyga) [reply classes/publication/Repository.php:177]: sure agree!

File: controllers/grid/issues/form/IssueForm.php (hunk 5 above)
Comment (ajnyga) [reply controllers/grid/issues/form/IssueForm.php:248]: The `datePublished` field is acting very weird. I think the `datepicker` functionality is affecting it.

I added `error_log(print_r($this->getData('datePublished'), true));` to the form `validation` method to see what I get there when I save the form.

```
Form input: (empty) -> Error output (empty) (this value is saved to the issue!)
Form input: 55 -> Error output (empty) (issue value still empty)
Form input: 2019-01 -> Error output (empty)  (issue value still empty)
Form input: 2019-01-01 -> Error output 2019-01-01 (this value is saved to the issue!)
Form input: 55 -> Error output 2019-01-01   (issue value still  2019-01-01)
Form input: 2019-01 -> Error output 2019-01-01 (issue value still  2019-01-01)
Form input: (empty) -> Error output (empty) (this value is saved to the issue!)
```

So it looks like there is some internal validation going on in the datepicker?

Comment (NateWr) [reply controllers/grid/issues/form/IssueForm.php:248]: There may be client-side validation, but it can't replace server-side validation.

Comment (ajnyga) [reply controllers/grid/issues/form/IssueForm.php:248]: yes, looking at the source with developer tools I can see that the `value` attribute in `input` is only updated when you enter a valid date.

Comment (ajnyga) [reply controllers/grid/issues/form/IssueForm.php:248]: so basically you can not get an invalid date from the form when you get to validation. The problem is of course that you won't see any error. It will just use the last valid value it had.

Comment (NateWr) [reply controllers/grid/issues/form/IssueForm.php:248]: You can't rely on the UI to do validation because the user can send any data in the HTTP request.

Comment (ajnyga) [reply controllers/grid/issues/form/IssueForm.php:255]: sure, but you can not get an invalid date at the moment to the validation method.

Comment (ajnyga) [reply controllers/grid/issues/form/IssueForm.php:255]: I had something like

```
        if ($this->getData('datePublished')) {

            $validator = ValidatorFactory::make(
               ['value' => $this->getData('datePublished')],
               ['value' => ['required', 'date_format:Y-m-d']]
            );

            if (!$validator->passes()){
                $this->addError('datePublished', __('editor.issues.datePublished.invalid'));
                $this->addErrorField('datePublished');
            }
        }
```

in the validation, but the error is never triggered because the date is always valid.

**Comment by ajnyga** [comment]
One test failed, probably would be ok if restarted.

**Comment by ajnyga** [comment]
@NateWr for some reason I remembered that this got merged already a while ago. The tests went through back in December and all the changes should be there. I rebased and started the tests again.

File: templates/controllers/grid/issues/form/issueForm.tpl [hunk 6]
```diff
@@ -43,16 +43,15 @@
...
```
defstat, line 51: Here we are adding the same `fbvElement` with the addition of the `label`. Would it be better to use the `if` clause to define a `label` string to `""` or `"editor.issues.datePublished.notPublished.description"` and add the `fbvElement` once?

Or to refrain the wording of the label so we can add that changed label to both cases?

defstat (commented):
@ajnyga Just one or two questions/recommendations..

templates/controllers/grid/issues/form/issueForm.tpl
defstat, line 51: On my browser I am getting something like:

image

Not really important but perhaps by adding a simple `autocomplete="off"` there will do the trick.

defstat, line 51: Here we are adding the same `fbvElement` with the addition of the `label`. Would it be better to use the `if` clause to define a `label` string to `""` or `"editor.issues.datePublished.notPublished.description"` and add the `fbvElement` once?

Or to refrain the wording of the label so we can add that changed label to both cases?

ajnyga: check

defstat: Because the `$issuePublished` flag is used only here, I was thinking something like 

```
{if $issue && $issue->getPublished()}
	{assign var=label value=""}
{else}
	{assign var=label value="editor.issues.datePublished.notPublished.description"}
{/if}

{fbvFormArea id="datePublishedArea" title="editor.issues.datePublished"}
	{fbvFormSection}
		{fbvElement type="text" id="datePublished" value=$datePublished size=$fbvStyles.size.SMALL class="datepicker" label=$label}
	{/fbvFormSection}
{/fbvFormArea}
```

It's not a big deal, just saving the replication of the `fbvElement` there.

ajnyga:
ah ok I see, I had upgraded to PHP 8.3 and it seems that PHP-CS-Fixer does not work with that version and that leads to empty commits, just did not read what the console was telling me :D

ajnyga:
Some of the tests are failing, but it seems that in a point where my additions should not have an effect?

classes/publication/Repository.php
```diff
@@ -151,19 +151,35 @@ public function version(Publication $publication): int
...
+
+        if ($issue) {
+            if ($issue->getData('published')) {
+                $publication->setData('status', Submission::STATUS_PUBLISHED);
+                if (!$publication->getData('datePublished')) {
+                    $publication->setData('datePublished', $issue->getData('datePublished'));
+                }
+            } else {
+                $publication->setData('status', Submission::STATUS_SCHEDULED);
+            }
```
jonasraoni, line 182: Not sure if it looks simpler to you, but then it would be enough to drop also the `else` that comes later on:

```suggestion
        if ($issue && !$issue->getData('published')) {
            $publication->setData('status', Submission::STATUS_SCHEDULED);
            return;
        }

        $publication->setData('status', Submission::STATUS_PUBLISHED);
        $publication->setData('datePublished', $issue?->getData('published') ? $issue->getData('datePublished') : Core::getCurrentDate());

```

controllers/grid/issues/form/IssueForm.php
```diff
@@ -143,6 +145,29 @@ public function validate($callHooks = true)
...
+                $this->addError('datePublished', __('editor.issues.datePublished.invalid'));
+                $this->addErrorField('datePublished');
+            }
+        }
+
+        // Published issue has to have datePublished set
+        if (!$this->getData('datePublished') && ($this->issue && $this->issue->getPublished())) {
+            $this->addError('datePublished', __('editor.issues.datePublished.requiredWhenPublished'));
+            $this->addErrorField('datePublished');
+        }
```
jonasraoni, line 169:
```suggestion
        } else if ($this->issue?->getPublished()) {
          // A published issue must have a published date
          $this->addError('datePublished', __('editor.issues.datePublished.requiredWhenPublished'));
          $this->addErrorField('datePublished');
        }
```

controllers/grid/issues/form/IssueForm.php
```diff
@@ -243,9 +268,16 @@ public function execute(...$functionArgs)
...
+
+        // If issue is not published, allow empty datePublished
+        if (!$this->getData('datePublished') && !$issue->getPublished()) {
+            $issue->setDatePublished(null);
+        }
+
+        if ($this->getData('datePublished')) {
             $issue->setDatePublished($this->getData('datePublished'));
         }
+
```
jonasraoni, line 280:
```suggestion
        $issue->setDatePublished(!$this->getData('datePublished') && !$issue->getPublished() ? null : $this->getData('datePublished'));
```

locale/en/editor.po
```diff
@@ -120,6 +120,15 @@ msgstr "Unpublished"
 msgid "editor.issues.datePublished"
 msgstr "Date Published"
 
+msgid "editor.issues.datePublished.notPublished.description"
+msgstr "Leave this empty and it will be set automatically when the issue is published."
```
jonasraoni, line 124:
```suggestion
msgstr "If left empty, the date will be set automatically when the issue is published."
```

jonasraoni (approved):

classes/publication/Repository.php
ajnyga: But this would overwrite the `datePublished` value even if there is one already?

ajnyga: So maybe use 

```
if ($issue && !$issue->getData('published')) {
    $publication->setData('status', Submission::STATUS_SCHEDULED);
    return;
}

$publication->setData('status', Submission::STATUS_PUBLISHED);

if (!$publication->getData('datePublished')) {
    $publication->setData('datePublished', $issue ? $issue->getData('datePublished') : Core::getCurrentDate());
}

```

controllers/grid/issues/form/IssueForm.php
ajnyga: check

locale/en/editor.po
ajnyga: check

PR #3524, opened 2022-08-31
[OJS][stable-3_3_0] #7249 ***TESTS ONLY***

defstat:


asmecher:
Since this change requires a migration, it shouldn't be included in `stable-3_3_0`.

PR #3529, opened 2022-09-01
 pkp/pkp-lib#7330 consider version in the HTML image download URL 

bozana:
s. https://github.com/pkp/pkp-lib/issues/7330


ctgraham:
This resolved both the issue of previewing the galley, and of viewing past versions for me in my example.

PR #3545, opened 2022-09-15
pkp/pkp-lib#6545 Clean up misnamed fileId variable

asmecher:


pages/article/ArticleHandler.php
```diff
@@ -396,9 +397,9 @@ public function viewFile($args, $request)
     {
```
jonasraoni, line 397: You've removed this method in the PR for OPS, but here it was kept, I guess OPS inherited the code from OJS (and didn't even make use of this old URL), right?

I think it's needed to also update [the `pages/article/index.php`](https://github.com/pkp/ojs/blob/main/pages/article/index.php#L20) here and for the other PR (which also has a reference to the `downloadSuppFile`).

p.s.: These hard-coded strings at the `pages/index.php` looks like a good candidate for a refactor, but I guess it's already on the radar =]

jonasraoni (approved):
LGTM. I remember we had issues with the `fileId` being mixed with the `submissionFileId` at the Native Import Export plugin as well, so such fixes are helpful :)

PR #3572, opened 2022-10-12
[OJS][stable-3_3_0] #8318 Search SQL should consider only contexts that don't have PUBLISHING_MODE_NONE

defstat:


classes/search/ArticleSearchDAO.inc.php
```diff
@@ -80,8 +80,12 @@ public function getPhraseResults($journal, $phrase, $publishedFrom = null, $publ
 				JOIN publication_settings ps ON (ps.publication_id = p.publication_id AND ps.setting_name=\'issueId\' AND ps.locale=\'\')
 				JOIN issues i ON (CAST(i.issue_id AS CHAR(20)) = ps.setting_value AND i.journal_id = s.context_id)
 				JOIN submission_search_objects o ON (s.submission_id = o.submission_id)
+				JOIN journals con ON con.journal_id = s.context_id
+				LEFT OUTER JOIN journal_settings con_s_publishingMode on con.journal_id = con_s_publishingMode.journal_id AND con_s_publishingMode.setting_name = \'publishingMode\'
```
bozana, line 84: This is only a very minor comment, but:
I find easier readable short table aliases -- I also think we use them generally where possible -- thus I would prefer to have `journals j` and `journal_settings js`. For OPS main branch: for servers maybe `servers j` or `servers c`
Also, we mostly use only LEFT JOIN (without OUTER) and write `ON` (instead of `on`).

classes/search/ArticleSearchDAO.inc.php
```diff
@@ -80,8 +80,12 @@ public function getPhraseResults($journal, $phrase, $publishedFrom = null, $publ
...
 				NATURAL JOIN ' . $sqlFrom . '
 			WHERE
+				(con_s_publishingMode.setting_value <> \'' . PUBLISHING_MODE_NONE . '\' OR
```
bozana, line 87: Do we need `'` around the constant PUBLISHING_MODE_NONE ?

classes/search/ArticleSearchDAO.inc.php
```diff
@@ -80,8 +80,12 @@ public function getPhraseResults($journal, $phrase, $publishedFrom = null, $publ
...
+				con_s_publishingMode.setting_value is null) AND
```
bozana, line 88: Also maybe `IS NULL` (instead of `is null`) :-)

bozana (commented):
@defstat, just a few very minor comments -- if you agree :-)
Sorry for being picky :-P

classes/search/ArticleSearchDAO.inc.php
defstat: 1. Changed the aliases to `j` and `js` for all changes.
2. Removed `OUTER` from `LEFT OUTER JOIN`
3. Changed `on` to `ON`

classes/search/ArticleSearchDAO.inc.php
defstat: I am using `'` because the `js.setting_value` is a `varchar` column and the value to be compared should be varchar one, thus the `'` enclosure of the `PUBLISHING_MODE_NONE`.

PR #3575, opened 2022-10-13
pkp/pkp-lib#6093 Add indexes to FK relationships

asmecher:


classes/migration/install/MetricsMigration.php
```diff
@@ -25,130 +25,187 @@ public function up(): void
...
             $table->date('date');
             $table->integer('metric');
+
             $table->foreign('context_id')->references('journal_id')->on('journals')->onDelete('cascade');
-            $table->index(['load_id'], 'metrics_context_load_id');
             $table->index(['context_id'], 'metrics_context_context_id');
         });
         Schema::create('metrics_submission', function (Blueprint $table) {
             $table->string('load_id', 255);
+            $table->index(['load_id'], 'ms_load_id');
```
jonasraoni, line 39: As these entities are new, perhaps it makes sense to update the names to follow the previous pattern (e.g. `ms_load_id` > `metrics_submission_load_id`). If yes, then there are other cases below.

asmecher: @bozana, that line is actually just moved from below (to group the column creation, constraint, and index lines together) -- I think this is actually your LOC :)

bozana: Hihihihih... :-D OK :+1:

PR #3593, opened 2022-10-29
pkp/pkp-lib#7318 Stats Context Page

bozana:
s. https://github.com/pkp/pkp-lib/issues/7318


templates/stats/issues.tpl
```diff
@@ -172,6 +172,19 @@
 										</pkp-button>
 									</template>
 								</action-panel>
+								<action-panel class="pkpStats__reportAction">
```
ewhanson, line 175: I also wasn't seeing this action panel show up in the UI anywhere.

bozana: Screenshot from 2022-11-04 14-40-08

bozana: Hi @ewhanson, this should be the `Timeline` part in the screenshot.
This modal opens when you click on the Download button on the Statistics > Issue i.e. Journal/Press/Server page.
Do you see it?

ewhanson: Yes, I see it now. Somehow missed the modal. All good!

PR #3658, opened 2022-12-13
pkp/pkp-lib#8490 consider mariaDB in Travis

bozana:
@asmecher, I am not sure at all what everything is needed to have the tests for mariaDB too.
Is this here OK and should I add something else too?


asmecher:
I think you'll also have to adjust `lib/pkp/tools/travis/prepare-tests.sh` so that it knows what a `TEST` of `mariadb` means.

PR #3698, opened 2023-01-10
#6340 Set secure flag to true when force_ssl is on

salmanm2003:
@asmecher I have run an automated test for OJS regarding #6340 issue


asmecher:
Manually committed submodule update.

PR #3701, opened 2023-01-14
Bugfix/main/8544 fix session issues

jonasraoni:


config.TEMPLATE.inc.php
```diff
@@ -45,6 +45,11 @@
 ; (set to 0 to force expiration at end of current session)
 session_lifetime = 30
 
+; SameSite configuration for the cookie, see possible values and explanations
+; at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
+; To set the "Secure" attribute for the cookie see the setting force_ssl at the [security] group
+session_samesite = lax
```
asmecher, line 51: This should be updated to `Lax` as a default.

PR #3702, opened 2023-01-15
pkp/pkp-lib#5726 Use lists for authors and author biographies

israelcefrin:
Adjust the list of authors bellow the abstract to show as undefined list and helping screen readers users to know how many authors are listed.

Related to issue [#5726 ](https://github.com/pkp/pkp-lib/issues/5726)


templates/frontend/objects/article_details.tpl
```diff
@@ -212,24 +212,26 @@
 							{translate key="submission.authorBiography"}
 						{/if}
 					</h2>
+					<ul class="authors">
```
NateWr, line 215: Have you checked that this doesn't change the visual display of the author biographies? I think it will change the display to a visual bullet-point list, but we don't want there to be any change to the way biographies appear.

If you take a before/after screenshot that will help evaluate the change.

templates/frontend/objects/article_details.tpl
```diff
@@ -212,24 +212,26 @@
...
 					{foreach from=$publication->getData('authors') item=author}
 						{if $author->getLocalizedData('biography')}
-							<section class="sub_item">
-								<h3 class="label">
+							<li class="sub_item">
+								<span class="label">
```
NateWr, line 219: Are you sure we don't want a `<h3>` here? I think even within a list it is good to keep heading hierarchy where it makes sense.

templates/frontend/objects/article_details.tpl
```diff
@@ -212,24 +212,26 @@
...
 										{$author->getFullName()|escape}
 									{/if}
-								</h3>
-								<div class="value">
+								</span>
+								<span class="value">
 									{$author->getLocalizedData('biography')|strip_unsafe_html}
-								</div>
-							</section>
+								</span>
```
NateWr, line 230: Why change this to `<span>`?

templates/frontend/objects/article_details.tpl
israelcefrin: I've used the same elements from the authors list on the top. The ul.authors class removes the bullets. See the before:
image

And the after:
image

templates/frontend/objects/article_details.tpl
israelcefrin: I was applying the same approach of listing authors we have on the top. We don't use headings on the list itself. Should we consider the author's name a heading section of the document? I can see it could be interchangeable with `dl > dt/dd` elements which also conflicts with the Heading approach. 
See the top:
image

And the bottom:
image

templates/frontend/objects/article_details.tpl
israelcefrin: Good catch!! I probably brought the element from the list on the top. However, we need to keep it using `DIV` because it is a fallback for bios without the `P` tag and we force a block element to separate from the previous element. I will adjust it.

templates/frontend/objects/article_details.tpl
israelcefrin: I was reviewing our other issues related to this topic, and this one is related to the heading removal from listing author biographies: https://github.com/pkp/pkp-lib/issues/6901

NateWr: Ok, if the auditor suggested a list with no headings then let's go with that.

templates/frontend/objects/article_details.tpl
NateWr: In your after screenshot, the style of the affiliation data has changed (grey instead of black). But I can't see any change in this PR that would cause that. What am I missing?

israelcefrin: The affiliation color has changed because ` ul.authors > li > span.affiliation` has a grey color set on the CSS (see the same color on the top list). I can adjust it on CSS adding a rule for `.author_bios .affiliation` to keep it black.

templates/frontend/objects/article_details.tpl
israelcefrin: Perfect, I will use `DIV` as suggested by the auditor to keep it in block elements.

NateWr:
It looks like you may have accidentally brought in a number of extra commits to this PR.

templates/frontend/objects/article_details.tpl
NateWr: I'd say maybe remove the `<span class="affiliation">` tag from the authors in this section. The name and affiliation can be just one HTML element. That way we don't need to change any existing CSS, and the styles stay the same.

PR #3740, opened 2023-02-02
pkp/pkp-lib#8091 Add payment request notification email

Vitaliy-1:


NateWr (commented):
Looks good, just a couple changes to the email itself.

PR #3758, opened 2023-02-12
pkp/orcidProfile#232  Display Orcid Icon only for authenticated users

withanage:
- https://github.com/pkp/orcidProfile/issues/232


withanage:
Hi @asmecher ,
This  PR adresses https://github.com/pkp/orcidProfile/issues/232

PR #3761, opened 2023-02-14
[OJS][main] 8598 DB Migration changes

defstat:


asmecher:
@defstat, I think this is also outdated; can you confirm?

PR #3776, opened 2023-02-17
Fix test regression from pkp/pkp-lib#8310

ewhanson:


ewhanson:
Hey @asmecher, here's the OJS fix for the tests.

PR #3784, opened 2023-02-20
pkp/pkp-lib#7128 fix customSectionOrdering

bozana:
s. https://github.com/pkp/pkp-lib/issues/7128


bozana:
Hi @asmecher, could you please take a look (and eventually overtake, if it should be differently solved)? Thanks a lot!

asmecher:
Merged, thanks for the quick turn-around!

PR #3804, opened 2023-03-02
Feature/stable 3 3 0/8710 address recommend by similarity performance

jonasraoni:


plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.inc.php
```diff
@@ -56,45 +62,87 @@ function getDescription() {
...
+
+		/** @var Builder */
+		$queryBuilder = (new SubmissionQueryBuilder())
+			->filterByContext($journal->getId())
+			->filterByStatus(STATUS_PUBLISHED)
+			->getQuery();
+
+		$queryBuilder->where('s.submission_id', '<>', $articleId);
+
+		$keywordIdFields = [];
```
jonasraoni, line 98: I'll add comments/prettify it a bit, but before doing that, I wanted to know if the approach is ok.

I'm basically replacing the `ArticleSearch::retrieveResults()`, and while doing that, I've tried to keep the "ranking".
1. Number of keywords matched (A record with 1 `keywordA` + 1 `keywordB" has higher priority than a record with 100 `keywordA`)
2. Total number of keyword matches

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.inc.php
```diff
@@ -56,45 +62,87 @@ function getDescription() {
...
+				$query->from('submission_search_objects', 'o')
+					->join('submission_search_object_keywords AS ok', 'o.object_id', '=', 'ok.object_id')
+					->whereIn('ok.keyword_id', array_map('\\Illuminate\\Database\\Capsule\\Manager::raw', $keywordIdFields))
+					->whereColumn('s.submission_id', '=', 'o.submission_id');
+			});
+
+		/** @var SubmissionDAO */
+		$submissionDao = DAORegistry::getDAO('SubmissionDAO');
+		$result = $submissionDao->retrieveRange($sql = $queryBuilder->toSql(), $params = $queryBuilder->getBindings(), $rangeInfo);
+		$resultSet = new DAOResultFactory($result, $submissionDao, '_fromRow', [], $sql, $params, $rangeInfo);
```
jonasraoni, line 129: I've tried to use the `Service('submission')->getMany()`, but:
1. The returned iterator doesn't work with the Smarty tag (`{page_links}`).
2. Requires adding a hot (should be dropped after executed) hook to overwrite the query builder.

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.inc.php
```diff
@@ -56,45 +62,87 @@ function getDescription() {
...
+			'articlesBySimilarityQuery' => implode(' ', $keywords),
+			'journal' => $journal,
+			'plugin' => $this
+		]);
 		$output .= $smarty->fetch($this->getTemplateResource('articleFooter.tpl'));
 		return false;
 	}
-}
 
+	public function getIssue(int $issueId): ?Issue
```
jonasraoni, line 141: I'll probably use the `remember()` at the submission list, to retrieve all the related issues on a single query 🤔

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.inc.php
jonasraoni: In fact, I'd like to drop the ranking for extra performance, perhaps I can add it as a plugin option...

Then as a last improvement, which I'm not going to cover here, would be to add some <script> magic to avoid a full page reload.

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.inc.php
```diff
@@ -56,45 +62,87 @@ function getDescription() {
...
+			$orderByMatches[] = "CASE WHEN {$orderBy} = 0 THEN 1 ELSE 0 END";
+		}
+		// Clear any previous ORDER BY
+		$queryBuilder->orders = [];
+		$queryBuilder->orderBy(Capsule::raw(implode(' + ', $orderByMatches)))
+			->orderByDesc(Capsule::raw(implode(' + ', $orderByMatchCount)))
+			->whereExists(function (Builder $query) use ($keywordIdFields) {
+				$query->from('submission_search_objects', 'o')
+					->join('submission_search_object_keywords AS ok', 'o.object_id', '=', 'ok.object_id')
+					->whereIn('ok.keyword_id', array_map('\\Illuminate\\Database\\Capsule\\Manager::raw', $keywordIdFields))
```
asmecher, line 122: It'll be nice to have access to [first-class callable syntax](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) when we set our baseline at PHP8.1.

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.inc.php
```diff
@@ -56,45 +62,87 @@ function getDescription() {
...
-					null, null,
-					$rangeInfo,
-					array($articleId)
-			),
-			'articlesBySimilarityQuery' => $query,
-		));
+		$rangeInfo->setCount(static::RECOMMEND_BY_SIMILARITY_PLUGIN_COUNT);
+
+		/** @var Builder */
+		$queryBuilder = (new SubmissionQueryBuilder())
```
asmecher, line 91: This probably contains some necessary evils for 3.3.0, but for future releases, it would be more maintainable if it didn't mix idioms so much (using the SubmissionQueryBuilder, extending it locally using the query builder, and extending it locally with string-based construction). Sometimes it's worse to try to re-use code when the impedance match isn't perfect, than it is  to just fully code the query in the plugin where it can be understood without having to paw through the interactions.

If the additions to `main` to get `SubmissionQueryBuilder` are non-invasive (i.e. just adding a few functions/clauses) and possibly useful elsewhere, I'd be open to a PR there as well. Another approach might be to use `SubmissionQueryBuilder` as is, fetch or chunk its results, and perform the next operations separately (if that's not prohibitive to performance, which might be a problem here).

Long story short, I'm OK to go ahead with this, particular for `stable-3_3_0` where we're unlikely to refactor things -- but this would be a real tangle for someone to have to sort out if they changed an underlying implementation that broke this.

asmecher (approved):
Thanks, @jonasraoni, and sorry for the wait! As the notes say, I'm OK to go ahead with this, but please review the comments.

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.inc.php
jonasraoni: Okay! This was just an initial validation, as the performance is ok and you accepted to skip the `ArticleSearch::retrieveResults()`, I'll try to improve the code 👌

PR #3853, opened 2023-03-25
Bugfix/main/8845 general fixes

jonasraoni:


classes/subscription/InstitutionalSubscriptionDAO.php
```diff
@@ -563,7 +563,7 @@ public function getByDateEnd($dateEnd, $journalId, $rangeInfo = null)
     {
         $dateEnd = explode('-', $dateEnd);
 
-        $params = array_merge([$dateEnd[0], $dateEnd[1], dateEnd[2], (int) $journalId], $this->getInstitutionNameFetchParameters());
+        $params = array_merge([$dateEnd[0], $dateEnd[1], $dateEnd[2], (int) $journalId], $this->getInstitutionNameFetchParameters());
```
bozana, line 566: :+1:

classes/services/OJSServiceProvider.php
```diff
@@ -37,11 +37,6 @@ public function register(Container $pimple)
             return new PKPFileService();
         };
 
-        // Section service
-        $pimple['section'] = function () {
-            return new SectionService();
-        };
-
```
bozana, line 44: :+1:

plugins/generic/htmlArticleGalley/HtmlArticleGalleyPlugin.php
```diff
@@ -87,6 +88,8 @@ public function articleViewCallback($hookName, $args)
         }
 
         $submissionFile = $galley->getFile();
+        /** @var ?Publication */
+        $galleyPublication = null;
```
bozana, line 92: :+1:

controllers/grid/issues/form/IssueGalleyForm.php
```diff
@@ -218,6 +218,7 @@ public function execute(...$functionArgs)
 
             if ($this->getData('label') == null) {
                 // Generate initial label based on file type
+                $fileType = $issueGalleyFile->getContentType();
```
jonasraoni, line 221: @asmecher The `$fileType` is undefined at this point, so I guessed it should be set this way.

classes/plugins/PubObjectCache.php
```diff
@@ -43,13 +43,13 @@ public function add($object, $parent)
...
             $this->_insertInternally($object, 'articles', $object->getId());
-            $this->_insertInternally($object, 'articlesByIssue', $object->getcurrentPublication()->getData('issueId'), $object->getId());
+            $this->_insertInternally($object, 'articlesByIssue', $object->getCurrentPublication()->getData('issueId'), $object->getId());
         }
         if ($object instanceof Galley) {
             assert($parent instanceof Submission);
             $this->_insertInternally($object, 'galleys', $object->getId());
             $this->_insertInternally($object, 'galleysByArticle', $object->getData('submissionId'), $object->getId());
-            $this->_insertInternally($object, 'galleysByIssue', $parent->getIssueId(), $object->getId());
+            $this->_insertInternally($object, 'galleysByIssue', $parent->getCurrentPublication()->getData('issueId'), $object->getId());
```
jonasraoni, line 52: @asmecher Here I just trusted the `assert()` above, I didn't investigate how this code is being used.

classes/install/Upgrade.php
```diff
@@ -212,7 +213,74 @@ public function migrateSRLocale()
 
         $oldLocaleStringLength = 's:5';
 
-        $journalSettingsDao = new \PKP\db\DAO();
+        $journalSettingsDao = new class () extends \PKP\db\DAO {
```
jonasraoni, line 216: @asmecher There are several calls to `updateSetting()`/`getSetting()` using the `$journalSettingsDao` below, but the `\PKP\db\DAO` doesn't have such methods.

So I just copied from somewhere and adapted here.

controllers/grid/issues/form/IssueGalleyForm.php
asmecher: It looks fine at a glance; is it tested to work when you create an issue galley?

classes/plugins/PubObjectCache.php
asmecher: That looks OK, @jonasraoni, but interesting that it made it through RC1 testing without a fatal error on missing `getIssueId` function being raised. @ewhanson, is the `PubObjectCache` being used in the CrossRef code still?

ewhanson: It's not used in the Crossref code, but it looks like the cache is still used in the Datacite plugin, but only as it relates to galleys.

https://github.com/pkp/ojs/blob/e3b451eae41d077412ab72da0b90271242b6bdd4/plugins/generic/datacite/DataciteExportPlugin.php#L477-L482

classes/install/Upgrade.php
asmecher: This will only affect users who are upgrading from 3.1.x and are using the `sr_SR` locale, which will be a very small demographic, but please go ahead and add this to `main`.

classes/plugins/PubObjectCache.php
asmecher: Galleys are the case Jonas found the issue with above. That explains why testing didn't turn it up, thanks! Jonas, please go ahead with the change you've proposed.

asmecher: (^ @jonasraoni)

controllers/grid/issues/form/IssueGalleyForm.php
jonasraoni: I watched around and removed this whole piece (the label field is required).

PR #3865, opened 2023-04-04
pkp/orcidProfile#232 Display Orcid Icon only for authenticated users

withanage:
### Issue
- https://github.com/pkp/orcidProfile/issues/232

### DIFF
--- a/templates/frontend/objects/article_details.tpl
+++ b/templates/frontend/objects/article_details.tpl
@@ -120,6 +120,9 @@
 							{if $author->getData('orcid')}
 								<span class="orcid">
 									{$orcidIcon}
+                                    {if $author->getData('orcidAccessToken')}
+                                        {$orcidIcon}
+                                    {/if}
 									<a href="{$author->getData('orcid')|escape}" target="_blank">
 										{$author->getData('orcid')|escape}
 									</a>


withanage:
Hi @asmecher ,
This PR  backports  https://github.com/pkp/orcidProfile/issues/232, which was included in [main branch](https://github.com/pkp/ojs/pull/3758)

Would be very useful to have this shipped in 3.3.0-15

asmecher:
Merged, thanks!

templates/frontend/objects/article_details.tpl
```diff
@@ -120,6 +120,9 @@
 							{if $author->getData('orcid')}
 								<span class="orcid">
 									{$orcidIcon}
```
mhvezina, line 122: Shouldn't this line be removed? If not the icon will likely appear twice...

withanage: @mhvezina  thanks for  your review. 

Actullay, no.    AcessToken is only checked for displaying the orcidLogo. Here is the current  stable-3_3_0  branch.

https://github.com/pkp/ojs/blob/stable-3_3_0/templates/frontend/objects/article_details.tpl#L120

mhvezina: I meant this line : https://github.com/pkp/ojs/pull/3865/files#diff-959b67b46d8905ed019270b42e656fe1286d6747716787b4681b5de6710c1eaeL122  But indeed now I see the line has been removed in https://github.com/pkp/ojs/blob/stable-3_3_0/templates/frontend/objects/article_details.tpl#LL121C23-L121C23 Thanks.

PR #3872, opened 2023-04-11
Bugfix/main/8905 fix event discovery

jonasraoni:


classes/issue/Collector.php
```diff
@@ -348,8 +348,8 @@ public function getQueryBuilder(): Builder
         // Context
         // Never permit a query without a context_id unless the PKPApplication::CONTEXT_ID_ALL wildcard
         // has been explicitly set
-        if (!in_array(PKPApplication::CONTEXT_ID_ALL, $this->contextIds)) {
-            $q->whereIn('i.journal_id', $this->contextIds);
+        if (!in_array(PKPApplication::CONTEXT_ID_ALL, $this->contextIds ?? [])) {
```
jonasraoni, line 351: At this point `$this->contextIds` might be `null`.

jonasraoni: I'll probably update to follow the existing pattern:
https://github.com/pkp/pkp-lib/blob/53736488b005a72a3485ed012da389a07cda54fb/classes/submission/Collector.php#L267-L271

Vitaliy-1: Doesn't seem related to the events but linked example looks like a good way to proceed 👍

PR #3874, opened 2023-04-12
Feature/main/8710 address recommend by similarity performance

jonasraoni:


plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.php
```diff
@@ -8,104 +8,130 @@
...
+        if (!strlen($searchPhrase = implode(' ', (new ArticleSearch())->getSimilarityTerms($submissionId)))) {
+            return null;
+        }
+
+        $request = Application::get()->getRequest();
+        $router = $request->getRouter();
+        $context = $router->getContext($request);
+
+        $rangeInfo = Handler::getRangeInfo($request, 'articlesBySimilarity');
+        $rangeInfo->setCount(static::DEFAULT_RECOMMENDATION_COUNT);
```
jonasraoni, line 72: The `$rangeInfo` has access to the page and the count, but it doesn't compute the offset, is this expected or a bug?

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.php
```diff
@@ -8,104 +8,130 @@
...
+
+        $templateManager->assign('articlesBySimilarity', (object) [
+            'submissions' => $submissions,
+            'query' => $searchPhrase,
+            'issues' => $issues,
+            'start' => $offset + 1,
+            'end' => $offset + $submissions->count(),
+            'total' => $submissionCount,
+            'nextUrl' => $nextPage,
+            'previousUrl' => $previousPage
```
jonasraoni, line 116: I'm passing all these variables to fill the `frontend/components/pagination.tpl`, perhaps this can be transformed into a Smarty tag 🤔

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.php
Vitaliy-1: Indeed

Vitaliy-1 (approved):

PR #3875, opened 2023-04-13
Feature/main/8915 improve search indexer

jonasraoni:


classes/search/ArticleSearchIndex.php
```diff
@@ -109,24 +113,28 @@ public function deleteTextIndex($articleId, $type = null, $assocId = null)
...
+            return;
+        }
+        try {
+            $searchDao = DAORegistry::getDAO('ArticleSearchDAO'); /** @var ArticleSearchDAO $searchDao */
+            $objectId = $searchDao->insertObject($articleId, $type, $submissionFile->getId());
 
-                while (($text = $parser->read()) !== false) {
-                    $this->_indexObjectKeywords($objectId, $text);
+            do {
+                for ($buffer = ''; ($chunk = $parser->read()) !== false && strlen($buffer .= $chunk) < static::MINIMUM_DATA_LENGTH;);
```
jonasraoni, line 131: The `$parser->read()` reads the file line-by-line. As one line might have only a couple of words, that's not enough to process in batches (I've introduced some in this PR and there was another here: https://github.com/pkp/pkp-lib/blob/25b382fb04ac3ceb8377d79f8f8396e9fddd4eab/classes/search/SubmissionSearchDAO.php#L155-L158, which I've modified), so a small data accumulation is needed.

Another related issue, which I didn't address here, might happen at the `SearchHTMLParser`, which will probably have a not very well defined behavior with tags that span through N lines: https://github.com/pkp/pkp-lib/blob/25b382fb04ac3ceb8377d79f8f8396e9fddd4eab/classes/search/SearchHTMLParser.php#L23-L26

classes/search/ArticleSearchIndex.php
```diff
@@ -109,24 +113,28 @@ public function deleteTextIndex($articleId, $type = null, $assocId = null)
...
-        // If no search plug-in is activated then fall back to the
-        // default database search implementation.
-        if ($hookResult === false || is_null($hookResult)) {
-            $parser = SearchFileParser::fromFile($submissionFile);
-            if (isset($parser) && $parser->open()) {
-                $searchDao = DAORegistry::getDAO('ArticleSearchDAO'); /** @var ArticleSearchDAO $searchDao */
-                $objectId = $searchDao->insertObject($articleId, $type, $submissionFile->getId());
+        // If no search plug-in is activated then fall back to the default database search implementation.
+        $parser = SearchFileParser::fromFile($submissionFile);
+        if (!$parser?->open()) {
```
jonasraoni, line 122: Hmm, the `try/finally` should enclose this line (I wrote it before seeing the `close()`) 😁

asmecher:
Manually cherry-picked to resolve submodule conflict.

PR #3890, opened 2023-05-03
pkp/pkp-lib#8956 fix getPublishedSubmissions, getPublishedIssues and …

bozana:
…getArticleGalleys

s. https://github.com/pkp/pkp-lib/issues/8956

### DIFF
--- a/classes/issue/Collector.php
+++ b/classes/issue/Collector.php
@@ -244,9 +244,9 @@ class Collector implements CollectorInterface
     }
 
     /**
-     * Limit results to submissions that do/don't have any DOIs assign to their sub objects
+     * Limit results to issues that do/don't have any DOIs assign to their sub objects
      *
-     * @param array|null $enabledDoiTypes TYPE_* constants to consider when checking submission has DOIs
+     * @param array|null $enabledDoiTypes TYPE_* constants to consider when checking issue has DOIs
      *
      * @return $this
      */
--- a/classes/plugins/DOIPubIdExportPlugin.php
+++ b/classes/plugins/DOIPubIdExportPlugin.php
@@ -130,11 +130,18 @@ abstract class DOIPubIdExportPlugin extends PubObjectsExportPlugin
      */
     public function getPublishedSubmissions($submissionIds, $context)
     {
+        $allSubmissionIds = Repo::submission()
+            ->getCollector()
+            ->filterByContextIds([$context->getId()])
+            ->filterByStatus([PKPSubmission::STATUS_PUBLISHED])
+            ->getIds()
+            ->toArray();
+        $validSubmissionIds = array_intersect($allSubmissionIds, $submissionIds);
         $submissions = array_map(function ($submissionId) {
             return Repo::submission()->get($submissionId);
-        }, $submissionIds);
+        }, $validSubmissionIds);
         return array_filter($submissions, function ($submission) {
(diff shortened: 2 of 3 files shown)


classes/plugins/DOIPubIdExportPlugin.php
```diff
@@ -148,34 +155,39 @@ public function getPublishedSubmissions($submissionIds, $context)
...
     {
-        $publishedIssues = [];
-        foreach ($issueIds as $issueId) {
-            $publishedIssue = Repo::issue()->get($issueId);
-            $publishedIssue = $publishedIssue->getJournalId() == $context->getId() ? $publishedIssue : null;
-            if ($publishedIssue && $publishedIssue->getStoredPubId('doi')) {
-                $publishedIssues[] = $publishedIssue;
-            }
-        }
+        $publishedIssues = Repo::issue()
```
ewhanson, line 158: This could be returned directly instead of being assigned to the intermediate `$publishedIssues`.

classes/plugins/PubObjectsExportPlugin.php
```diff
@@ -785,12 +785,16 @@ public function executeCLICommand($scriptName, $command, $context, $outputFile,
      */
     public function getPublishedSubmissions($submissionIds, $context)
     {
-        $submissions = array_map(function ($submissionId) {
+        $allContextSubmissionIds = Repo::submission()
```
ewhanson, line 788: Not super important but `$allContextSubmissionIds` sounds like it should be for all contexts, not all submissions for a single context.

classes/plugins/DOIPubIdExportPlugin.php
```diff
@@ -148,34 +155,39 @@ public function getPublishedSubmissions($submissionIds, $context)
...
     {
-        $galleys = [];
-        foreach ($galleyIds as $galleyId) {
-            $articleGalley = Repo::galley()->get((int) $galleyId);
-            if ($articleGalley && $articleGalley->getStoredPubId('doi')) {
-                $galleys[] = $articleGalley;
-            }
-        }
-        return $galleys;
+        $allContextGalleyIds = Repo::galley()
```
ewhanson, line 179: Same as `$allContextSubmissionIds` below.

classes/plugins/DOIPubIdExportPlugin.php
```diff
@@ -130,11 +130,18 @@ protected function _getObjectAdditionalSettings()
      */
     public function getPublishedSubmissions($submissionIds, $context)
     {
+        $allContextSubmissionIds = Repo::submission()
```
ewhanson, line 133: Same as `$allContextSubmissionIds` below.

classes/issue/Collector.php
```diff
@@ -244,9 +244,9 @@ public function filterByDoiStatuses(?array $statuses): static
     }
 
     /**
-     * Limit results to submissions that do/don't have any DOIs assign to their sub objects
+     * Limit results to issues that do/don't have any DOIs assign to their sub objects
```
ewhanson, line 247: Oops, thanks for catching this!

classes/plugins/PubObjectsExportPlugin.php
bozana: :+1: I will then call it `allSubmissionIds` :-) Hope that that is OK...

PR #3895, opened 2023-05-09
Bugfix/main/8929 migration issues

jonasraoni:


classes/migration/upgrade/OJSv3_3_1_i6872_UrlPathMigration.php
```diff
@@ -1,42 +0,0 @@
...
-use Illuminate\Support\Facades\DB;
-
-class OJSv3_3_1_i6872_UrlPathMigration extends \PKP\migration\Migration
-{
-    /**
-     * Run the migrations.
-     */
-    public function up(): void
-    {
-        DB::table('publications')->whereNull('url_path')->update(['url_path' => null]);
```
jonasraoni, line 28: I think this migration was supposed to fix something that I've fixed a couple of days ago 🤔
But it's not doing anything and without the code updates, the data pollution would come back.

dbscripts/xml/upgrade.xml
```diff
@@ -132,11 +132,11 @@
 	</upgrade>
 
 	<upgrade minversion="3.1.0.0" maxversion="3.3.9.9">
+		<migration class="PKP\migration\upgrade\v3_4_0\I7167_RemoveDuplicatedUserSettingsAndDeprecatedFields" />
```
jonasraoni, line 135: Moved this one to the top, the other option would be to include it on the `preflight`.

Also, about the `preflight`, I think we shouldn't be doing updates there, only checks 🤔

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -35,163 +34,249 @@ protected function getContextSettingsTable(): string
...
+            'users' => ['submission_files', 'review_assignments', 'subscriptions', 'notifications', 'event_log', 'email_log', 'user_user_groups', 'user_settings', 'user_interests', 'temporary_files', 'submission_comments', 'subeditor_submission_group', 'stage_assignments', 'sessions', 'query_participants', 'notification_subscription_settings', 'notes', 'email_log_users', 'edit_decisions', 'completed_payments', 'access_keys'],
+            'submissions' => ['submission_files', 'publications', 'review_rounds', 'review_assignments', 'submission_search_objects', 'library_files', 'submission_settings', 'submission_comments', 'stage_assignments', 'review_round_files', 'edit_decisions'],
+            'submission_files' => ['submission_files', 'publication_galleys', 'submission_file_settings', 'submission_file_revisions', 'review_round_files', 'review_files'],
+            // publication_settings dependency added manually
+            'issues' => [$this->getContextTable(), 'issue_galleys', 'issue_files', 'issue_settings', 'custom_section_orders', 'custom_issue_orders', 'publication_settings'],
```
jonasraoni, line 243: `publication_settings` added manually to cover the `issueId` field.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -35,163 +34,249 @@ protected function getContextSettingsTable(): string
...
-            foreach ($orphanedIds as $userId) {
-                DB::table('completed_payments')->where('user_id', '=', $userId)->delete();
-            }
-        } catch (\Exception $e) {
-            if ($fallbackVersion = $this->setFallbackVersion()) {
-                $this->_installer->log("A pre-flight check failed. The software was successfully upgraded to {$fallbackVersion} but could not be upgraded further (to " . $this->_installer->newVersion->getVersionString() . '). Check and correct the error, then try again.');
-            }
-            throw $e;
-        }
+    protected function getEntityRelationships(): array
```
jonasraoni, line 235: This list is specific for each application, I'll write a comment in the issue with the queries that I used to generate the cleanup code and this list.

classes/migration/upgrade/v3_4_0/PreflightCheckMigration.php
```diff
@@ -35,163 +34,249 @@ protected function getContextSettingsTable(): string
...
+            $affectedRows += $this->deleteRequiredReference('subscription_type_settings', 'type_id', 'subscription_types', 'type_id');
+            return $affectedRows;
+        });
 
-            // Clean orphaned completed_payments entries by context_id
-            $orphanedIds = DB::table('completed_payments AS cp')->leftJoin('journals AS j', 'j.journal_id', '=', 'cp.context_id')->whereNull('j.journal_id')->distinct()->pluck('cp.context_id');
-            foreach ($orphanedIds as $journalId) {
-                DB::table('completed_payments')->where('context_id', '=', $journalId)->delete();
+        // Support for the issueId setting
+        $this->addTableProcessor('publication_settings', function (): int {
```
jonasraoni, line 215: Here I decided to delete the `issueId` when it's invalid. This is a fresh update, I didn't check the outcome yet,

dbscripts/xml/upgrade.xml
asmecher: > Also, about the preflight, I think we shouldn't be doing updates there, only checks

I'm OK with doing cleanup changes, but *only if* they can be run repeatedly without breaking anything.

The process of checking for bad data and the process of cleaning it up are almost the same, so the preflight migration ended up doing both. In most cases we can find and clean up at the same time; in other cases (e.g. critical issues that we can't assume we can correct automatically), we detect and fail.

asmecher: Running this before the preflight check does cause a problem, though: when the preflight fails, I get a subsequent fatal error trying to revert this migration.
```
2023-05-27 00:09:40 [A pre-flight check failed. The software was successfully upgraded to 3.3.9.9 but could not be upgraded further (to 3.4.0.0). Check and correct the error, then try again.]
2023-05-27 00:09:40 [revert migration: PKP\migration\upgrade\v3_4_0\I7167_RemoveDuplicatedUserSettingsAndDeprecatedFields]
2023-05-27 00:09:40 [downgrade for "PKP\migration\upgrade\v3_4_0\I7167_RemoveDuplicatedUserSettingsAndDeprecatedFields" unsupported: Downgrade unsupported due to removed data]
```

jonasraoni: This is a bug for me, after setting a save point, we shouldn't fallback the previous migrations. I've created an issue here with extra points/ideas: https://github.com/pkp/pkp-lib/issues/9030

But in this specific case, it's a cosmetic issue, that migration can be executed again.

jonasraoni:
Superseded by https://github.com/pkp/ojs/pull/3944

PR #3900, opened 2023-05-11
Bugfix/main/8989 fix mariadb build

jonasraoni:


.php-cs-fixer.php
```diff
@@ -3,46 +3,37 @@
...
+            new RecursiveDirectoryIterator(__DIR__ . '/plugins', FilesystemIterator::SKIP_DOTS | FilesystemIterator::UNIX_PATHS | FilesystemIterator::CURRENT_AS_FILEINFO),
+            RecursiveIteratorIterator::SELF_FIRST
+        );
+        $recursiveIterator->setMaxDepth(1);
+        $gitFolders = new CallbackFilterIterator(
+            $recursiveIterator,
+            fn (SplFileInfo $file) => $recursiveIterator->getDepth() === $recursiveIterator->getMaxDepth()
+                && $file->isDir()
+                // Covers submodules (.git file) and external repositories (.git directory)
+                && file_exists("{$file}/.git")
```
jonasraoni, line 27: The main difference from the previous code (besides the comment I left on [pkp-lib](https://github.com/pkp/pkp-lib/pull/8990/files#r1191976990)) is that it also skips plugins that have a `.git` folder.

PR #3906, opened 2023-05-15
[OJS][main] #8719 Add Send to review Action for recommending editors

defstat:


classes/decision/Repository.php
```diff
@@ -82,4 +82,11 @@ protected function getReviewNotificationTypes(): array
     {
         return [Notification::NOTIFICATION_TYPE_PENDING_EXTERNAL_REVISIONS];
     }
+
+    public function isRecommendation(int $decision): bool
+    {
+        return in_array($decision, [
```
Vitaliy-1, line 88: Should we also take into account parent decision here also? E.g.:
```php
if (parent::isRecommendation()) {
  return true
}
...
```

asmecher:
Cherry-picked manually due to submodule collision.

PR #3937, opened 2023-06-01
pkp/pkp-lib#9042 Fixed foreign key setup for the publications.primary…

jonasraoni:
…_contact_id

### DIFF
--- a/classes/migration/upgrade/v3_4_0/I6093_AddForeignKeys.php
+++ b/classes/migration/upgrade/v3_4_0/I6093_AddForeignKeys.php
@@ -88,7 +88,7 @@ class I6093_AddForeignKeys extends \PKP\migration\upgrade\v3_4_0\I6093_AddForeig
         Schema::table('publications', function (Blueprint $table) {
             $table->foreign('section_id', 'publications_section_id')->references('section_id')->on('sections')->onDelete('set null');
             $table->foreign('submission_id', 'publications_submission_id')->references('submission_id')->on('submissions')->onDelete('cascade');
-            $table->foreign('primary_contact_id', 'publications_author_id')->references('author_id')->on('authors')->onDelete('set null');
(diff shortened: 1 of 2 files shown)


classes/migration/upgrade/v3_4_0/I6093_AddForeignKeys.php
```diff
@@ -88,8 +88,8 @@ public function up(): void
         Schema::table('publications', function (Blueprint $table) {
             $table->foreign('section_id', 'publications_section_id')->references('section_id')->on('sections')->onDelete('set null');
             $table->foreign('submission_id', 'publications_submission_id')->references('submission_id')->on('submissions')->onDelete('cascade');
-            $table->foreign('primary_contact_id', 'publications_author_id')->references('user_id')->on('users')->onDelete('set null');
-            $table->index(['primary_contact_id'], 'publications_user_id');
+            $table->foreign('primary_contact_id', 'publications_author_id')->references('author_id')->on('authors')->onDelete('set null');
+            $table->index(['primary_contact_id'], 'publications_primary_contact_id');
```
asmecher, line 92: `classes/migration/install/OJSMigration.php` needs to be updated to match this index name.

jonasraoni: Yep, I'll also update the foreign key index name in both places to `publications_primary_contact_id`.

In general, I think it would be better to let the library handle the name, to avoid having different patterns:
https://github.com/pkp/ojs/blob/04d2f382b020f33649f97756b96df843b9f008f1/classes/migration/install/MetricsMigration.php#L105

asmecher: Unfortunately we can't rely 100% on Laravel to handle it, as it can generate overlong names.

jonasraoni: I've just checked and the max length for both MySQL and PgSQL seems to be 64, so yeah, better to keep the names under control 😁

PR #3946, opened 2023-06-07
Feature/main/8333 Add remaining FKs

jonasraoni:


dbscripts/xml/upgrade.xml
```diff
@@ -204,6 +204,10 @@
 		<note file="docs/release-notes/README-3.4.0" />
 	</upgrade>
 
+	<upgrade minversion="3.1.0.0" maxversion="3.4.9.9">
+		<migration class="APP\migration\upgrade\v3_5_0\I8333_AddMissingForeignKeys" />
```
asmecher, line 208: Does this need to be run so early in the upgrade process? I would prefer if any 3.5 upgrades ran after all 3.4 upgrades (and so on), so that the upgrade process is linear.

jonasraoni: I didn't inspect how the integration between the `minversion`/`maxversion` and the list of `<upgrade>`s work, so I just added it here to not be forgotten =]

asmecher: The `minversion`/`maxversion` don't change during the list of upgrade steps, so this should go at the end.

jonasraoni:
@asmecher you may review this PR and the one for [pkp-lib](https://github.com/pkp/pkp-lib/pull/10023), the others are outdated.

I've tried to deprecate/merge constants from the `Application`, and use `null` everywhere instead of converting to/from `0`.
Special cases are using `Application::SITE_CONTEXT_ID_ALL`.

jonasraoni:
A mixed type could make things clearer for the programmer (which functions work with the site context) `int|SiteContext $contextId`.

classes/issue/Collector.php
```diff
@@ -328,14 +326,7 @@ public function getQueryBuilder(): Builder
         );
 
         // Context
-        // Never permit a query without a context_id unless the PKPApplication::CONTEXT_ID_ALL wildcard
-        // has been set explicitly.
-        if (!isset($this->contextIds)) {
-            throw new Exception('Submissions can not be retrieved without a context id. Pass the Application::CONTEXT_ID_ALL wildcard to get submissions from any context.');
```
asmecher, line 334: This check was coded to make it intentionally harder to forget the context ID filter; I think it's a good idea. It could be coded a different way (e.g. requiring `[null]` to be set as a context ID filter), but I'd rather it wasn't totally removed.

classes/services/StatsIssueService.php
```diff
@@ -131,10 +131,7 @@ public function getDefaultArgs(): array
         return [
             'dateStart' => StatisticsHelper::STATISTICS_EARLIEST_DATE,
             'dateEnd' => date('Y-m-d', strtotime('yesterday')),
-
-            // Require a context to be specified to prevent unwanted data leakage
-            // if someone forgets to specify the context.
-            'contextIds' => [\PKP\core\PKPApplication::CONTEXT_ID_NONE],
+            'contextIds' => [],
```
asmecher, line 134: As in the pkp-lib PR, I'm not sure we want to remove this check.

asmecher (changes requested):
Just a comment or two!

classes/issue/Collector.php
jonasraoni: I removed because I see this as a kind of exception treatment. Even though the context represents a big wall, data might be leaked in many different ways/places, and other places might be missing such checks.
But nothing against reverting the changes, just wanted to raise this small discussion.

asmecher: It's true that we're not applying these checks universally (and it would be tough to set a firm policy on when to apply them) -- but they are present here for issues in OJS and [submissions in pkp-lib](https://github.com/pkp/pkp-lib/blob/main/classes/submission/Collector.php#L364). Those are arguably the two most critical access points for published data (and by extension, related entities like publications, files, etc). They're just a few lines of code, and I don't see much of a downside. I'd feel better if we didn't remove them.

jonasraoni: Ok! In general, I'm against such protections, I'd like to think programmers know what they're doing, but we fail more often that we'd like to admit 😂

PR #3952, opened 2023-06-21
pkp/pkp-lib#5953 Issue publish mail notification toc support

touhidurabir:
for pkp/pkp-lib#5953


classes/mail/variables/IssueEmailVariable.php
```diff
@@ -67,4 +71,33 @@ protected function getIssueUrl(): string
...
         );
     }
+
+    protected function getIssueToc(): string
+    {
+        $request = Application::get()->getRequest();
+
+        $submissions = Repo::submission()->getCollector()
+            ->filterByContextIds([$this->getContext()->getId()])
+            ->filterByIssueIds([$this->issue->getId()])
```
asmecher, line 81: I think you'll also need a filter on submission status:
```php
->filterByStatus([PKPSubmission::STATUS_PUBLISHED])
```
...and the right ordering:
```php
->orderBy(\APP\submission\Collector::ORDERBY_SEQUENCE, \APP\submission\Collector::ORDER_DIR_ASC)
```

classes/mail/variables/IssueEmailVariable.php
```diff
@@ -67,4 +71,33 @@ protected function getIssueUrl(): string
...
+            ->getMany();
+
+        if($submissions->count() <= 0) {
+            return '';
+        }
+
+        $list = $submissions->map(function ($submission) use ($request) {
```
asmecher, line 88: This will just include the submission titles, without the section information. I wonder whether it would be possible to generate this for the email by re-using `templates/frontend/objects/issue_toc.tpl` rather than trying to reimplement the functionality there over again...

classes/mail/variables/IssueEmailVariable.php
```diff
@@ -67,4 +71,33 @@ protected function getIssueUrl(): string
...
+            $publication = $submission->getCurrentPublication();
+            $url = $request->getDispatcher()->url(
+                $request,
+                Application::ROUTE_PAGE,
+                $this->getContext()->getPath(),
+                'article',
+                'view',
+                $publication->getData('urlPath') ?? $submission->getId()
+            );
+            return '<li><a href="' . $url . '">' . $publication->getLocalizedFullTitle(null, 'html') . '</a></li>';
```
asmecher, line 98: I think the title will still need to get passed through the safe HTML filter, to ensure malicious HTML isn't passed through directly.

asmecher (changes requested):

classes/mail/variables/IssueEmailVariable.php
touhidurabir: Thanks, updated .

classes/mail/variables/IssueEmailVariable.php
touhidurabir: used the `templates/frontend/objects/issue_toc.tpl` but this is how it looks, not very good looking . 

CleanShot 2023-07-25 at 17 36 05@2x

asmecher: I'm still on the fence about whether or not we should use the same template for the email and normal TOC presentations -- but in any case, it would be nice not to have to go through duplicating all the effort of fetching the data that would be included in the template. We're likely to badly maintain it and miss out on important features that way (thinking ordering, subscriptions, plugin-based interventions, etc).

We already have `IssueHandler::_setupIssueTemplate` made `static` so that it can be reused (see e.g. `pages/index/IndexHandler.php`, which uses it to present the current issue on the journal homepage when that option is selected). I can think of a few options...
- Add some flags to the existing `issue_toc.tpl` template to allow calling code to have better control over the presentation (e.g. showing/hiding the publication date, if we don't want that in the email)
- Reuse `IssueHandler::_setupIssueTemplate`, but implement a new template file that is a stripped-down version of the current `issue_toc.tpl`
- Re-use both `IssueHandler::setupIssueTemplate` and `issue_toc.tpl`, but experiment with styling the output of the current `issue_toc.tpl` in order to make it nicer when presented in an email

classes/mail/variables/IssueEmailVariable.php
```diff
@@ -67,4 +76,72 @@ protected function getIssueUrl(): string
...
+            'currentJournal' => $this->getContext(),
+            'issue' => $this->issue,
+            'issueGalleys' => $issueGalleyDao->getByIssueId($this->issue->getId()),
+            'publishedSubmissions' => $issueSubmissionsInSection,
+            'primaryGenreIds' => $primaryGenreIds,
+            'authorUserGroups' => $authorUserGroups,
+            'locale' => Locale::getLocale(),
+        ]);
+
+        return PKPString::stripUnsafeHtml($templateMgr->fetch('frontend/objects/issue_toc.tpl'));
```
asmecher, line 145: I'm not sure about this `stripUnsafeHtml` wrapper. Is there specific content in the template that should be stripped? Or is this in lieu of having better control of the formatting?

asmecher (changes requested):

classes/mail/variables/IssueEmailVariable.php
touhidurabir: Reused the `IssueHandler::_setupIssueTemplate` but not possible to make it prettier as this only load the HTML, not the CSS and to to do that we need to build a different template file . take a look at the current view and if we decide to build a new template file, we can do that 

CleanShot 2023-08-01 at 17 52 07@2x

classes/mail/variables/IssueEmailVariable.php
```diff
@@ -43,6 +50,7 @@ public static function descriptions(): array
             static::ISSUE_ID => __('emailTemplate.variable.issueId'),
             static::ISSUE_IDENTIFICATION => __('emailTemplate.variable.issue.issueIdentification'),
             static::ISSUE_URL => __('emailTemplate.variable.issue.issuePublishedUrl'),
+            static::ISSUE_TOC => __('emailTemplate.variable.issue.issueTableOfContent'),
```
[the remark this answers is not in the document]
touhidurabir: Done .

classes/mail/variables/IssueEmailVariable.php
touhidurabir: used this to removed the extra lines and white spaces but probably better to not use it as that make the list too compact . I have removed it .

PR #3984, opened 2023-07-24
I9180 cypress12

jardakotesovec:


cypress/tests/data/10-ApplicationSetup/50-CreateSections.cy.js
```diff
@@ -18,7 +18,7 @@ describe('Data suite tests', function() {
 		// Edit Articles section to add section editors
 		cy.get('div#sections a[class=show_extras]').click();
 		cy.get('a[id^=component-grid-settings-sections-sectiongrid-row-1-editSection-button-]').click();
-		cy.waitJQuery(); // Avoid occasional failure due to form init taking time
+		cy.wait(1000); // Avoid occasional failure due to form init taking time, waitJQuery is not sufficient
```
jonasraoni, line 21: I understand Cypress might capture a bad element (e.g. interface got re-rendered and the element was removed in-between) or in this case, probably captured the element before it was initialized... Therefore, if you couldn't find a condition to keep Cypress retrying the query, no objections from me ¯\_(ツ)_/¯

cypress/tests/integration/Statistics.cy.js
```diff
@@ -28,8 +28,8 @@ describe('Statistics Tests', function() {
 			'Article Details',
 			'articles',
 			['Mwandenga', 'Karbasizaed'],
-			4,
-			2
+			2,
+			1
```
jonasraoni, line 32: The tests should be more independent, so we can run them in parallel and also avoid side-effects, but until we get it done, that's what we can do 😁

jonasraoni (approved):

cypress/tests/data/10-ApplicationSetup/50-CreateSections.cy.js
jardakotesovec: @Vitaliy-1 I was getting TinyMCE related error with older cypress (likely due to older electron). That went away reliably after moving to cypress 12. But in this case the form changes were not picked up. I was not able to exactly pin point why. Some experiments suggested that its because of the animation of the modal. Not sure. I concluded that for older stack its fine to fallback to `wait` if its improves reliability. If that would be happening with Vue.js stack I would investigate it further.

Vitaliy-1: Sounds good, let's try then.

asmecher: I think I'd favour leaving both waits in case -- the `waitJQuery()` *should* be a good interlocked wait, and the `wait(1000)` is an additional wait that should be noted as a `FIXME`. (I see that's what you've done in one case below.)

cypress/tests/integration/Statistics.cy.js
asmecher: Were the earlier numbers doubled to accommodate the changes introduced to the submission counts by the import/export tests running first?

jardakotesovec: @asmecher Yes exactly. I moved the import/export to the end as it was breaking Subscription.py in intermittent way. As the same submission existed twice with same title. And since there is mechanism to resolve the submission (on frontend) by the title - it in some cases (by luck and database - it was failing more on psql) pick wrong submission in database.

cypress/tests/data/10-ApplicationSetup/50-CreateSections.cy.js
jardakotesovec: @asmecher Technically neither of these would be needed. Cypress by default waits for element to show up. So once the input is in DOM it will start typing in.  `waitJQuery` is really just waiting for ajax calls to finish - which as result creates some delay before it starts typing in, which increase chances that the form will be ready, but not deterministically.  I was trying to exactly pin point what to wait for deterministically, but failed to identify it in couple of hours so I moved on..  I suspect its related to transition animations - when I disabled animation the binding fall apart completely.  At that point I was thinking lets stick with `wait` for these old stack pieces, which seems to work good enough. 

Also note that in this case it does not interact with tinyMCE textarea, just with simple inputs.

In Vue.js we generally should not end up in situation where you have some form elements in DOM and events are not binded correctly before it can be interact with. 

Using `wait` on Vue.js code base would suggest that we need to do some fixing for sure.

So I can add it.. but I don't think it makes difference.

I might be wrong about some of these observations..  it gets bit esoteric if I can reproduce it just some times and just in head less mode.

cypress/tests/integration/Statistics.cy.js
jardakotesovec: @asmecher resetting database for each integration test to the "after submissions state" would be worthy improvement.

asmecher: Gotcha, thanks! I agree, let's resolve the database reset process as a priority.

PR #3987, opened 2023-07-26
pkp/pkp-lib#9109 Migrate missing email templates

Vitaliy-1:


classes/migration/upgrade/v3_4_0/InstallEmailTemplates.php
```diff
@@ -38,6 +38,8 @@ protected function getEmailTemplateKeys(): array
             'SUBMISSION_NEEDS_EDITOR',
             'PAYMENT_REQUEST_NOTIFICATION',
             'VERSION_CREATED',
+            'REVIEW_COMPLETE',
```
asmecher, line 41: @Vitaliy-1, I think this will leave a "hole" for users who have already updated to 3.4.0-[01]. Subsequent 3.4.0-x upgrades won't re-run this migration, and when they later upgrade to 3.5.0 or whatever, these templates won't get installed.

In cases like this, we generally fix the broken migration (which you've done here) but also add a catch-up migration to the 3.5.0 upgrade set to catch the missing cases.

Vitaliy-1: I've planned to open an issue for 3.5 migration

asmecher: Perfect, thanks -- please go ahead!

PR #4008, opened 2023-08-16
 [Exercise] Developer Orientation

Henrique523:
Pull request to answer the Exercise 7 of the Developer Orientation guide


pages/exercise/AnnouncementHandler.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+/**
+ * @file pages/about/AboutHandler.php
+ *
+ * Copyright (c) 2014-2021 Simon Fraser University
```
asmecher, line 6: Not a big deal -- we're not 100% consistent with this -- but for new code, it should be (c) the current year:
```
 * Copyright (c) 2023 Simon Fraser University
 * Copyright (c) 2023 John Willinsky
```

pages/exercise/AnnouncementHandler.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        $templateMgr = TemplateManager::getManager($request);
+        $templateMgr->assign([
+            'pageTitle' => __('exercises.exercises'),
+        ]);
+        $templateMgr->display('frontend/pages/exerciseIndex.tpl');
+    }
+
+    public function announcements($args, Request $request)
```
asmecher, line 63: Also not a big deal :) but we're making increased use of typehints in new code. For the `main` and `stable-3_4_0` branches, please add formal typehints as much as possible (e.g. `array $args` here).

pages/exercise/index.php
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright (c) 2003-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @ingroup pages_exercise
+ *
+ * @brief Handle requests for about the journal functions.
+ *
+ */
+
+use APP\pages\exercise\AnnouncementHandler;
```
asmecher, line 20: There's a mismatch here between `exercise` (the page name, which becomes part of the URL) and announcements (the handler name, the `@defgroup`, etc). If this were going to be merged, we would reconcile those for consistency. But as this is just an exercise, no need to do that here, I think.

templates/frontend/pages/announcements.tpl
```diff
@@ -0,0 +1,27 @@
+{**
+ * templates/frontend/pages/announcements.tpl
+ *
+ * Copyright (c) 2013-2021 Simon Fraser University
+ * Copyright (c) 2003-2021 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * About the Journal Subscriptions.
```
asmecher, line 8: (Incorrect self-documentation)

templates/frontend/pages/announcements.tpl
```diff
@@ -0,0 +1,27 @@
...
+    <pkp-header></pkp-header>
+
+    <h1 class="app__pageHeading">
+        {translate key=$pageTitle}
+    </h1>
+
+    <ul>
+        {foreach from=$announcements item=announcement}
+            <li style="margin: 8px 0;">
+                {$announcement->getId()}  / {$announcement->getLocalizedData('title')} / {$announcement->getDatePosted()} / {$announcement->getLocalizedData('keyword')} / <pkp-button element="a" href="http://localhost:8000/index.php/first-journal/exercise/announcements/{$announcement->getId()|strip_unsafe_html}" is-link>View</pkp-button>
```
asmecher, line 23: Be careful with escaping; the announcement title is expected to be plain text, so it should be passed through the `|escape` Smarty modifier to make sure that we aren't vulnerable to XSS leaks, and that `<` / `>` / `&` characters in announcement titles are appropriately presented. Generally it's a good idea to `|escape` just about everything; there are only a few cases where it's not a good idea. Likewise, `$announcement->getId()` should be passed through `|escape`, not `|strip_unsafe_html`.

asmecher: URL construction should be done with the `{url ...}` Smarty function; here the host, port, etc. are hard-coded, which will be incorrect.

asmecher: `View` is hard-coded in English; it's best to localize all language.

PR #4032, opened 2023-09-06
I8919 vue3&vite migration

jardakotesovec:


.prettierignore
```diff
@@ -0,0 +1,2 @@
+cypress/tests/**/*
+lib/pkp/cypress/**/*
```
jardakotesovec, line 2: This we can remove in future, its just to avoid big formatting changes for this PR.

cypress/tests/data/60-content/AmwandengaSubmission.cy.js
```diff
@@ -337,11 +337,11 @@ describe('Data suite: Amwandenga', function() {
...
-		cy.get('#contributors [name="familyName-en"]').type('Riouf', {delay: 0});
-		cy.get('#contributors [name="email"]').type('nriouf@mailinator.com', {delay: 0});
-		cy.get('#contributors [name="country"]').select('South Africa');
-		cy.get('#contributors button').contains('Save').click();
+		cy.get('.modal [name="givenName-en"]').type('Nicolas', {delay: 0});
+		cy.get('.modal [name="familyName-en"]').type('Riouf', {delay: 0});
+		cy.get('.modal [name="email"]').type('nriouf@mailinator.com', {delay: 0});
+		cy.get('.modal [name="country"]').select('South Africa');
+		cy.get('.modal button').contains('Save').click();
 		cy.wait(500);
```
jardakotesovec, line 345: These changes are just because of different DOM structure with new modal.

js/load.js
```diff
@@ -65,7 +63,3 @@ window.pkp = Object.assign(PkpLoad, {
 		SubmissionsPage,
 	},
 });
-
-// Required by the URN plugin
-window.pkp.Vue.component('field-text', FieldText);
-window.pkp.Vue.component('field-pub-id', FieldPubId);
```
jardakotesovec, line 71: This is moved to pkp/js/load.js as it needs to be inside the pkpCreateVueApp to be available in every vue instance.

PR #4033, opened 2023-09-07
pkp/pkp-lib#9282 Duplicate DOI setting keys

ewhanson:


tools/resolveAgencyDuplicates.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return DB::table('submission_settings')
+            ->whereIn('setting_name', array_map(fn ($item) => $item['status'], $this->agencies))
+            ->groupBy('submission_id')
+            ->havingRaw('COUNT(submission_id) > 1')
+            ->select(['submission_id'])
+            ->get()
+            ->map(function ($item) {
+                return $item->submission_id;
+            });
```
bozana, line 78: Would here be the same to use `->pluck('submission_id')` instead of 
```
->select(['submission_id'])
->get()
->map(function ($item) {
return $item->submission_id;
})
```

ewhanson: Yep!

ewhanson:
@bozana, this should be ready for merge now. I can cherry pick/update main once this is merged. Thanks!

bozana:
Thanks a lot @ewhanson! Merged...

PR #4080, opened 2023-10-20
#9449 Attach the issue properly when calling issueform::execute

henriquejsfj:
I decided to call the Hook in the end with the `parent` call instead of previous `Hook::call('issueform::execute', [$this, $issue]);`.

This may broke some plugins that rely on that `$issue` as an argument. However, I didn´t find any plugin using that Hook on my installation. Besides, nobody noticed it before... But we may change it if you prefer.

### DIFF
--- a/controllers/grid/issues/form/IssueForm.php
+++ b/controllers/grid/issues/form/IssueForm.php
@@ -221,7 +221,7 @@ class IssueForm extends Form
             $issue = $this->issue;
         } else {
             $issue = Repo::issue()->newDataObject();
-            $this->issue =& $issue;
+            $this->issue = $issue;
             switch ($journal->getData('publishingMode')) {
                 case \APP\journal\Journal::PUBLISHING_MODE_SUBSCRIPTION:
                 case \APP\journal\Journal::PUBLISHING_MODE_NONE:
--- a/controllers/grid/issues/form/IssueForm.php
+++ b/controllers/grid/issues/form/IssueForm.php
@@ -213,8 +213,6 @@ class IssueForm extends Form
      */
     public function execute(...$functionArgs)
     {
-        parent::execute(...$functionArgs);
-
         $request = Application::get()->getRequest();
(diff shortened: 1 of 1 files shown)


henriquejsfj:
Issue: https://github.com/pkp/pkp-lib/issues/9449

controllers/grid/issues/form/IssueForm.php
```diff
@@ -278,7 +277,7 @@ public function execute(...$functionArgs)
 
         $issue->setCoverImageAltText($this->getData('coverImageAltText'), $locale);
 
-        Hook::call('issueform::execute', [$this, $issue]);
+        parent::execute(...$functionArgs);
```
bozana, line 280: Instead of having 
`$this->issue =& $issue;` 
I would suggest to have the following here:
```
$functionArgs[] = $issue;
parent::execute(...$functionArgs);
```
This way, the user of the hook will have both information, about the old and about the new issue -- if, for example, also an access to the issue's old metadata is needed for a reason.

I think that we are not consistent in the forms, but I think the old hook 
`Hook::call('issueform::execute', [$this, $issue]);`
was enabling/considering this (case/example mentioned above), so lets keep it.

bozana: Ah, yes, I was wrong -- it seems we did not have the old issue metadata...
Thus I think your solution is good, just do we need that call by reference $this->issue =**&** $issue; ?

henriquejsfj: I guess don't need, I will check

bozana:
Great @henriquejsfj!
For what OJS version do you need this change? -- We merge it now into main branch which will go into 3.5. I can cherry-pick it to stable-3_4_0 if you would need it there i.e. in the OJS 3.4.0.x release?

henriquejsfj:
@bozana I need it in the OJS 3.4
Thanks!

PR #4088, opened 2023-10-30
pkp/pkp-lib#8887 REST API for dashboard views

Vitaliy-1:


asmecher (commented):
Looks good, @Vitaliy-1, but shouldn't most of the locale keys be added in `pkp-lib` if the other apps will be using them too?

PR #4096, opened 2023-11-04
pkp/pkp-lib#9487 Skipped non-conforming setting tables from failing the migration

jonasraoni:


asmecher (changes requested):
This will work for cases like `rt_settings` where there is no `setting_type` column, but other cases where the result columns are not as expected will not work. What do you think about adding an explicit list of necessary columns that are used in the `_toJSON` function, and checking them against the table's column list? Alternately, it would be OK IMO to specifically exclude the `rt_settings` table, as that's a known pitfall for users with old databases.

PR #4097, opened 2023-11-05
pkp/pkp-lib#9475 Move layout files (OJS 2.x) from the stage SUBMISSION_FILE_PROOF to SUBMISSION_FILE_PRODUCTION_READY

jonasraoni:


asmecher:
@jonasraoni, I think this assumes `file_stage=10` will only exist for old content migrated from OJS 2.x; taking a glance at a random DB, I see much newer entries with `file_stage=10`. Are you sure this won't have side-effects for installs with 3.x data?

jonasraoni:
@asmecher The `file_stage` 10 is used by the galleys, so this is expected to exist on newer installations.
I'm not just checking the `file_stage`, there's also the check for nulled `assoc_type` + `assoc_id` (which are required for galleys to be displayed on the UI), so I think it's safe.

asmecher:
Thanks, please go ahead!

PR #4132, opened 2024-01-04
pkp/pkp-lib#9566 Laravel auth system integration

touhidurabir:
for pkp/pkp-lib#9566

### DIFF
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -147,9 +147,9 @@ sandbox = Off
 
 driver = mysqli
 host = localhost
-username = ojs
-password = ojs
-name = ojs
+username = ojs-ci
+password = ojs-ci
+name = ojs-ci
 
 ; Set the non-standard port and/or socket, if used
 ; port = 3306
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -147,9 +147,9 @@ sandbox = Off
 
 driver = mysqli
 host = localhost
-username = ojs-ci
-password = ojs-ci
-name = ojs-ci
+username = ojs
+password = ojs
+name = ojs
 
 ; Set the non-standard port and/or socket, if used
 ; port = 3306
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -50,10 +50,10 @@ session_lifetime = 30
 ; To set the "Secure" attribute for the cookie see the setting force_ssl at the [security] group
 session_samesite = Lax
 
-; Enable this if you want to enable cookie encryption
-; The length of cookie encryption key lenght must be 16
+; Enable this if want to enable cookie encryption
+; The length of the cookie encryption key must be 16 characters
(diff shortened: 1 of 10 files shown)


config.TEMPLATE.inc.php
```diff
@@ -50,6 +50,11 @@
 ; To set the "Secure" attribute for the cookie see the setting force_ssl at the [security] group
 session_samesite = Lax
 
+; Enable this if you want to enable cookie encryption
+; The length of cookie encryption key lenght must be 16
+; Note that updating or removing cookie encryption key will result in logout from all devices
+; session_cookie_enctyption_key = ''
```
asmecher, line 56: Typo: should be `encryption`, not `enctyption`. (Probably needs syncing elsewhere.)

config.TEMPLATE.inc.php
```diff
@@ -50,6 +50,11 @@
 ; To set the "Secure" attribute for the cookie see the setting force_ssl at the [security] group
 session_samesite = Lax
 
+; Enable this if you want to enable cookie encryption
+; The length of cookie encryption key lenght must be 16
```
asmecher, line 54: *`length`, not `lenght`

Suggest instead: "The length of the cookie encryption key must be 16 characters"

Is that 16 characters or more?

asmecher (changes requested):

config.TEMPLATE.inc.php
touhidurabir: > Is that 16 characters or more

No , it must be 16 chars as we are using the default `cipher algorithms` as `aes-128-cbc` which set by default . what laravel does is that based on what set in their `app` config file , it generate `app key` and that being used for encryption . But we do not have such mechanism in our system right now so we are using the default one defined in `\Illuminate\Encryption\Encrypter` which is `aes-128-cbc` . We have these following 
```php
/**
     * The supported cipher algorithms and their properties.
     *
     * @var array
     */
    private static $supportedCiphers = [
        'aes-128-cbc' => ['size' => 16, 'aead' => false],
        'aes-256-cbc' => ['size' => 32, 'aead' => false],
        'aes-128-gcm' => ['size' => 16, 'aead' => true],
        'aes-256-gcm' => ['size' => 32, 'aead' => true],
    ];
```
but if we want to support other, we need it as the part of our config information .

asmecher: I think we're eventually going to need a Laravel-compatible application key, so we might want to just bite the bullet and add one to the configuration file on upgrade (or ask the user to). There are some (old) tools to do this in `lib/pkp/classes/config/Config.inc.php` -- see e.g. `updateConfig`, `writeConfig`, and `getFileContents`. Watch out for cases where the upgrade tools don't have permission to edit the configuration file; in that case, the user should be instructed to add the missing line of configuration to the file.

touhidurabir: @asmecher I think we should add this as part of different issue as this will kind of interact with some different part of app bootstrapping process . Once this is merged, I can create a new issue and work on that . what do you think ?

asmecher:
Merged manually (submodule collision)

PR #4133, opened 2024-01-05
pkp/pkp-lib#6296 Article breadcrumbs use aria-current but do not point to the current page  - OJS 3.4

israelcefrin:
This PR fixes the issue https://github.com/pkp/pkp-lib/issues/6296 and also implements:

- [x] `aria-current="page"` attribute was not being use on a link to the current page
- [x] Removed `aria-current="page"` attribute from li element, it should to be used on link elements only
- [x] Include a link, using the page title as the label, to the last item on the breadcrumb (article and archive pages)
- [x] Added `aria-hidden="true"` to the breadcrumb separators (they were announced as "slash" by screen readers which is annoying for screen reader users and don't add any information to the context)

Even though showing the full article title might look overwhelming for sighted people, it is helpfull to people that rely on assistive technology like screen readers or screen magnifiers. They can double-check whether they are on the expected page through the breadcrumb interface element.

### DIFF
--- a/templates/frontend/components/breadcrumbs_article.tpl
+++ b/templates/frontend/components/breadcrumbs_article.tpl
@@ -14,37 +14,37 @@
  * @uses $currentTitleKey string Translation key for title of current page.
  * @uses $issue Issue Issue this article was published in.
  *}
-
-<nav class="cmp_breadcrumbs" role="navigation" aria-label="{translate key="navigation.breadcrumbLabel"}">
+{assign var=articlePath value=$publication->getData('urlPath')|default:$article->getId()}
+<nav class="cmp_breadcrumbs" aria-label="{translate key="navigation.breadcrumbLabel"}">
 	<ol>
 		<li>
 			<a href="{url page="index" router=\PKP\core\PKPApplication::ROUTE_PAGE}">
 				{translate key="common.homepageNavigationLabel"}
 			</a>
-			<span class="separator">{translate key="navigation.breadcrumbSeparator"}</span>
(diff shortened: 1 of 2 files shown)


israelcefrin:
Hi @asmecher , could you please have a look and let me know if everything is fine? This PR fixes the main issue from this one: https://github.com/pkp/pkp-lib/issues/6296 and also covers more improvements on ARIA attributes that were missing or misplaced.

asmecher:
@jardakotesovec, could you review this? I'm probably the last person who should be looking at front-end markup :)

jardakotesovec:
@israelcefrin Thank you! Looks good to me. You mentioned that the aria-current should be used only on links? If thats the case, should not we also update the other cases to be links? For example if we go to /index.php/publicknowledge/about, than  `About the Journal` is not link.

israelcefrin:
> @israelcefrin Thank you! Looks good to me. You mentioned that the aria-current should be used only on links? If thats the case, should not we also update the other cases to be links? For example if we go to /index.php/publicknowledge/about, than `About the Journal` is not link.

Hi @jardakotesovec , you are correct. And I am already tackling it , i.e. the top menu , on this issue: https://github.com/pkp/pkp-lib/issues/5199

israelcefrin:
Hi @jardakotesovec , I've implemented the same feature on the breadcrumbs of OPS and OMP Default Theme breadcrumb template files.

OPS 3.4 PR: https://github.com/pkp/ops/pull/630
OMP 3/4 PR: https://github.com/pkp/pkp-lib/pull/9684

PR #4137, opened 2024-01-08
pkp/pkp-lib#8248 COUNTER R5 TSV reports 

bozana:
s. https://github.com/pkp/pkp-lib/issues/8248

### DIFF
--- a/classes/sushi/IR.php
+++ b/classes/sushi/IR.php
@@ -19,6 +19,9 @@ namespace APP\sushi;
 
 use APP\core\Services;
 use APP\facades\Repo;
+use Illuminate\Support\Collection;
+use PKP\components\forms\FieldOptions;
+use PKP\components\forms\FieldText;
 use PKP\statistics\PKPStatisticsHelper;
 use PKP\sushi\CounterR5Report;
 
@@ -94,7 +97,8 @@ class IR extends CounterR5Report
             'attributes_to_show',
             'include_component_details',
             'include_parent_details',
-            'granularity'
+            'granularity',
+            '_', // for ajax requests
         ];
     }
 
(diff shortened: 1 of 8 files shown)


classes/sushi/IR.php
```diff
@@ -243,11 +245,16 @@ public function getReportItems(): array
...
         }
+        return $results;
+    }
+
+    /** Get report items */
+    public function getReportItems(): array
+    {
+        $results = $this->getQueryResults();
 
         $resultsGroupedBySubmission = $items = [];
```
ewhanson, line 256: `$resultsGroupedBySubmission` is immediately overwritten, so this line can be removed.

classes/sushi/IR.php
```diff
@@ -371,4 +378,254 @@ public function getReportItems(): array
...
+
+        return [$columnRow];
+    }
+
+    /** Get TSV report rows */
+    public function getTSVReportItems(): array
+    {
+        $results = $this->getQueryResults();
+
+        $resultsGroupedBySubmission = $resultRows = [];
```
ewhanson, line 446: `$resultsGroupedBySubmission` is immediately overwritten, so this line can be removed.

classes/sushi/IR.php
```diff
@@ -371,4 +378,254 @@ public function getReportItems(): array
...
+                        $this->platformName, // Platform
+                    ];
+
+                    if (in_array('Authors', $this->attributesToShow)) {
+                        $authors = $currentPublication->getData('authors');
+                        $authorRowValue = '';
+                        foreach ($authors as $author) {
+                            $authorRowValue = $author->getFullName(true, false, $submissionLocale);
+                            $orcid = $author->getOrcid();
+                            if (isset($orcid) && !empty($orcid)) {
```
ewhanson, line 484: I think `!empty($orcid)` should be all that's needed because if `$author->getOrcid()` returns nothing, `$orcid` will be set to `null` and the `!empty($orcid)` check should be enough.

classes/sushi/IR.php
ewhanson: I also just noticed this means `$items is being set to an empty array as well. I think only `$reusltsGroupedBySubmisison` needs to be removed here.

classes/sushi/IR.php
ewhanson: Same as above about only removing `$resultsGroupedBySubmission`, not `$resultRows`.

PR #4143, opened 2024-01-15
pkp/pkp-lib#9296 Sandbox settings

touhidurabir:
for pkp/pkp-lib#9296

### DIFF
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -136,7 +136,6 @@ user_validation_period = 28
 
 ; Turn sandbox mode to On in order to prevent the software from interacting with outside systems.
 ; Use this for development or testing purposes.
-: Affected services are email, datacite, crossref, orcid and paypal payment
 sandbox = Off
 
 
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -558,3 +558,30 @@ job_runner_max_memory = 80
 ; Remove failed jobs from the database after the following number of days.
 ; Remove this setting to leave failed jobs in the database.
 delete_failed_jobs_after = 180
+
+
+;;;;;;;;;;;;;;;;;;;;
+; Sandbox Settings ;
+;;;;;;;;;;;;;;;;;;;;
+
+[sandbox]
+
+; Enable the sandbox mode for email
+; Uncomment the following line to disable email sending
+; email = Off
+
+; Enable the sandbox mode for crossref
+; Uncomment the following line to disable crossref deposite
+; crossref = Off
+
+; Enable the sandbox mode for datacite
+; Uncomment the following line to disable datacite deposite
+; datacite = Off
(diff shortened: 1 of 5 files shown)


plugins/generic/datacite/DatacitePlugin.php
```diff
@@ -59,6 +60,11 @@ public function register($category, $path, $mainContextId = null)
     {
         $success = parent::register($category, $path, $mainContextId);
         if ($success) {
+            // Application is set to sandbox mode and will not run the features of plugin
+            if (Config::getVar('general', 'sandbox', false)) {
+                return false;
```
asmecher, line 65: I don't think this is where you want to intervene -- it's actually the `depositXML` function that sends the XML to the outside service. This will prevent the plugin from showing up at all, and it's just the external interaction you want to prevent.

touhidurabir: fixed

plugins/generic/datacite/DataciteExportPlugin.php
```diff
@@ -242,6 +242,12 @@ public function exportAsDownload(Context $context, array $objects, ?bool $noVali
      */
     public function depositXML($object, $context, $filename)
     {
+        // Application is set to sandbox mode and will not run the features of plugin
+        if (Config::getVar('general', 'sandbox', false)) {
+            error_log('Application is set to sandbox mode and datacite will not do any deposition');
+            return [[__('common.sandbox')]];
```
asmecher, line 248: Have you tested this? I think this will result in double-translation of the text, looking at the other return values below.

plugins/paymethod/paypal/PaypalPaymentPlugin.php
```diff
@@ -176,6 +178,12 @@ public function isConfigured($context)
      */
     public function handle($args, $request)
```
asmecher, line 179: This will prevent responses from PayPal from being handled by OJS, but doesn't prevent OJS from creating transactions with PayPal. See also `PaypalPaymentForm.php` in the `display` function.

asmecher (changes requested):
A couple of things needing a look over!

config.TEMPLATE.inc.php
```diff
@@ -134,6 +134,11 @@
 ; 0, unvalidated accounts will never be removed. Use this setting to automatically remove bot registrations.
 user_validation_period = 28
 
+; Turn sandbox mode to On in order to prevent the software from interacting with outside systems.
+; Use this for development or testing purposes.
+: Affected services are email, datacite, crossref, orcid and paypal payment
```
asmecher, line 139: Probably best not to list affected services -- many of them are in plugins, so we can't list them all. But also, the leading character here is a `:`, which is not a comment.

asmecher:
Replaced by https://github.com/pkp/ojs/pull/4190

PR #4146, opened 2024-01-19
pkp/pkp-lib#699 Show locale in url in multilingual contexts

jyhein:
- Show the current locale in the url in multilingual contexts.
- Old links without the locale are redirected to include the current locale.
- Monolingual contexts do not have the locale

### DIFF
--- a/cypress/tests/integration/Pubmed.cy.js
+++ b/cypress/tests/integration/Pubmed.cy.js
@@ -18,7 +18,7 @@ describe('Pubmed tests', function () {
 		.then((win) => {
 			const csrfToken = win.pkp.currentUser.csrfToken;
 			cy.request({
-					url: '/index.php/publicknowledge/management/importexport/plugin/PubMedExportPlugin/exportSubmissions',
+					url: '/index.php/publicknowledge/en/management/importexport/plugin/PubMedExportPlugin/exportSubmissions',
 					method: 'POST',
 					headers: {
 						'X-Csrf-Token': csrfToken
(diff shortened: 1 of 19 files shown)


jyhein:
Hey @asmecher,

I have PRs ready for review of the code...

PKP: https://github.com/pkp/pkp-lib/pull/9628
OJS: https://github.com/pkp/ojs/pull/4146

asmecher:
My apologies for the delay, @jyhein! @bozana, would you mind reviewing this?

bozana:
Hi @jyhein, all well here.
There would be a few more places to consider the default article view URLs, like this:
`$request->getDispatcher()->url($request, Application::ROUTE_PAGE, $journal->getPath(), 'article', 'view', [$article->getBestId()], urlLocaleForPage: '');`
Z.B. 
- plugins/generic/googleScholar -- we consider the the article primary language for the metadata we provide here, so I suppose the URL should stay the same for now i.e. without language in the URL
- plugins/metadata/dc11/filter/
- plugins/oaiMetadataFormats/marc and marcxml
- eventually also plugins/generic/recommendBySimilarity and recommendByAuthor

What about Sitemap, how are there the Google rules? -- I believe I read somewhere that all languages needs to be listed there..

I will need to test it with the usage stats log file re-processing -- the older log files that contain the (very) old URLs, if this now all works correctly...

Hmmm... I am not sure if I missed something, but if I find it out I will let you know :-)

Thanks a lot!

PR #4147, opened 2024-01-19
pkp/pkp-lib#9627 Add indexes to usage stats temporary tables 

bozana:
s. https://github.com/pkp/pkp-lib/issues/9627

### DIFF
--- a/dbscripts/xml/upgrade.xml
+++ b/dbscripts/xml/upgrade.xml
@@ -215,6 +215,7 @@
 	<upgrade minversion="3.1.0.0" maxversion="3.4.0.9">
 		<migration class="PKP\migration\upgrade\v3_4_0\I9136_MigrateUniqueSiteId"/>
 		<migration class="APP\migration\upgrade\v3_4_0\I9231_FixMetricsIndexes"/>
+		<migration class="PKP\migration\upgrade\v3_4_0\I9627_AddUsageStatsTemporaryTablesIndexes"/>
 	</upgrade>
 
 	<!-- update plugin configuration - should be done as the final upgrade task -->
--- a/classes/migration/install/MetricsMigration.php
+++ b/classes/migration/install/MetricsMigration.php
@@ -269,7 +269,7 @@ class MetricsMigration extends \PKP\migration\Migration
             $table->bigIncrements('usage_stats_temp_total_id');
 
             $table->dateTime('date', $precision = 0);
-            $table->string('ip', 255);
+            $table->string('ip', 64);
             $table->string('user_agent', 255);
             $table->bigInteger('line_number');
             $table->string('canonical_url', 255);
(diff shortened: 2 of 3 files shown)


dbscripts/xml/upgrade.xml
```diff
@@ -215,6 +215,7 @@
 	<upgrade minversion="3.1.0.0" maxversion="3.4.0.9">
```
jonasraoni, line 215: Hmm, the latest release is `3.4.0.4`, and these migrations look like extra fixes for the `3.4`. Therefore, shouldn't the `maxversion` here be `3.4.0.3`?

jonasraoni: I see it this way:
- User upgrades to `3.4.0.4`
- These migrations will be executed
- User upgrades to `3.4.0.5` (next release), the current `maxversion` will still fit, and the migrations will probably be executed again (hopefully they can be re-executed without errors).

jonasraoni: Following the same line of thought I think this new migration should be placed on another `<upgrade>` entry, perhaps: `<upgrade minversion="3.1.0.0" maxversion="3.4.0.4">`

jonasraoni: As it's late here, tagging @asmecher to confirm 🤪

bozana: @asmecher said once -- if you remember we already have this discussion :-) -- that the user do not execute DB upgrade for minor versions, that's why we always consider the 9 at the end...

jonasraoni: I just did a test (upgraded my 3.4.0.4 to a fake 3.4.0.5) and the migrations were re-executed, so I think it's better to fix.

bozana: I will let @asmecher decide. 
From this upgrade group:
- I9136_MigrateUniqueSiteId was introduce in 3.4.0.1
- I9231_FixMetricsIndexes in 3.4.0.2
- I9627_AddUsageStatsTemporaryTablesIndexes (this one) will be in 3.4.0.5/3.4.1
It seems that all of them can be executed several times without problems, but...

bozana:
@jonasraoni, the second commit here is where I decrease the column IP to 64 characters.
If OK I will merge it all into one commit...

jonasraoni (approved):
I approved! There's just the `upgrade.xml` issue remaining, but I think it can be handled by a separated commit.

dbscripts/xml/upgrade.xml
jonasraoni: I think this can be moved to another issue as well.

jonasraoni: I've created this issue: https://github.com/pkp/pkp-lib/issues/9637 and added a milestone for 3.4.0-5 to not forget about it, so I think you can merge this :)

PR #4164, opened 2024-01-30
pkp/pkp-lib#9664 Use jobs chain for usage stats log files processing 

bozana:
s. https://github.com/pkp/pkp-lib/issues/9664

### DIFF
--- a/jobs/statistics/CompileCounterSubmissionDailyMetrics.php
+++ b/jobs/statistics/CompileCounterSubmissionDailyMetrics.php
@@ -30,17 +30,13 @@ class CompileCounterSubmissionDailyMetrics extends BaseJob
     public $tries = 1;
 
     /**
-     * The load ID = usage stats log file name
-     */
-    protected string $loadId;
-
-    /**
      * Create a new job instance.
+     *
+     * @param string $loadId Usage stats log file name
      */
-    public function __construct(string $loadId)
+    public function __construct(protected string $loadId)
     {
         parent::__construct();
-        $this->loadId = $loadId;
     }
 
     /**
--- a/jobs/statistics/CompileCounterSubmissionDailyMetrics.php
+++ b/jobs/statistics/CompileCounterSubmissionDailyMetrics.php
@@ -1,0 +1,60 @@
+<?php
+
+/**
+ * @file jobs/statistics/CompileCounterSubmissionDailyMetrics.php
+ *
+ * Copyright (c) 2024 Simon Fraser University
+ * Copyright (c) 2024 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class CompileCounterSubmissionDailyMetrics
+ *
+ * @ingroup jobs
+ *
+ * @brief Compile COUNTER submission daily metrics.
+ */
+
+namespace APP\jobs\statistics;
+
+use APP\statistics\TemporaryItemInvestigationsDAO;
+use APP\statistics\TemporaryItemRequestsDAO;
+use APP\statistics\TemporaryTotalsDAO;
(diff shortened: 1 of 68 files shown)


jobs/statistics/ProcessUsageStatsLogFile.php
```diff
@@ -0,0 +1,246 @@
...
+            }
+
+            // Avoid bots.
+            if (Core::isUserAgentBot($entryData->userAgent)) {
+                continue;
+            }
+
+            $this->insertTemporaryUsageStatsData($entryData, $lineNumber);
+        }
+        fclose($fhandle);
```
jonasraoni, line 127: Not a big deal, but it's safer to use:
```php
fopen();
try{
  ...
} finally {
  fclose();
}
```

You might also use the [SplFileObject](https://www.php.net/splfileobject).

jobs/statistics/ProcessUsageStatsLogFile.php
```diff
@@ -0,0 +1,246 @@
...
+        $temporaryItemInvestigationsDao->deleteByLoadId($this->loadId);
+        $temporaryItemRequestsDao->deleteByLoadId($this->loadId);
+    }
+
+    /**
+     * Validate the usage stats log entry
+     *
+     * @throws Exception.
+     */
+    protected function isLogEntryValid(object $entry): void
```
jonasraoni, line 151: As it's not returning a boolean, I think it's better to rename it to `validateLogEntry()` or something like that.

jobs/statistics/ProcessUsageStatsLogFile.php
```diff
@@ -0,0 +1,246 @@
...
+        $validFileTypes = [
+            StatisticsHelper::STATISTICS_FILE_TYPE_PDF,
+            StatisticsHelper::STATISTICS_FILE_TYPE_DOC,
+            StatisticsHelper::STATISTICS_FILE_TYPE_HTML,
+            StatisticsHelper::STATISTICS_FILE_TYPE_OTHER,
+        ];
+        if (!empty($entry->fileType) && !in_array($entry->fileType, $validFileTypes)) {
+            throw new Exception(__('admin.job.processLogFile.invalidLogEntry.fileType'));
+        }
+        if (!empty($entry->country) && (!ctype_alpha($entry->country) || !(strlen($entry->country) == 2))) {
```
jonasraoni, line 178: I think it's clearer `strlen($entry->country) !== 2`

jobs/statistics/ProcessUsageStatsLogFile.php
```diff
@@ -0,0 +1,246 @@
...
+            throw new Exception(__('admin.job.processLogFile.invalidLogEntry.country'));
+        }
+        if (!empty($entry->region) && (!ctype_alnum($entry->region) || !(strlen($entry->region) <= 3))) {
```
jonasraoni, line 181: And here `strlen($entry->region) > 3`

jonasraoni (approved):

jobs/statistics/CompileCounterSubmissionDailyMetrics.php
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * The load ID = usage stats log file name
+     */
+    protected string $loadId;
+
+    /**
+     * Create a new job instance.
+     */
+    public function __construct(string $loadId)
```
jonasraoni, line 40: You can use this new syntax to declare the properties that should be also set at the constructor:
```php
public function __construct(protected string $loadId) {
}
```

jobs/statistics/ProcessUsageStatsLogFile.php
bozana: I will then use SplFileObject...

PR #4224, opened 2024-03-24
pkp/pkp-lib#9829 Attempt to improve the search query performance

jonasraoni:


classes/search/ArticleSearchDAO.php
```diff
@@ -42,83 +46,65 @@ public function getPhraseResults($journal, $phrase, $publishedFrom = null, $publ
...
+            )
+            ->join('publications AS p', 'p.publication_id', '=', 's.current_publication_id')
+            ->join(
+                'publication_settings AS ps',
+                fn (JoinClause $j) => $j
+                    ->where('ps.setting_name', '=', 'issueId')
+                    ->whereColumn('ps.publication_id', '=', 'p.publication_id')
+                    ->where('ps.locale', '=', '')
+            )
+            ->join('issues AS i', 'i.issue_id', '=', DB::raw('CAST(ps.setting_value AS ' . (DB::connection() instanceof MySqlConnection ? 'UNSIGNED' : 'INTEGER') . ')'))
```
jonasraoni, line 66: The biggest improvement came from this modification.
I'll create another issue to move the `issue_id` to a physical field, I've seen this `CAST()` in several places.

asmecher:
@jonasraoni, I think on balance I'd rather leave this as an available modification for those who need it. As you've pointed out, the right fix is to address the casting issue, which is filed at [#10157](https://github.com/pkp/pkp-lib/issues/10157) (which is a breakout from [#7975](https://github.com/pkp/pkp-lib/discussions/7975)) -- or to adopt an external search index/toolset, for which currently support Lucene and will need to do some future work. Meanwhile, the search indexing toolset is in need of bigger work, and I'd rather not do that on a stable branch.

On the other hand, these changes might've been proven out already within hosting, which might help tip the balance for me.

PR #4236, opened 2024-04-03
Fix sequence casting error

kennydude:
This fixes an error when importing native XML using PHP 8 whereby it expects a strict integer and fails with this error otherwise:

```
1.Generic Items
- PKP\section\PKPSection::setSequence(): Argument #1 ($sequence) must be of type float, string given, called in /home/oiccpres/public_html/plugins/importexport/native/filter/NativeXmlIssueFilter.php on line 340
```


jonasraoni (approved):

plugins/importexport/native/filter/NativeXmlIssueFilter.php
```diff
@@ -337,7 +337,7 @@ public function parseSection($node)
         $section->setContextId($context->getId());
         $reviewFormId = $node->getAttribute('review_form_id');
         $section->setReviewFormId($reviewFormId ? (int) $reviewFormId : null);
-        $section->setSequence($node->getAttribute('seq'));
+        $section->setSequence((int) $node->getAttribute('seq'));
```
jonasraoni, line 340: This should be ok, but as the field is a float, it's better to use another typecast `(float)`.

PR #4241, opened 2024-04-11
pkp/pkp-lib#9813 upgrade script to fix frong submission progress values inserted by QuickSubmit plugin

bozana:
s. https://github.com/pkp/pkp-lib/issues/9813

### DIFF
--- a/dbscripts/xml/upgrade.xml
+++ b/dbscripts/xml/upgrade.xml
@@ -223,7 +223,11 @@
 	</upgrade>
 
 	<upgrade minversion="3.1.0.0" maxversion="3.4.0.5">
-		<migration class="PKP\migration\upgrade\v3_4_0\I9830_FixEmptyUserLocales" />
+		<migration class="PKP\migration\upgrade\v3_4_0\I9830_FixEmptyUserLocales"/>
+	</upgrade>
+
+	<upgrade minversion="3.4.0.0" maxversion="3.4.9.9">
+		<migration class="APP\migration\upgrade\v3_4_0\I9813_QuickSubmitSubmissionProgressType"/>
 	</upgrade>
 
 	<!-- update plugin configuration - should be done as the final upgrade task -->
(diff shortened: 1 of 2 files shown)


dbscripts/xml/upgrade.xml
```diff
@@ -223,7 +223,11 @@
 	</upgrade>
 
 	<upgrade minversion="3.1.0.0" maxversion="3.4.0.5">
-		<migration class="PKP\migration\upgrade\v3_4_0\I9830_FixEmptyUserLocales" />
+		<migration class="PKP\migration\upgrade\v3_4_0\I9830_FixEmptyUserLocales"/>
+	</upgrade>
+
+	<upgrade minversion="3.4.0.0" maxversion="3.4.0.6">
+		<migration class="PKP\migration\upgrade\v3_4_0\I9813_QuickSubmitSubmissionProgressType"/>
```
jonasraoni, line 230: As the current version is `3.4.0-5`, you can move this migration to the tag above.

jonasraoni (changes requested):
Just one fix and I think everything is ok here :)

dbscripts/xml/upgrade.xml
bozana: And it does not matter that also 3.1.x.x - 3.3.9.9 would run it too, after the script I7191_SubmissionProgressType has been run :thinking: -- not so clean but...
Or to actually use `<upgrade minversion="3.4.0.0" maxversion="3.4.0.5">`?

bozana: @jonasraoni, OK now?

bozana:
Hi @jonasraoni, I have just moved the upgrade script into the application (OJS) because the QuickSubmit is not supported for OMP and OPS in 3.4 release...
When tests pass I will then merge...

jonasraoni (approved):
That makes sense :)

PR #4257, opened 2024-04-28
pkp/pkp-lib#9895 app key and encryption service integration

touhidurabir:
for pkp/pkp-lib#9895

### DIFF
--- a/dbscripts/xml/upgrade.xml
+++ b/dbscripts/xml/upgrade.xml
@@ -13,7 +13,6 @@
 
 <install version="3.5.0.0">
 	<code function="checkPhpVersion" />
-	<code function="addAppKey" />
 
 	<upgrade minversion="2.0.0.0" maxversion="3.2.9.9">
 		<!-- OJS < 3.1.0 upgrade unsupported -->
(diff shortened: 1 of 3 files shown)


dbscripts/xml/upgrade.xml
```diff
@@ -13,6 +13,7 @@
 
 <install version="3.5.0.0">
 	<code function="checkPhpVersion" />
+	<code function="addAppKey" />
```
asmecher, line 16: This should be added as a migration, not a `<code ...>` addition in the `Upgrade` class.

asmecher (changes requested):
Just some small rearrangement, thanks!

dbscripts/xml/upgrade.xml
touhidurabir: @asmecher I understand it should part of upgrade and need to moved to upgrade class but I am confused about the migration part . I does not modify/update any DB tables but update the config.inc.php to add a unique app key if the `app_key` variable defined/set in the `config.inc.php` file . thats why it's added as `<code ...>` to run. or I am misunderstanding something ?

asmecher: The `<code ...>` pattern leads to accumulating cruft in the `Upgrade` class, so we're using Migration classes instead for new code. Migrations are good for whatever, not just database schema adaptations!

docs/release-notes/README-3.5.0
```diff
@@ -11,9 +11,15 @@ See config.TEMPLATE.inc.php for a description and examples of all supported
 configuration parameters.
 
 New config.inc.php parameters added for general:
-    - session_cookie_enctyption_key (default value: ''), allow cookie encryption when set
+    - app_key (default value: ''), application specific key will used internally for encryption/decryption
```
asmecher, line 14: Add a brief instruction for how to create this for those who are upgrading

asmecher (changes requested):
One suggestion for a change (please port to the other apps)!

PR #4272, opened 2024-05-08
pkp/pkp-lib#8826 Add missing FK and index on email_log.sender_id

asmecher:


asmecher:
(Will commit submodule manually.)

PR #4275, opened 2024-05-18
[OJS][main] #9840 Remove vue3-highlightjs and from highlight.js use only XML language

defstat:


package.json
```diff
@@ -38,8 +38,7 @@
     "vue": "^3.3.8",
     "vue-chartjs": "^5.2.0",
     "vue-draggable-plus": "^0.2.4",
-    "vue-scrollto": "^2.20.0",
-    "vue3-highlightjs": "^1.0.5"
+    "vue-scrollto": "^2.20.0"
   },
```
jardakotesovec, line 42: Its odd that there is no package-lock.json update as part of it. Its important to remove it via `npm uninstall` to also update package-lock.json. Can you check that?

defstat: I have not added the `package-lock.json` to the commit. I am generally not adding it to the commits when doing such changes. Should I?

jardakotesovec: @defstat package-lock.json is bit pain when conflict happen. But you want to keep that in sync with package.json

For example if you just remove it from package.json, and someone than runs npm install - it will create update package-lock.json, which is confusing.

And in general package-lock.json purpose is to ensure that everyone, including tests and deployments are all using exactly the same version in their dependencies. So if there is some change in dependencies - package-lock.json should get updated as well.

defstat:
Closed in favor of pkp/ojs#4281

PR #4279, opened 2024-05-21
pkp/pkp-lib#9972 Explicitely define folders with javascript to improv…

jardakotesovec:
…e VS code performance

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -2,8 +2,8 @@
 	"compilerOptions": {
 		"baseUrl": ".",
 		"paths": {
-			"@/components/*": ["lib/ui-library/src/components/*"]
+			"@/*": ["lib/ui-library/src/*"]
 		}
 	},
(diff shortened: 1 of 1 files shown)


jsconfig.json
```diff
@@ -2,8 +2,8 @@
 	"compilerOptions": {
 		"baseUrl": ".",
 		"paths": {
-			"@/components/*": ["lib/ui-library/src/components/*"]
+			"@/*": ["lib/ui-library/src/*"]
 		}
 	},
-	"exclude": ["node_modules"]
+	"include": ["lib/ui-library/src/**/*", "lib/pkp/js/**/*.js"]
```
ewhanson, line 8: Does anything else need to be included here, e.g. the top level `js` directory? This includes the `build.js` and the `pkp.min.js` but there are also a few legacy controller files. My guess is this is fine since this is all related to our legacy use of JS, but just wanted to double check.

PR #4283, opened 2024-05-21
pkp/pkp-lib#9976 Display Galley language in Galleys Grid

ajnyga:
https://github.com/pkp/pkp-lib/issues/9976

### DIFF
--- a/controllers/grid/articleGalleys/ArticleGalleyGridCellProvider.php
+++ b/controllers/grid/articleGalleys/ArticleGalleyGridCellProvider.php
@@ -22,6 +22,7 @@ use APP\submission\Submission;
 use PKP\controllers\api\file\linkAction\DownloadFileLinkAction;
 use PKP\controllers\grid\DataObjectGridCellProvider;
 use PKP\controllers\grid\GridHandler;
+use PKP\facades\Locale;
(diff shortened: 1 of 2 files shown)


controllers/grid/articleGalleys/ArticleGalleyGridCellProvider.php
```diff
@@ -64,6 +65,10 @@ public function getTemplateVarsFromRowColumn($row, $column)
                 return [
                     'label' => !$element->getData('urlRemote') && $element->getData('submissionFileId') ? '' : $element->getLabel()
                 ];
+            case 'language':
+                return [
+                    'label' => Locale::getSubmissionLocaleDisplayNames([$element->getLocale()])[$this->getLocale()]
```
bozana, line 70: I think this should be:
`Locale::getSubmissionLocaleDisplayNames([$element->getLocale()], $this->getLocale())[$element->getLocale()]`

The array returned from the function has submission languages as keys, not the UI languages.
E.g. if the UI language is fr_CA, then there will be a warning: Undefined array key "fr_CA".

ajnyga: ah true, thanks

PR #4305, opened 2024-06-10
Update scheduledTasks.xml

bdcaf:
day="0" here causes the usageStats to be only collected once per month (when executed via cron job, don't think it affects acron). I believe this is by mistake as it would leave the default statistics page (= last month) completely empty. 

See https://github.com/pkp/pkp-lib/blob/a60e4930128362be6e961b6014374d8eacc83b62/classes/scheduledTask/ScheduledTaskHelper.php#L116 - it compares with `(int)date('j')` - which always fails as the result is a number in range 1-31 (see https://www.php.net/manual/en/datetime.format.php), never 0 - so the job only triggers when `$passTimestamp` exceeds 1 month.

### DIFF
--- a/registry/scheduledTasks.xml
+++ b/registry/scheduledTasks.xml
@@ -57,7 +57,7 @@
 	</task>
 	<task class="APP\tasks\UsageStatsLoader">
 		<descr>Process the usage stats logs and load the numbers into the DB tables.</descr>
-		<frequency day="0"/>
+		<frequency day="*"/>
 	</task>
 	<task class="PKP\task\ProcessQueueJobs">
 		<descr>Process pending queue jobs in the default queue driver and queue</descr>
--- a/registry/scheduledTasks.xml
+++ b/registry/scheduledTasks.xml
@@ -57,7 +57,7 @@
 	</task>
 	<task class="APP\tasks\UsageStatsLoader">
 		<descr>Process the usage stats logs and load the numbers into the DB tables.</descr>
-		<frequency day="*"/>
+		<frequency hour="0"/>
 	</task>
 	<task class="PKP\task\ProcessQueueJobs">
 		<descr>Process pending queue jobs in the default queue driver and queue</descr>


asmecher:
@bozana, could you have a look at this? Thanks!

bozana:
Hi @bdcaf, oh, good catch!
However, I would like us to then use hour="0" -- as in other examples, the scheduled tasks that should be run daily. In that case this condition would apply: https://github.com/pkp/pkp-lib/blob/stable-3_4_0/classes/scheduledTask/ScheduledTaskHelper.php#L269.
Would it be maybe possible for you to test it with hour="0" and provide that PR?
Thanks a lot!

bdcaf:
sure. I'll let you know tomorrow.

bozana:
Thanks a lot!

bdcaf:
modified to hour, and looks good on my end. Also yesterdays logs were processed without me triggering anything. 
Processed meaning logs were moved from usageEventLogs to archives and journal and issue statistics updated to non-zero number (and for some reason article stats always only updates the day after - so currently still 0).

I guess that's as good as it gets for such a tiny change :)

bozana:
Thanks a lot @bdcaf!
Where exactly do you look at the journal, issue and article usage numbers (and always see for articles 0 for yesterday) -- so I can double check that...

jonasraoni (approved):

bdcaf:
I mean the graph on /stats/context/context /stats/issues/issues and /stats/publications/publications
btw - just saw one where issues stats also was still on 0. let me know if there's any data i can provide.

bozana:
Hi @bdcaf, I have taken a look and the SQL we use uses "between startDate and yesterday" which would always exclude "yesterday". Thus, actually you should always see 0 for all objects for yesterday.

bdcaf:
It's no major issue. just funny (and journal managers sometimes asked about) that the journal and issue statistics behave different than articles. maybe it would make sense to exclude yesterday from the graph.

PR #4312, opened 2024-06-12
add method to calculate seq value for new Journal entry in order to maintain order consistency

taslangraham:
For https://github.com/pkp/pkp-lib/issues/9191

### DIFF
--- a/classes/journal/JournalDAO.inc.php
+++ b/classes/journal/JournalDAO.inc.php
@@ -124,4 +124,16 @@ class JournalDAO extends ContextDAO {
 		}
 		return false;
 	}
+
+	/**
+	 * Determines the seq value to be assigned to a new Journal entry.
+	 * @return int The next sequence value for a new journal entry.
+	 */
+	public function getNextSeqValue()
+	{
+		$result = $this->retrieve('SELECT seq FROM journals ORDER BY seq DESC LIMIT 1;');
+
+		return $result->current()->seq ? $result->current()->seq + 1 : 1;
+	}
+
 }


asmecher:
@taslangraham, did you miss a `git add`? I see the new function, but not the call to use it.

asmecher:
...ah, I've already found it :) https://github.com/pkp/pkp-lib/pull/10047/files

asmecher:
I think you can generalize this function into `pkp-lib` by using `$this->tableName` (see `lib/pkp/classes/context/ContextDAO.php` for other examples).

jonasraoni:
@asmecher I think this approach is unsafe to be used as a generic sequencing solution, while we got a value, another concurrent action might be getting the same value, even something more encapsulated/atomic like the code below isn't safe:
```sql
INSERT INTO table (sequence) VALUES ((SELECT COALESCE(MAX(sequence), 0) + 1 FROM table))
```

Another solution, assuming that `sequence <= auto increment`, and more performatic (i.e. doesn't require sorting over a non-index column), must be to use something like `LAST_INSERT_ID()` (in PostgreSQL we would need to grab the value from the sequence).

But I see the current code doesn't have room for that, so a less fancy solution, but safer, might be doing two database operations, basically to insert, then update the sequence field with the inserted ID.

Without using fancy SQL stuff, another generic solution that I can imagine is to emulate a generator and reserve values:
- Table `sequences (string name, int value)`
- Select the value with a **lock for update** `SELECT value FROM sequences WHERE name = 'xxx'`
- Increment the value
- Release the lock

Also, adding a unique index sounds ok, will avoid duplicated sequences and improve the `order by` performance.

taslangraham:
Closing tihs PR. Work for the issue will be done on https://github.com/pkp/pkp-lib/pull/10047

PR #4321, opened 2024-06-17
Bugfix stable 3 4 0 8592 fix notification subscription settings integrity

jonasraoni:


dbscripts/xml/upgrade.xml
```diff
@@ -226,8 +226,9 @@
 		<migration class="PKP\migration\upgrade\v3_4_0\I9830_FixEmptyUserLocales"/>
 	</upgrade>
 
-	<upgrade minversion="3.4.0.0" maxversion="3.4.9.9">
+	<upgrade minversion="3.4.0.0" maxversion="3.4.0.5">
```
jonasraoni, line 229: Migrations don't need to be executed on every upgrade.

bozana: Hi @jonasraoni, I think this QuickSubmit upgrade script should be called till the preflight check is not implemented -- because the users could continue to use the old plugin that inserts the wrong values in the DB.
S. https://github.com/pkp/pkp-lib/pull/9848#issuecomment-2058918104.

jonasraoni: That makes sense, I'll revert and add a comment.

touhidurabir (approved):

PR #4332, opened 2024-06-21
pkp/pkp-lib#9678 Replacing task schedular with laravel scheduler

touhidurabir:
for pkp/pkp-lib#9678


tools/runScheduledTasks.php
```diff
@@ -12,13 +12,15 @@
...
-class runScheduledTasks extends \PKP\cliTool\ScheduledTaskTool
-{
-}
+// We need to push the appropriate option to the command line tool and in this case
+// as we intend to run the schedule tasks, it's `run`
+array_push($argv, 'run');
 
-$tool = new runScheduledTasks($argv ?? []);
-$tool->execute();
+require(BASE_SYS_DIR . '/lib/pkp/tools/scheduler.php');
```
touhidurabir, line 26: wondering should we remove this file also as don't see any reason to just to provider another way when we have better command `php lib/pkp/tools/scheduler.php run` .

classes/scheduler/Scheduler.php
```diff
@@ -0,0 +1,52 @@
...
+            ->withoutOverlapping();
+
+        $this
+            ->schedule
+            ->call(fn () => (new OpenAccessNotification())->execute())
+            ->hourly()
+            ->name(OpenAccessNotification::class)
+            ->withoutOverlapping();
+
+        parent::registerSchedules();
```
jonasraoni, line 50: I think the base tasks should be added first, but not a big deal.

config.TEMPLATE.inc.php
```diff
@@ -577,6 +572,38 @@
...
+
+[schedule]
+
+; Whether or not to turn on the built-in schedule task runner
+;
+; When enabled, schedule tasks will be processed at the end of each web
+; request to the application.
+;
+; Use of the built-in schedule task runner is highly discouraged for high-volume 
+; sites. Instead, system's crontab should be configured to run every minutes as
```
jonasraoni, line 587:
```suggestion
; sites. Use your operational system's task scheduler instead, and configure it to run the task scheduler every minute.
```

config.TEMPLATE.inc.php
```diff
@@ -577,6 +572,38 @@
...
+[schedule]
+
+; Whether or not to turn on the built-in schedule task runner
+;
+; When enabled, schedule tasks will be processed at the end of each web
+; request to the application.
+;
+; Use of the built-in schedule task runner is highly discouraged for high-volume 
+; sites. Instead, system's crontab should be configured to run every minutes as 
+; * * * * * cd /path-to-your-project && php lib/pkp/tools/scheduler.php run >> /dev/null 2>&1
```
jonasraoni, line 588: I removed the `cd`, because the commands are expected to know where they are running.

```suggestion
; Sample for the *nix crontab
; * * * * * php lib/pkp/tools/scheduler.php run >> /dev/null 2>&1
```

config.TEMPLATE.inc.php
```diff
@@ -577,6 +572,38 @@
...
+;
+; See: <link-to-documentation>
+task_runner = On
+
+; How often should be built in schedule task runner should run schedule tasks at the
```
jonasraoni, line 593:
```suggestion
; How often should the built-in schedule task runner run scheduled tasks at the
```

config.TEMPLATE.inc.php
```diff
@@ -577,6 +572,38 @@
...
+; end of web request life cycle defined the seconds.
```
jonasraoni, line 594:
```suggestion
; end of web request life cycle (value defined the seconds).
```

config.TEMPLATE.inc.php
```diff
@@ -577,6 +572,38 @@
...
+; 
+; This configuration will only have impact for build in task runner, not to the system
```
jonasraoni, line 596:
```suggestion
; This configuration will only have effect for the build-it task runner, it doesn't apply to the system
```

config.TEMPLATE.inc.php
```diff
@@ -577,6 +572,38 @@
...
+; crontab configuration. 
+;
+; The default value set to 60 seconds and it is recommened not to to set any value less
+; that it to avoid making possibility of application beign slow.
```
jonasraoni, line 600:
```suggestion
; The default value is set to 60 seconds, a value smaller than that might affect the application performance negatively.
```

config.TEMPLATE.inc.php
```diff
@@ -577,6 +572,38 @@
...
+task_runner_interval = 60
+
+; This allow to sent the schedule task result notification mail only when there is an
+; error has occurred.
```
jonasraoni, line 604:
```suggestion
; When enabled, an email with the scheduled task result will be sent only when an error has occurred. Otherwise, all tasks will generate a notification.
```

tools/runScheduledTasks.php
jonasraoni: I think it's ok, but it will be less disruptive for sysadmins if we keep it this way.

jonasraoni (changes requested):
Just left minor suggestions.

tools/runScheduledTasks.php
touhidurabir: As this is already a breaking change, better to clean up all the deprecated functionalities . I have removed it .

config.TEMPLATE.inc.php
```diff
@@ -575,6 +570,40 @@
...
+; sites. Use your operational system's task scheduler instead, and configure 
+; it to run the task scheduler every minute.
+; Sample for the *nix crontab
+; * * * * * php lib/pkp/tools/scheduler.php run >> /dev/null 2>&1
+;
+; See: <link-to-documentation>
+task_runner = On
+
+; How often should the built-in schedule task runner run scheduled tasks at the
+; end of web request life cycle (value defined the seconds).
```
jonasraoni, line 594: My suggestion was still incorrect 😁

```suggestion
; end of web request life cycle (value defined in seconds).
```

jonasraoni (approved):

tools/runScheduledTasks.php
jonasraoni: @mfelczak FYI, the `tools/runScheduledTasks.php` will cease to exist on 3.5.

PR #4349, opened 2024-07-03
Update record.tpl  (change 786 marc field by 773 and subfields )

horaciod:
according to official documentation of the marc bibliographic manual by loc.gov  The journal title and volume  must to be in 773 fields with subfield "t" and "g".  the 786 field is use for other purposes . 
This is important for journals imports by tools like Vufind.


asmecher:
@bozana, could you have a look at this one as well?

PR #4350, opened 2024-07-03
Update marc record.tpl (change 786 marc field by 773 and subfields)  

horaciod:
according to official documentation of the marc bibliographic manual by loc.gov  The journal title and volume  must to be in 773 fields with subfield "t" and "g".  the 786 field is use for other purposes . 
This is important for journals imports by tools like Vufind.


asmecher:
@bozana, could you have a look at this one?

bozana:
@asmecher, do you think it is enough to have this change only in main, or shell we back port it to our stable branches as well?

asmecher:
@bozana, I've chatted with @kaitlinnewson, and we both think there's a risk of disruption for existing users if we make this change on a point release. I think it's best to leave to the `main` branch and next major release, but anyone wanting to can patch their installs. Thanks! (Also to @horaciod!)

horaciod:
Thank you all.
I have a meeting with my librarians and catalogers to do a complete review of the tpls.

PR #4359, opened 2024-07-10
pkp/pkp-lib#10132 move plugin validation locale key for use in omp

kaitlinnewson:


bozana:
@kaitlinnewson, it seems there are some conflicts with the pkp-lib submodule.
Could you please rebase again? (e.g. git fetch official, git rebase -i official/stable-3_4_0, where official should be your name for the pkp repository). First pkp-lib, then forced push it to your repository.
In OJS you would need to submit the lib/pkp submodule in an extra commit, as the last commit. At the end of commit name you would need to define your repository and branch like this: `##kaitlinnewson/10132-3_4_0##`. See as an example: https://github.com/pkp/ojs/pull/4344/commits.
Thus, you would need to first reset this commit (locally it is your last commit) (e.g. `git reset HEAD~1`), then you would need to add only the changes from OJS `locale` folder, and then commit them. Then rebase with the latest changes from the pkp OJS repository. Then you can use `git add lib/pkp`, and commit with the name like said above. Use forced push to push it all then to your GitHub repository.
Let me know if you would need any help, eventually we can do it also together...

kaitlinnewson:
Thanks @bozana, that was very helpful - I think I've resolved things now here and in main (no conflicts showing in 3_3_0), can you take another look?

PR #4367, opened 2024-07-16
pkp/pkp-lib#9892 upgrade to laravel 11 and PHP 8.2+

touhidurabir:
pkp/pkp-lib#9892

### DIFF
--- a/classes/migration/install/JournalsMigration.php
+++ b/classes/migration/install/JournalsMigration.php
@@ -29,7 +29,7 @@ class JournalsMigration extends \PKP\migration\Migration
             $table->comment('A list of all journals in the installation of OJS.');
             $table->bigInteger('journal_id')->autoIncrement();
             $table->string('path', 32);
-            $table->float('seq', 8, 2)->default(0)->comment('Used to order lists of journals');
+            $table->float('seq', 53)->default(0)->comment('Used to order lists of journals');
             $table->string('primary_locale', 28);
             $table->smallInteger('enabled')->default(1)->comment('Controls whether or not the journal is considered "live" and will appear on the website. (Note that disabled journals may still be accessible, but only if the user knows the URL.)');
             $table->unique(['path'], 'journals_path');
--- a/classes/migration/install/JournalsMigration.php
+++ b/classes/migration/install/JournalsMigration.php
@@ -29,7 +29,7 @@ class JournalsMigration extends \PKP\migration\Migration
             $table->comment('A list of all journals in the installation of OJS.');
             $table->bigInteger('journal_id')->autoIncrement();
             $table->string('path', 32);
-            $table->float('seq', 53)->default(0)->comment('Used to order lists of journals');
+            $table->float('seq')->default(0)->comment('Used to order lists of journals');
             $table->string('primary_locale', 28);
             $table->smallInteger('enabled')->default(1)->comment('Controls whether or not the journal is considered "live" and will appear on the website. (Note that disabled journals may still be accessible, but only if the user knows the URL.)');
             $table->unique(['path'], 'journals_path');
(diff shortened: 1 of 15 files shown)


classes/migration/install/JournalsMigration.php
```diff
@@ -29,7 +29,7 @@ public function up(): void
             $table->comment('A list of all journals in the installation of OJS.');
             $table->bigInteger('journal_id')->autoIncrement();
             $table->string('path', 32);
-            $table->float('seq', 8, 2)->default(0)->comment('Used to order lists of journals');
```
asmecher, line 32: [Motivation for this change here](https://laravel.com/docs/11.x/upgrade#floating-point-types). I'd rather just go to `->float('seq')` rather than specifying the magic number `53` -- at some point we can convert these to integers, but let's not do it in this PR.

classes/migration/install/OJSMigration.php
```diff
@@ -318,14 +324,14 @@ public function up(): void
             $table->foreign('journal_id', 'subscription_types_journal_id')->references('journal_id')->on('journals')->onDelete('cascade');
             $table->index(['journal_id'], 'subscription_types_journal_id');
 
-            $table->float('cost', 8, 2);
+            $table->float('cost', 53);
```
asmecher, line 327: The `cost` column should definitely be converted over to a `decimal` now that we're losing the precision and scale.

classes/migration/install/OJSMigration.php
```diff
@@ -415,7 +421,7 @@ public function up(): void
             $table->index(['user_id'], 'completed_payments_user_id');
 
             $table->bigInteger('assoc_id')->nullable();
-            $table->float('amount', 8, 2);
+            $table->float('amount', 53);
```
asmecher, line 424: As with https://github.com/pkp/ojs/pull/4367/files#diff-4da55738934ca5e2abc12dd7c73319687c83ec86d36b1e60b8145d2d79ff7c1eR327.

asmecher (changes requested):
Thanks, I've suggested a few changes!

classes/migration/install/OJSMigration.php
touhidurabir: Right . In that case, we should have a migration also, added .

dbscripts/xml/upgrade.xml
```diff
@@ -68,6 +68,7 @@
 		<migration class="PKP\migration\upgrade\v3_4_0\I2890_AddSetNullForOnDeleteToReviewRoundIdForeign"/>
 		<migration class="APP\migration\upgrade\v3_4_0\I7901_Duplicate_OAI_IDs"/>
 		<migration class="PKP\migration\upgrade\v3_4_0\I7366_UpdateUserAPIKeySettings"/>
+		<migration class="PKP\migration\upgrade\v3_4_0\I7463_LocaleColumn"/>
```
touhidurabir, line 71: weirdly this migration was missing in the upgrade list and causing to fail the upgrade .

dbscripts/xml/upgrade.xml
```diff
@@ -128,6 +129,7 @@
 	<upgrade minversion="3.1.0.0" maxversion="3.4.9.9">
 		<migration class="PKP\migration\upgrade\v3_5_0\PreflightCheckMigration" fallback="3.4.9.9" />
 		<migration class="APP\migration\upgrade\v3_5_0\I8333_AddMissingForeignKeys" />
+		<migration class="APP\migration\upgrade\v3_5_0\I9892_FloatToDecimalColumnTypeUpdate" />
```
touhidurabir, line 132: see https://github.com/pkp/ojs/pull/4367#discussion_r1707686359

dbscripts/xml/upgrade.xml
asmecher: Filed an issue related to this: https://github.com/pkp/pkp-lib/issues/10293

asmecher (approved):

PR #4368, opened 2024-07-16
pkp/pkp-lib#9910 Use UI Language in Announcement Feeds

bozana:
s. https://github.com/pkp/pkp-lib/issues/9910

### DIFF
--- a/plugins/generic/announcementFeed/AnnouncementFeedGatewayPlugin.php
+++ b/plugins/generic/announcementFeed/AnnouncementFeedGatewayPlugin.php
@@ -19,6 +19,7 @@ use APP\template\TemplateManager;
 use Carbon\Carbon;
 use PKP\announcement\Announcement;
 use PKP\db\DAORegistry;
+use PKP\facades\Locale;
 use PKP\plugins\GatewayPlugin;
 use PKP\site\VersionDAO;
 
@@ -161,6 +162,7 @@ class AnnouncementFeedGatewayPlugin extends GatewayPlugin
             'dateUpdated' => $dateUpdated,
             'announcements' => $announcements,
             'journal' => $journal,
+            'language' => str_replace(['_', '@'], '-', Locale::getLocale()),
         ]);
 
         $templateMgr->setHeaders(['content-type: ' . $mimeTypeMap[$type] . '; charset=utf-8']);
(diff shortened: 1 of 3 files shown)


jonasraoni (changes requested):
This is very similar to the https://github.com/pkp/webFeed/pull/3, so the same comment can be applied here.

plugins/generic/announcementFeed/AnnouncementFeedGatewayPlugin.php
```diff
@@ -161,6 +162,7 @@ public function fetch($args, $request)
             'dateUpdated' => $dateUpdated,
             'announcements' => $announcements,
             'journal' => $journal,
+            'language' => str_replace(['_', '@'], '-', Locale::getLocale()),
```
jonasraoni, line 165: I've seen this in other places, I think it should be a method somewhere ` toBcp47()`?!

plugins/generic/announcementFeed/templates/rss.tpl
```diff
@@ -28,7 +28,7 @@
 
 		{* optional elements *}
 		{if $journal->getPrimaryLocale()}
-		<dc:language>{$journal->getPrimaryLocale()|replace:'_':'-'|strip|escape:"html"}</dc:language>
+		<dc:language>{$language|escape}</dc:language>
```
jonasraoni, line 31: As the `$language` doesn't depend on the `$context->getPrimaryLocale()` (which I guess should be always available anyway), then the `{if}` can be removed.

plugins/generic/announcementFeed/templates/rss2.tpl
```diff
@@ -23,7 +23,7 @@
 
 		{* optional elements *}
 		{if $journal->getPrimaryLocale()}
-			<language>{$journal->getPrimaryLocale()|replace:'_':'-'|strip|escape:"html"}</language>
+			<language>{$language|escape}</language>
```
jonasraoni, line 26: As the `$language` doesn't depend on the `$context->getPrimaryLocale()` (which I guess should be always available anyway), then the `{if}` can be removed.

jonasraoni (changes requested):

plugins/generic/announcementFeed/AnnouncementFeedGatewayPlugin.php
bozana: s. https://github.com/pkp/pkp-lib/issues/10795

PR #4376, opened 2024-07-26
pkp/pkp-lib#9813 cast int to string

bozana:
s. https://forum.pkp.sfu.ca/t/error-during-upgrade-ojs-3-4-0-5-to-3-4-0-6-invalid-datetime-format-1292-truncated-incorrect-double-value-files/89623


jonasraoni (approved):

bozana:
Thanks a lot @jonasraoni, I will merge when the tests pass...

PR #4378, opened 2024-07-26
pkp/pkp-lib#9744 [OJS] Migrate older Tables to newer Table API

blesildaramirez:


js/load.js
```diff
@@ -34,9 +34,9 @@ import StatsIssuesPage from '@/components/Container/StatsIssuesPage.vue';
 import StatsUsersPage from '@/components/Container/StatsUsersPage.vue';
 import SubmissionWizardPage from '@/components/Container/SubmissionWizardPage.vue';
 import WorkflowPage from '@/components/Container/WorkflowPageOJS.vue';
-import JobsPage from '@/components/Container/JobsPage.vue';
-import FailedJobsPage from '@/components/Container/FailedJobsPage.vue';
-import FailedJobDetailsPage from '@/components/Container/FailedJobDetailsPage.vue';
+import JobsPage from '@/pages/jobs/JobsPage.vue';
+import FailedJobsPage from '@/pages/jobs/FailedJobsPage.vue';
+import FailedJobDetailsPage from '@/pages/jobs/FailedJobDetailsPage.vue';
 import DashboardPage from '@/pages/dashboard/DashboardPage.vue';
```
jardakotesovec, line 40: I think, these should not be necessary because there are now imported via Page.vue?

templates/stats/issues.tpl
```diff
@@ -166,37 +165,37 @@
...
+								<pkp-table-cell>{{ row.tocViews }}</pkp-table-cell>
+								<pkp-table-cell>{{ row.issueGalleyViews }}</pkp-table-cell>
+								<pkp-table-cell>{{ row.totalViews }}</pkp-table-cell>
+							</pkp-table-row>
 							<template #no-content v-if="!items.length">
-								<table-row class="pkpStats__noRecords">
-									<table-cell :colspan="tableColumns.length" class="!py-8 !px-4 !text-center">
+								<pkp-table-row class="pkpStats__noRecords">
+									<pkp-table-cell :colspan="tableColumns.length" class="!py-8 !px-4 !text-center">
 										<template v-if="isLoadingItems">
```
jardakotesovec, line 188: Ops..  here you can just use table-cell.

Reason why we expose the components with pkp prefix globally is so plugins can use them.

On our own pages&components we always import and register them locally. So for stats you are adding table-cell via 'components:'..   therefore no need to prefix that.

Also the no-content thingy is already supposed to insert correct row and cell with correct colspan. So you should just put there the spinner. But if its not working as expected let me know.

jardakotesovec (changes requested):

jardakotesovec (approved):

PR #4379, opened 2024-07-28
Bugfix stable 3 3 0 10249 fix data loss on settings

jonasraoni:


dbscripts/xml/upgrade.xml
```diff
@@ -208,6 +208,10 @@
 		<note file="docs/release-notes/README-3.3.0" />
 	</upgrade>
 
+	<upgrade minversion="3.3.0.16" maxversion="3.3.0.19">
```
asmecher, line 211: Since this will presumably be included in 3.3.0-19, shouldn't this be `maxversion="3.3.0.18"`?

classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
```diff
@@ -162,17 +162,37 @@ private function _toJSON($row, $tableName, $searchBy, $valueToConvert)
...
 
-		// Remove empty filters
-		$searchBy = array_filter($searchBy, function ($item) use ($row) {
-			if (empty($row->{$item})) return false;
-			return true;
-		});
+		if (!count($primaryKeys)) {
+			foreach (array_keys($row) as $column) {
+				if (substr($column, -3, '_id')) {
+					$primaryKeys[] = $column;
```
asmecher, line 182: I think this is an incorrect assumption e.g. for [`review_form_responses`](https://github.com/pkp/ojs/blob/8517ea2ddd6a88ed713067b81824a2495549df45/classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php#L127).

asmecher (changes requested):
Thanks, @jonasraoni! Added a couple comments.

dbscripts/xml/upgrade.xml
jonasraoni: Yes!

jonasraoni, line 211:
```suggestion
	<upgrade minversion="3.3.0.16" maxversion="3.3.0.18">
```

classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
jonasraoni: This part of the code shouldn't be executed in general, I left it as a last resort. The other option would be to compare all the fields except the `setting_value` to find a unique row.

The `review_form_responses` is an exception that will end up here. It doesn't have a unique/primary key, but I think the definition is incorrect (it should probably be using `$table->unique()`) and needs another PR:

https://github.com/pkp/pkp-lib/blob/3704af0dc5ce4454c3f0dbddde2a9734cb1351c5/classes/migration/install/ReviewsMigration.php#L119

Anyway, after passing through the `foreach`, it should end up with the expected columns (review_form_element_id + review_id), which was the behavior of the previous code (first column of the result set + review_id).

asmecher: OK, that makes sense, and I think you're right that the index should be unique. Could you...
- Go ahead with this fix, and
- File the need to change the index to `unique` as a new issue against 3.5.0?

classes/migration/upgrade/OJSv3_3_0UpgradeMigration.inc.php
```diff
@@ -162,17 +162,37 @@ private function _toJSON($row, $tableName, $searchBy, $valueToConvert)
 		if (is_array($oldValue) && $this->_isNumerical($oldValue)) $oldValue = array_values($oldValue);
 		$newValue = json_encode($oldValue, JSON_UNESCAPED_UNICODE); // don't convert utf-8 characters to unicode escaped code
 
-		$id = array_key_first((array)$row); // get first/primary key column
+		// Ensure ID fields are included on the filter to avoid updating similar rows
+		$tableDetails = Capsule::connection()->getDoctrineSchemaManager()->listTableDetails($tableName);
+		$primaryKeys = [];
+		try {
+			$primaryKeys = $tableDetails->getPrimaryKeyColumns();
+		} catch(Exception $e) {
```
jonasraoni, line 170:
```suggestion
		} catch (Exception $e) {
```

PR #4381, opened 2024-07-29
pkp/pkp-lib#9813 cast int to string

bozana:
s. https://forum.pkp.sfu.ca/t/error-during-upgrade-ojs-3-4-0-5-to-3-4-0-6-invalid-datetime-format-1292-truncated-incorrect-double-value-files/89623


asmecher:
@bozana, could you create an issue for this? Does it need back-porting to e.g. `stable-3_4_0`?

bozana:
Hi @asmecher, I added the PRs to the old/original issue (https://github.com/pkp/pkp-lib/issues/9813) but yes, you are right, it would be better to maybe have a new issue assigned for the next release. Here is the new issue: https://github.com/pkp/pkp-lib/issues/10270.
The first fix/PR was done for stable-3_4_0 and then ported to main. Everything is merged.

PR #4390, opened 2024-08-07
pkp/pkp-lib#8564 order export by issue ordering in pubmed plugin

kaitlinnewson:


plugins/importexport/pubmed/PubMedExportPlugin.php
```diff
@@ -191,14 +191,22 @@ public function exportIssues($issueIds, $context, $user)
...
-            ->filterByContextIds([$context->getId()])
-            ->filterByIssueIds($issueIds)
-            ->getMany();
+        $input = [];
+
+        foreach ($issueIds as $issueId) {
+            $sections = Repo::section()->getByIssueId($issueId);
+            foreach ($sections as $section) {
+                $submissionsInSections = [];
+                $submissionsInSections[] = Repo::submission()->getInSections($issueId, $context->getId());
```
bozana, line 200: I think either you can move this line out of/before `foreach ($sections...` or 
you can use 
```
$this->getCollector()
  ->filterByContextIds([$contextId])
  ->filterByIssueIds([$issueId])
  ->filterBySectionIds([$section->getId()])
  ->orderBy(Collector::ORDERBY_SEQUENCE, Collector::ORDER_DIR_ASC)
  ->getMany();
```

plugins/importexport/pubmed/PubMedExportPlugin.php
```diff
@@ -191,14 +191,22 @@ public function exportIssues($issueIds, $context, $user)
...
+                foreach ($submissionsInSections as $articles) {
+                    foreach ($articles[$section->getId()]['articles'] as $article) {
+                        $input[] = $article;
```
bozana, line 203: If you have `$submissionsInSections = Repo::submission()->getInSections($issueId, $context->getId());`  as said above then maybe you can only use
`$input = array_merge($input, $submissionsInSections[$section->getId()]);`

plugins/importexport/pubmed/PubMedExportPlugin.php
kaitlinnewson: I noticed that if I don't have `$submissionsInSections = [];` at the beginning to reset the array, then the export will have duplicated articles (because it still has data from the previous issue in the array)

bozana: maybe you can use `unset($submissionsInSections)` at the end of the loop?

PR #4405, opened 2024-08-21
pkp/pkp-lib#9468 submodues updates ##bozana/9468-3_4_0##

bozana:
s. https://github.com/pkp/pkp-lib/issues/9468

only submodule update


bozana:
https://data.crossref.org/schemas/crossref5.3.1.xsd seems to be unavailable :-\

PR #4410, opened 2024-08-26
pkp/pkp-lib#10263 Relax editing metadata on published/posted materials

Hafsa-Naeem:
for [#10263](https://github.com/pkp/pkp-lib/issues/10263)


defstat (changes requested):
@Hafsa-Naeem thanks for the changes! this could be a tricky feature, involving many code parts. 

I am trying the change on my local installation and it seems that I can't hit "save" button when the publication is published.

if you see at the picture bellow, the "Save" button is "disabled".

image

Also are there PRs for OMP? 

Lastly, I can see that the tests are failing. Is that ready for the tests also?

defstat (changes requested):
Thanks @Hafsa-Naeem! 3 notes on that:

1. I suppose that all metadata should be editable, including Contributors, Galleys etc. If that is true, have you considered them for this fix?

2. As asked [here](https://github.com/pkp/ojs/pull/4410#pullrequestreview-2577988967), shouldn't OMP, OPS be also considered?

3. In the current tests there are the older version's text (like `This version has been published and can not be edited.`) for published Versions, and perhaps the older version's restrictions (like disabled fields, buttons not showing etc). You also need to consider those.

locale/en/submission.po
```diff
@@ -252,7 +252,7 @@ msgid "publication.datePublished"
 msgstr "Date Published"
 
 msgid "publication.editDisabled"
```
asmecher, line 254: This locale key is also used for presentation to the author, where editing will still be disabled in some cases, so this key should be left as-is and a new locale key created for the editor case.

asmecher (changes requested):

asmecher (commented):
Hi @Hafsa-Naeem! A partial review with a few comments.

cypress/tests/data/60-content/AmwandengaSubmission.cy.js
```diff
@@ -459,22 +459,22 @@ describe('Data suite: Amwandenga', function() {
 		cy.login('dbarnes');
 		cy.visit('/index.php/publicknowledge/workflow/access/' + submission.id);
 		cy.openWorkflowMenu('Title & Abstract')
-		cy.get('button').contains('Save').should('be.disabled');
+		// cy.get('button').contains('Save').should('be.disabled');
```
asmecher, line 462: This should just be removed rather than commented.

cypress/tests/data/60-content/AmwandengaSubmission.cy.js
```diff
@@ -489,7 +489,6 @@ describe('Data suite: Amwandenga', function() {
 
 		cy.get('div[role="dialog"] [name="familyName-en"]').type(' Version 2', {delay: 0});
 		cy.get('div[role="dialog"] button').contains('Save').click();
-		// cy.get('#contributors button').contains('Save').should("not.be.visible");
```
asmecher, line 492: This should be removed entirely.

cypress/tests/data/60-content/AmwandengaSubmission.cy.js
```diff
@@ -560,12 +559,15 @@ describe('Data suite: Amwandenga', function() {
 		cy.clickStageParticipantButton('Stephanie Berardo', 'Login As');
 		cy.get('button').contains('OK').click();
 		cy.openWorkflowMenu('Title & Abstract')
+		cy.wait(1000); // let the UI load fully
+	  
 		cy.get('button:contains("Publish")').should('not.exist');
 		cy.get('button:contains("Create Version")').should('not.exist');
-		cy.get('button').contains('All Versions').click();
-		cy.get('button').contains('Version 1').click();
-		cy.contains('This version has been published and can not be edited.');
```
asmecher, line 567: I'm not sure if the new warning text appears here -- but if it does, there should be a Cypress check for it.

cypress/tests/data/60-content/VkarbasizaedSubmission.cy.js
```diff
@@ -135,11 +135,10 @@ describe('Data suite tests', function() {
 		cy.get('[data-cy="sidemodal-header"]').contains("Published");
 		// check publication status
 		cy.get('[data-cy="workflow-controls-left"]').contains("Published");
-		cy.contains('This version has been published and can not be edited.');
```
asmecher, line 138: I'm not sure if the new warning text appears here -- but if it does, there should be a Cypress check for it.

locale/fa/submission.po
```diff
@@ -252,6 +252,8 @@ msgstr "تاریخ منتشر شده"
 msgid "publication.editDisabled"
 msgstr "این نسخه منتشر شده و نمی‌تواند ویرایش شود."
 
+
```
asmecher, line 255: @Hafsa-Naeem, all of these unnecessary edits should be removed from the PR.

PR #4411, opened 2024-08-27
pkp/pkp-lib#10306 Queue jobs unit tests

touhidurabir:
for pkp/pkp-lib#10306

### DIFF
--- a/classes/decision/types/traits/RequestPayment.php
+++ b/classes/decision/types/traits/RequestPayment.php
@@ -78,7 +78,6 @@ trait RequestPayment
         $authorIds = $this->getAssignedAuthorIds($submission);
         foreach ($authorIds as $authorId) {
             $notificationMgr->createNotification(
-                Application::get()->getRequest(),
                 $authorId,
                 Notification::NOTIFICATION_TYPE_PAYMENT_REQUIRED,
                 $context->getId(),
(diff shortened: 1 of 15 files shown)


jobs/notifications/IssuePublishedNotifyUsers.php
```diff
@@ -99,7 +99,7 @@ protected function createMailable(
         Issue $issue,
         User $recipient,
         EmailTemplate $template,
-        Notification $notification
+        Notification|\PKP\notification\Notification $notification
```
asmecher, line 102: I think there's a missing `use` statement here, not another class named `Notification`.

asmecher (changes requested):
Just one tiny change, thanks!

jobs/notifications/IssuePublishedNotifyUsers.php
touhidurabir: Kind of but not exactly , we need to have both the `APP\notification\Notification` and `PKP\notification\Notification` as APP Notification has few const and when type hinting in method, we need PKP Notification . updated it .

asmecher: That's better, thanks!

classes/decision/types/traits/RequestPayment.php
```diff
@@ -78,7 +78,6 @@ protected function requestPayment(Submission $submission, User $editor, Context
         $authorIds = $this->getAssignedAuthorIds($submission);
         foreach ($authorIds as $authorId) {
             $notificationMgr->createNotification(
-                Application::get()->getRequest(),
```
asmecher, line 81: @touhidurabir, are you sure about this change?

asmecher: Ah, I see this is removed further up the stack because it wasn't used. Never mind.

PR #4424, opened 2024-09-04
Update NativeXmlIssueFilter.php

klausru:
I had the error "must be of type int, string given" in line 347. By specifying with (int) the importer runs fine.


asmecher:
@klausru, the change you propose would always force an integer value to be set for that column; the column is also nullable, and should be null when the import attribute is not specified.

klausru:
Actually, I tested with `NULL` in `abstract_word_count` and it seems that the specification "(int)" is required as my import still runs fine

PR #4437, opened 2024-09-12
pkp/pkp-lib#10388 Thank Reviewer action crashes on main

ewhanson:


jobs/orcid/DepositOrcidReview.php
```diff
@@ -19,55 +19,63 @@
...
     {
+        $submission = Repo::submission()->get($this->submissionId);
+        $context = Application::getContextDAO()->getById($this->contextId);
+        $reviewAssignment = Repo::reviewAssignment()->get($this->reviewAssignmentId);
+
+        if (
+            $submission === null ||
+            $context === null ||
+            $reviewAssignment === null
+        ) {
```
asmecher, line 51: How about just including the review assignment ID, since the submission and context can be derived from that? Fewer sanity checks and less risk of inconsistent data sneaking through.

PR #4459, opened 2024-10-03
pkp/pkp-lib#4787 Reviewer suggestions

touhidurabir:
for pkp/pkp-lib#4787


classes/migration/upgrade/v3_5_0/I4787_ReviewSuggestions.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+namespace APP\migration\upgrade\v3_5_0;
+
+class I4787_ReviewSuggestions extends \PKP\migration\upgrade\v3_5_0\I4787_ReviewSuggestions
+{
+    protected string $CONTEXT_TABLE = 'journals';
+    protected string $CONTEXT_SETTINGS_TABLE = 'journal_settings';
+    protected string $CONTEXT_COLUMN = 'journal_id';
+}
```
touhidurabir, line 23: This should probably go to a `v3_6_0`  folder .

dbscripts/xml/install.xml
```diff
@@ -39,6 +39,7 @@
 	<migration class="PKP\migration\install\TombstoneMigration" />
 	<migration class="PKP\migration\install\DoiMigration" />
 	<migration class="PKP\migration\install\InstitutionsMigration" />
+	<migration class="PKP\migration\install\ReviewerSuggestionsMigration" />
```
touhidurabir, line 42: For now , this is only a place holder as this should be a part of `3.6.0` upgrade process .

defstat (changes requested):
thanks @touhidurabir! Only some wording comments for this one.

locale/en/default.po
```diff
@@ -152,3 +152,12 @@ msgstr "Reviewers"
 
 msgid "default.groups.abbrev.externalReviewer"
 msgstr "R"
+
+msgid "default.submission.step.reviewerSuggestions"
+msgstr ""
+"<p>When submitting your article, you have the option to suggest 2/3 potential "
```
defstat, line 158: is this "2/3 potential reviewers" the correct wording here?

defstat (approved):
Thanks @touhidurabir, just one minor wording comment on your discretion.

locale/en/default.po
touhidurabir: This is taken from what provided in the design but I think you are right , better to say it like `you have the option to suggest several potential reviewers`  .  Or you have anything better in mind .

dbscripts/xml/upgrade.xml
```diff
@@ -163,6 +163,8 @@
 		<migration class="PKP\migration\upgrade\v3_5_0\I7135_CreateNewRorRegistryCacheTables"/>
 		<migration class="PKP\migration\upgrade\v3_5_0\I10759_AddReviewAssignmentSettings"/>
 		<migration class="PKP\migration\upgrade\v3_5_0\I10819_OrcidOauthScopeMail"/>
+		<migration class="APP\migration\upgrade\v3_5_0\I4787_AddReviewSuggestionHelp"/>
```
defstat, line 166: @touhidurabir I am not sure `I4787_AddReviewSuggestionHelp` and  `I4787_InstallReviewerSuggestion` are in the correct folder. I find them under `v3_6_0`

defstat (changes requested):
Just one comment regarding the migrations @touhidurabir, thanks! After that I think it will be ok to get merged :tada:

Also is the `registry/uiLocaleKeysBackend.json` supposed to be committed as part of this PR?

dbscripts/xml/upgrade.xml
touhidurabir: Weird , I can see those migrations in `v3_5_0` folder .  Also tests seems to be passing as if in wrong folder , tests would have failed at upgrade stage for not being able to find the migration file .

defstat: Maybe rebasing all branches onto main will solve that, along with the submodule conflicts? 

Just to clarify, I am always getting the PRs found here: https://github.com/pkp/pkp-lib/issues/4787#issue-447125184

Correct?

defstat: @touhidurabir it was my bad, I correctly see the migrations in 3_5_0 now. Check this, though, if you can, regarding the conflicted submodules and json file. Thanks!!

PR #4505, opened 2024-11-06
pkp/pkp-lib#1660 Customizable Reviewer Recommendations

touhidurabir:
for pkp/pkp-lib#1660

### DIFF
--- a/api/v1/reviewers/recommendations/ReviewerRecommendationController.php
+++ b/api/v1/reviewers/recommendations/ReviewerRecommendationController.php
@@ -1,0 +1,219 @@
+<?php
+
+/**
+ * @file api/v1/reviewers/recommendations/ReviewerRecommendationController.php
+ *
+ * Copyright (c) 2025 Simon Fraser University
+ * Copyright (c) 2025 John Willinsky
+ * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
+ *
+ * @class ReviewerRecommendationController
+ *
+ * @brief API controller class to handle actions on reviewer recommendations
+ *
+ */
+
+namespace APP\API\v1\reviewers\recommendations;
+
+use APP\API\v1\reviewers\recommendations\formRequests\AddReviewerRecommendation;
+use APP\API\v1\reviewers\recommendations\formRequests\EditReviewerRecommendation;
+use APP\API\v1\reviewers\recommendations\formRequests\UpdateStatusReviewerRecommendation;
+use APP\API\v1\reviewers\recommendations\resources\ReviewerRecommendationResource;
+use APP\core\Application;
+use APP\security\authorization\RecommendationAccessPolicy;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
+use Illuminate\Support\Facades\Route;
+use PKP\core\PKPBaseController;
+use PKP\core\PKPRequest;
+use PKP\security\authorization\ContextAccessPolicy;
+use PKP\security\authorization\UserRolesRequiredPolicy;
+use PKP\security\Role;
+use PKP\submission\reviewer\recommendation\ReviewerRecommendation;
+
+class ReviewerRecommendationController extends PKPBaseController
+{
+    /**
+     * @copydoc \PKP\core\PKPBaseController::getHandlerPath()
+     */
+    public function getHandlerPath(): string
+    {
+        return 'reviewers/recommendations';
+    }
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::getRouteGroupMiddleware()
+     */
+    public function getRouteGroupMiddleware(): array
+    {
+        return [
+            'has.user',
+            'has.context',
+            self::roleAuthorizer([
+                Role::ROLE_ID_SITE_ADMIN,
+                Role::ROLE_ID_MANAGER,
+                Role::ROLE_ID_SUB_EDITOR,
+            ]),
+        ];
+    }
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::authorize()
+     */
+    public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
+    {
+        $illuminateRequest = $args[0]; /** @var \Illuminate\Http\Request $illuminateRequest */
+        $actionName = static::getRouteActionName($illuminateRequest);
+
+        $this->addPolicy(new UserRolesRequiredPolicy($request), true);
+        $this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
+
+        if (in_array($actionName, ['edit', 'updateStatus', 'delete'])) {
+            $this->addPolicy(
+                new RecommendationAccessPolicy(
+                    $request,
+                    static::getRequestedRoute($illuminateRequest)->parameter('recommendationId')
+                )
+            );
+        }
+
+        return parent::authorize($request, $args, $roleAssignments);
+    }
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::getGroupRoutes()
+     */
+    public function getGroupRoutes(): void
+    {
+        Route::get('{recommendationId}', $this->get(...))
+            ->name('reviewer.recommendations.get')
+            ->whereNumber(['recommendationId']);
+
+        Route::get('', $this->getMany(...))
+            ->name('reviewer.recommendations.getMany');
+
+        Route::post('', $this->add(...))
+            ->name('reviewer.recommendations.add');
+
+        Route::put('{recommendationId}', $this->edit(...))
+            ->name('reviewer.recommendations.edit')
+            ->whereNumber(['recommendationId']);
+
+        Route::put('{recommendationId}/status', $this->updateStatus(...))
+            ->name('reviewer.recommendations.edit.status')
+            ->whereNumber(['recommendationId']);
+
+        Route::delete('{recommendationId}', $this->delete(...))
+            ->name('reviewer.recommendations.delete')
+            ->whereNumber(['recommendationId']);
+    }
+
+    /**
+     * Get specific recommendation response
+     */
+    public function get(Request $illuminateRequest): JsonResponse
+    {
+        $recommendation = ReviewerRecommendation::find($illuminateRequest->route('recommendationId'));
+
+        return response()->json(
+            (new ReviewerRecommendationResource($recommendation))->toArray($illuminateRequest),
+            Response::HTTP_OK
+        );
+    }
+
+    /**
+     * Get all recommendations response
+     */
+    public function getMany(Request $illuminateRequest): JsonResponse
+    {
+        $recommendations = ReviewerRecommendation::query()
+            ->withContextId(Application::get()->getRequest()->getContext()->getId())
+            ->get();
+
+        return response()->json([
+            'items' => ReviewerRecommendationResource::collection($recommendations),
+            'itemMax' => $recommendations->count(),
+        ], Response::HTTP_OK);
+    }
+
+    /**
+     * Add new recommendation
+     */
+    public function add(AddReviewerRecommendation $illuminateRequest): JsonResponse
+    {
+        $validateds = $illuminateRequest->validated();
+
+        $recommendation = ReviewerRecommendation::create($validateds);
+
+        return response()->json(
+            (new ReviewerRecommendationResource($recommendation->refresh()))
+                ->toArray($illuminateRequest),
+            Response::HTTP_OK
+        );
+    }
+
+    /**
+     * Update existing recommendation
+     */
+    public function edit(EditReviewerRecommendation $illuminateRequest): JsonResponse
+    {
+        $validated = $illuminateRequest->validated();
+
+        $recommendation = ReviewerRecommendation::find($illuminateRequest->route('recommendationId'));
+
+        if (!$recommendation->removable) {
+            return response()->json([
+                'error' => __('api.406.notAcceptable'),
+            ], Response::HTTP_NOT_ACCEPTABLE);
+        }
+
+        if (!$recommendation->update($validated)) {
+            return response()->json([
+                'error' => __('api.409.resourceActionConflict'),
+            ], Response::HTTP_CONFLICT);
+        }
+
+        return response()->json(
+            (new ReviewerRecommendationResource($recommendation->refresh()))
+                ->toArray($illuminateRequest),
+            Response::HTTP_OK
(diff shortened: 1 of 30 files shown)


templates/reviewer/review/reviewerRecommendations.tpl
```diff
@@ -1,13 +1,31 @@
...
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * Include reviewer recommendations for OJS review assignment responses.
  *}
 
-{fbvFormSection label="reviewer.article.recommendation" description=$description|default:"reviewer.article.selectRecommendation"}
-	{fbvElement type="select" id="recommendation" from=$reviewerRecommendationOptions selected=$reviewAssignment->getRecommendation() size=$fbvStyles.size.MEDIUM required=$required|default:true disabled=$readOnly}
+{foreach from=$reviewerRecommendationOptions item=item key=key}
+	{$reviewerRecommendationOptions.$key = $item|escape}
+{/foreach}
```
touhidurabir, line 13: @asmecher is there any better way to do it for the select ? Our FVB select does not seems to have any way to apply escape for options or I did not find any .

asmecher (changes requested):
Just one response to your earlier question, Touhidur, thanks!

asmecher:
@ewhanson, could you review the API-related parts of this? The Swagger documentation needs consideration -- when and how we should update it.
@touhidurabir, there's also a conflict on this that needs resolving (beyond just the submodule).

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+
+    /**
+     * @copydoc \PKP\core\PKPBaseController::authorize()
+     */
+    public function authorize(PKPRequest $request, array &$args, array $roleAssignments): bool
+    {
+        $illuminateRequest = $args[0]; /** @var \Illuminate\Http\Request $illuminateRequest */
+        $actionName = static::getRouteActionName($illuminateRequest);
+
+        $this->addPolicy(new UserRolesRequiredPolicy($request), true);
```
ewhanson, line 69: `$request` here is a `PKPRequest` but `UserRolesRequiredPolicy` is expecting the more specific `APP\core\Request`.

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+        $this->addPolicy(new ContextAccessPolicy($request, $roleAssignments));
+
+        if (in_array($actionName, ['edit', 'updateStatus', 'delete'])) {
+            $this->addPolicy(
+                new RecommendationAccessPolicy(
+                    $request,
```
ewhanson, line 75: Same as above: `$request` here is a `PKPRequest` but `RecommendationAccessPolicy` is expecting the more specific `APP\core\Request`.

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+            ->name('reviewer.recommendations.delete')
+            ->whereNumber(['reviewerRecommendationId']);
+    }
+
+    /**
+     * Get specific recommendation response
+     */
+    public function get(Request $illuminateRequest): JsonResponse
+    {
+        $recommendation = ReviewerRecommendation::find($illuminateRequest->route('reviewerRecommendationId'));
```
ewhanson, line 117: This should account for a recommendation not being found, returning a 404 if none is found.

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+            'itemMax' => $recommendations->count(),
+        ], Response::HTTP_OK);
+    }
+
+    /**
+     * Add new recommendation
+     */
+    public function add(AddReviewerRecommendation $illuminateRequest): JsonResponse
+    {
+        $validateds = $illuminateRequest->validated();
```
ewhanson, line 145: Typo in variable name. Also, if the data is not validated, do we manually need to return a different response here?

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+
+        $recommendation = ReviewerRecommendation::create($validateds);
```
ewhanson, line 147: Likewise, do we need to separately consider if the recommendation cannot be created and send a response accordingly?

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /**
+     * Update existing recommendation
+     */
+    public function edit(EditReviewerRecommendation $illuminateRequest): JsonResponse
+    {
+        $validated = $illuminateRequest->validated();
+
+        $recommendation = ReviewerRecommendation::find($illuminateRequest->route('reviewerRecommendationId'));
```
ewhanson, line 163: Should throw a 404 if recommendation cannot be found.

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+            Response::HTTP_OK
+        );
+    }
+
+    /**
+     * Update the status of existing recommendation
+     */
+    public function updateStatus(UpdateStatusReviewerRecommendation $illuminateRequest): JsonResponse
+    {
+        $validated = $illuminateRequest->validated();
```
ewhanson, line 189: Shouldn't this return an error if not validated?

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+
+        $recommendation = ReviewerRecommendation::find($illuminateRequest->route('reviewerRecommendationId'));
```
ewhanson, line 191: Likewise for this: return a 404 if not found.

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+
+        $recommendation->update($validated);
```
ewhanson, line 193: Could this fail? If so should return an appropriate response upon failure.

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
```diff
@@ -0,0 +1,219 @@
...
+            Response::HTTP_OK
+        );
+    }
+
+    /**
+     * Delete existing recommendation
+     */
+    public function delete(Request $illuminateRequest): JsonResponse
+    {
+        $recommendation = ReviewerRecommendation::find($illuminateRequest->route('reviewerRecommendationId'));
```
ewhanson, line 207: Should return a 404 if can't be found.

ewhanson (changes requested):
Thanks @touhidurabir. I mostly took a look at API related things here, and my main questions were around error handling in the API. I mostly wanted to check that we're catching and returning appropriate responses for error cases (see comments below).

Additionally, the API swagger/open API spec doc will need to be updated (It lives here: [swagger-source.json](https://github.com/pkp/ojs/blob/main/docs/dev/swagger-source.json)). Instructions for building the API docs locally are in the [docs repository](https://github.com/pkp/pkp-docs#generate-rest-api-references). The DOI section may be a good place to look as it's a relatively small and self-contained section. Do let me know if you have any questions!

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: method `authorize` comes from `PKPBaseController` where it requires `PKPRequest $request` . As our app level `Request` extends the `PKPRequest` and app level controller always get injected with $request object of `Request` , it ok . Also as the core `PKPBaseController::authorize` require `PKPRequest $request`, we can not change it to `Request $request` as newer version of PHP does not allow it .

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: same as https://github.com/pkp/ojs/pull/4505#discussion_r2041122202 .

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: RIght , this should get included in `in_array($actionName, ['get', 'edit', 'updateStatus', 'delete'])`. Added it .

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: > Typo in variable name

Is this for `$validateds` ? this is intentionally done as plural from of `validated` so identify as all validated data .

> if the data is not validated, do we manually need to return a different response here?

As we are using `FromRequest` as `AddReviewerRecommendation`, so any failed validation will be handled there and return back from there without reaching to controller code . Here we will get validated data after successful validation, so no need to return different response as will never reach here without completing the validation successfully .

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: we can do that but will we get such as case ? After a successful validation along with proper guarded and cast data, we should always be able to create a new recommendation .

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: We have this set up in the policy `RecommendationRequiredPolicy` but it was a `401` as it's an extend of `AuthorizationPolicy`. But it should have status code `404` so updated it . However no need to throw or have different response in controller as handled in the policy .

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: As described at https://github.com/pkp/ojs/pull/4505#discussion_r2041123053 , if validation failed, it will be handled in the `UpdateStatusReviewerRecommendation` with proper response before reaching here . Also the `validated` does not run the validation but get only the validated data/input .

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: same as https://github.com/pkp/ojs/pull/4505#discussion_r2041127375

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: same https://github.com/pkp/ojs/pull/4505#discussion_r2041127375

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
touhidurabir: same as https://github.com/pkp/ojs/pull/4505#discussion_r2041123569 . As we run through proper validation case check as like if `removable` which identify if it's already in use (as we will not allow to update if already used) , we should be able to always update .

templates/reviewer/review/reviewerRecommendations.tpl
touhidurabir: > You should be able to apply the modifier directly where you use `$reviewerRecommendationOptions`:
> 
> ```
> {fbvElement from=$reviewerRecommendationOptions|escape}
> ```
> 
> From [the Smarty docs](https://www.smarty.net/docsv2/en/language.modifiers.tpl):
> 
> > If you apply a modifier to an array variable instead of a single value variable, the modifier will be applied to every value in that array. If you really want the modifier to work on an entire array as a value, you must prepend the modifier name with a `@` symbol.

This does not work as applying the `escape` on an array , it returns as `Array` string which is expected due the implementation of escape modifier in `/lib/pkp/lib/vendor/smarty/smarty/libs/plugins/modifier.escape.php` as 
```php
$string = (string)$string;
```
which will always result in just `Array` string when casting array to string .

touhidurabir:
> Thanks @touhidurabir. I mostly took a look at API related things here, and my main questions were around error handling in the API. I mostly wanted to check that we're catching and returning appropriate responses for error cases (see comments below).
> 
> Additionally, the API swagger/open API spec doc will need to be updated (It lives here: [swagger-source.json](https://github.com/pkp/ojs/blob/main/docs/dev/swagger-source.json)). Instructions for building the API docs locally are in the [docs repository](https://github.com/pkp/pkp-docs#generate-rest-api-references). The DOI section may be a good place to look as it's a relatively small and self-contained section. Do let me know if you have any questions!

Hi @ewhanson please take look again as I have addressed all of your comment . Most of the stuff were already done but I used a different pattern using [Laravel's Form Request](https://laravel.com/docs/11.x/validation#form-request-validation) for validation which is bit different approach that usually what we usually use . 

About the `API swagger/open API spec doc` , it seems there is some issue with current swagger-source which first needed to be resolved for 3.5 (we also don't have the 3.5 doc yet) . So I plan to work on that as a different take later if that is ok .

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
ewhanson: Perfect, thanks!

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
ewhanson: Gotcha, I thought that was probably the case but I didn't know if how we were using it would handle all that automatically.

api/v1/reviewers/recommendations/ReviewerRecommendationController.php
ewhanson: In that case I don't think we need it, thanks!

ewhanson:
Thanks @touhidurabir for the clarification. I wondered if the FormRequests would work that way, but wasn't sure how they'd work in our context. Glad to hear it! Otherwise it looks good!

As for the swagger docs, I'll look at getting the 3.5 version setup, but could you let me know the error you were running into with it? It shouldn't need to be versioned in order to update the source json file in the OJS repository, only the docs hub will need the versioned output. Thanks again!

PR #4510, opened 2024-11-08
pkp/pkp-lib#10571 limit email template access by user groups

taslangraham:
For pkp/pkp-lib#10571


taslangraham:
Please see https://github.com/pkp/pkp-lib/issues/10571#issue-2624382729

PR #4526, opened 2024-11-18
pkp/pkp-lib#10506 user group to use eloquent model

Hafsa-Naeem:


Vitaliy-1:
@Hafsa-Naeem, this won't trigger tests. The update of submodule should be the last commit with the message: 
```
Submodule update ##Hafsa-Naeem/i10506_UserGroup_to_Use_Eloquent_Model##
```
All branches in submodule should have the same name: https://docs.pkp.sfu.ca/dev/testing/en/continuous-integration#tests-for-pull-requests
And you would need to rebase all to current state of main branches

Vitaliy-1 (changes requested):
Thanks, @Hafsa-Naeem! I left minor comments. Submodule updates require specific commit message and this commit should be last. I. think you can reset the last commit and divide it into 2 separate with the last one to include submodule updates to the shared library, ui library and plugins

PR #4543, opened 2024-12-06
pkp/pkp-lib#10653 remove deprecated submission functions in OAI class…

kaitlinnewson:
…es and templates

### DIFF
--- a/plugins/oaiMetadataFormats/marc/OAIMetadataFormat_MARC.php
+++ b/plugins/oaiMetadataFormats/marc/OAIMetadataFormat_MARC.php
@@ -19,6 +19,7 @@
 namespace APP\plugins\oaiMetadataFormats\marc;
 
 use APP\journal\Journal;
+use APP\publication\Publication;
 use APP\submission\Submission;
 use APP\template\TemplateManager;
 use PKP\core\PKPString;
@@ -38,6 +39,9 @@ class OAIMetadataFormat_MARC extends OAIMetadataFormat
         /** @var Submission $article */
         $article = $record->getData('article');
 
+        /** @var Publication $publication */
+        $publication = $article->getCurrentPublication();
+
         /** @var Journal $journal */
         $journal = $record->getData('journal');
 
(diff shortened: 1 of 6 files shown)


bozana (changes requested):
Hi @kaitlinnewson, 
the most of the data now belongs to the publication object, so instead of using $article->getData() we would need $currentPublication->getData(). Some places have it correct, but lots of them still don't. 
Thanks a lot!

kaitlinnewson:
Hi @bozana, ready for a second look!

plugins/oaiMetadataFormats/marc/OAIMetadataFormat_MARC.php
```diff
@@ -33,27 +37,37 @@ class OAIMetadataFormat_MARC extends OAIMetadataFormat
...
     public function toXml($record, $format = null)
     {
+        /** @var Submission $article */
         $article = $record->getData('article');
+
+        /** @var Publication $publication */
+        $publication = $article->getCurrentPublication();
+
+        $representationDao = Application::getRepresentationDAO();
+        $representations = $representationDao->getByPublicationId($publication->getId());
```
bozana, line 47: Why do we need this?

plugins/oaiMetadataFormats/marc/templates/record.tpl
```diff
@@ -63,7 +63,7 @@
 		<subfield label="a">{$identifyType|escape}</subfield>
 	</varfield>{/if}
 
-	{foreach from=$article->getGalleys() item=galley}
+	{foreach from=$representations item=galley}
```
bozana, line 66: Could we use `$publication->getData('galleys');` here?

plugins/oaiMetadataFormats/marcxml/templates/record.tpl
```diff
@@ -25,9 +25,9 @@
 			<subfield code="a">{$journal->getData('printIssn')|escape}</subfield>
 		</datafield>
 	{/if}
-	{if $article->getStoredPubId('doi')}
+	{if $publication->getStoredPubId('doi')}
```
bozana, line 28: We now use `$publication->getDoi()`.
`getStoredPubId('doi')` is old -- from the time when DOI was a plugin.

bozana (commented):
Hi @kaitlinnewson, great!
Maybe one comment only -- if we can use `$publication->getData('galleys');` instead of getting the galleys anew.
And I see here is a mix of all kinds of locales in these plugins, so I created a new issue to address that https://github.com/pkp/pkp-lib/issues/10683.
Thanks a lot!

kaitlinnewson:
@bozana fixed now!

PR #4544, opened 2024-12-06
pkp/orcidProfile#322 Unable to submit review data with multilingual submission

ewhanson:


classes/orcid/OrcidReview.php
```diff
@@ -124,7 +124,9 @@ private function build(): array
 
         $allTitles = $currentPublication->getData('title');
         foreach ($allTitles as $locale => $title) {
-            if ($locale !== $submissionLocale) {
+            if ($locale === $submissionLocale) {
+                $orcidReview['subject-name']['title'] = ['value' => $title];
```
bozana, line 128: Then the above 
```
$orcidReview['subject-name'] = [
                'title' => ['value' => $this->submission->getCurrentPublication()->getLocalizedData('title') ?? '']
];
```
should be removed?

bozana: How is title connected with the review type (if open or not)?

ewhanson: Thanks! I don't know why this was confusing me so much, but I will double check. I couldn't find reference to the 'title' even though it was right there. It seems if you send it in any case, it won't appear, but I will check to see what happens if no titles are sent at all. I'm guessing it was probably a mistake to try and send translated titles in all cases (e.g. if the review is not open).

I think the conclusion should be: 1) if the review is open, both title and translated titles should be sent together, and 2) if the review is not open, neither title nor translated title should be sent. I'll update this across all PRs.

PR #4571, opened 2024-12-27
pkp/pkp-lib#10751 Submodule Update ##touhidurabir/i10751_main##

touhidurabir:
for pkp/pkp-lib#10751


touhidurabir:
Merged https://github.com/pkp/pkp-lib/pull/10754 , will let the submodule get updated later .

PR #4572, opened 2024-12-30
[OJS][main] #10630 ***TESTS ONLY***

defstat:


defstat:
Closed in favor of https://github.com/pkp/ojs/pull/4573

PR #4574, opened 2024-12-31
pkp/pkp-lib #10744 Increase opportunity for ORCID work/review deposits

taslangraham:
For pkp/pkp-lib #10744


classes/search/ArticleSearchIndex.php
```diff
@@ -74,6 +77,10 @@ public function submissionMetadataChanged($submission)
         $this->_updateTextIndex($submissionId, SubmissionSearch::SUBMISSION_SEARCH_TYPE, (array) $publication->getData('type'));
         $this->_updateTextIndex($submissionId, SubmissionSearch::SUBMISSION_SEARCH_COVERAGE, (array) $publication->getData('coverage'));
         // FIXME Index sponsors too?
+
+        if ($publication->getData('status') == PKPSubmission::STATUS_PUBLISHED) {
```
ewhanson, line 81: This looks good for now, but just as a note, I'd eventually like to move all things making use of this for downstream metadata consumers do so via event listeners, but not necessary for 3.5. We can revisit once 3.5 is out.

PR #4585, opened 2025-01-08
pkp/pkp-lib#10690 Submodule Update ##touhidurabir/i10690_main##

touhidurabir:
for pkp/pkp-lib#10690


touhidurabir:
fixed by https://github.com/pkp/pkp-lib/pull/10785, will let the submodule get updated later .

PR #4597, opened 2025-01-14
pkp/pkp-lib#10788 Submodule Update ##touhidurabir/i10788_main##

touhidurabir:
for pkp/pkp-lib#10788


touhidurabir:
fixed by https://github.com/pkp/pkp-lib/pull/10797 , will let the submodule get updated later .

PR #4604, opened 2025-01-15
[OJS][main] #10669 Change Version support for publications

defstat:


templates/frontend/objects/article_details.tpl
```diff
@@ -335,7 +335,7 @@
 						</h2>
 						<ul class="value">
 							{foreach from=array_reverse($article->getPublishedPublications()) item=iPublication}
-								{capture assign="name"}{translate key="submission.versionIdentity" datePublished=$iPublication->getData('datePublished')|date_format:$dateFormatShort version=$iPublication->getData('version')}{/capture}
+								{capture assign="name"}{translate key="submission.versionIdentity" datePublished=$iPublication->getData('datePublished')|date_format:$dateFormatShort version=$iPublication->getData('versionDataDisplay')}{/capture}
```
asmecher, line 338: @defstat, are you sure this works? I didn't see `setData('versionDataDisplay, ...)` happening anywhere (and would prefer we didn't augment the `DataObject`s with extra attributes beyond what's in the schema).

dbscripts/xml/upgrade.xml
```diff
@@ -158,6 +158,7 @@
 		<migration class="PKP\migration\upgrade\v3_5_0\I10292_UpdateControlledVocabAssocId"/>
 		<migration class="PKP\migration\upgrade\v3_5_0\I10292_UpdateControlledVocabEntrySettingName"/>
 		<migration class="PKP\migration\upgrade\v3_5_0\FilterClassNames"/>
+		<migration class="PKP\migration\upgrade\v3_5_0\I4860_MigratePublicationVersion"/>
```
asmecher, line 161: This should be in the 3.6.0 set of migrations.

asmecher (changes requested):

templates/frontend/objects/article_details.tpl
defstat: `versionDataDisplay` renamed into 'versionString' is a Schema property of the Publication object, and is set there.

See also [this](https://github.com/pkp/pkp-lib/pull/10810#discussion_r2043814615) comment.

dbscripts/xml/upgrade.xml
defstat: Changed! (I think)

asmecher:
@ewhanson, FYI, @defstat is working on additions to the Swagger doc -- and in the process dusting off the Swagger docs. Just in case you're tinkering with the same and want to join forces rather than colliding.

asmecher (approved):

PR #4615, opened 2025-01-21
pkp/pkp-lib#10830 pass array to ArrayIterator in driver plugin

### LINKED ISSUE #10830: DRIVER plugin not producing records in OAI
**Describe the bug**
The DRIVER plugin in OJS is currently returning no records in OAI.

**To Reproduce**
1. Login and enable the DRIVER plugin
2. Open the OAI interface (e.g. `/index.php/publicknowledge/oai`)
3. Click "List Sets" and then "Records" for the DRIVER set (usually near the bottom of the list of sets)
4. See that no records are returned

**What application are you using?**
 OJS 3.4/main (working in 3.3)

**Additional information**
- [Related forum post](https://forum.pkp.sfu.ca/t/issue-with-driver-plugin-on-ojs-3-4-0-x-no-records-via-oai-interface/94582)

**Pull requests**
- OJS 3.4: https://github.com/pkp/pkp-lib/issues/10830

kaitlinnewson:


plugins/generic/driver/DRIVERDAO.php
```diff
@@ -50,7 +50,7 @@ public function &getDRIVERRecordsOrIdentifiers($setIds, $from, $until, $offset,
     {
         $records = [];
 
-        $result = new \ArrayIterator($this->_getRecordsRecordSetQuery($setIds, $from, $until, null)->get());
+        $result = new \ArrayIterator($this->_getRecordsRecordSetQuery($setIds, $from, $until, null)->get()->toArray());
```
jonasraoni, line 53: If you don't need the extra transformation, which is attempted at the `toArray()`, then the `all()` is a better fit :)

```suggestion
        $result = new \ArrayIterator($this->_getRecordsRecordSetQuery($setIds, $from, $until, null)->get()->all());
```

jonasraoni (approved):

asmecher (approved):

PR #4623, opened 2025-01-25
pkp/pkp-lib#9822 Increase timeout and improve DB table indexes for usage stats jobs

bozana:
s. https://github.com/pkp/pkp-lib/issues/9822


dbscripts/xml/upgrade.xml
```diff
@@ -235,6 +235,7 @@
 	</upgrade>
 
 	<upgrade minversion="3.4.0.0" maxversion="3.4.9.9">
+		<migration class="PKP\migration\upgrade\v3_4_0\I9822_ChangeUsageStatsTemporaryTablesIndexes"/>
```
jonasraoni, line 238: I think it's better to add new migrations to the bottom, to avoid breaking expectations of previous migrations (not going to happen here, but...).

jonasraoni (approved):

dbscripts/xml/upgrade.xml
bozana: Oh, yes, I actually always do it -- no clue why I added it here 
Thanks!

PR #4651, opened 2025-02-09
Correction in translation in manager.po (es)

alejopachons:
Correction in translation in manager.po (es)

msgstr "imprimir ISSN" --> msgstr "ISSN impreso"


asmecher:
Thanks, @alejopachons! @marcbria, could you have a look at this suggestion?

PR #4678, opened 2025-02-20
Added migration for date_considered column and submodule update ##hafsanaeem/pkp/pkp-lib#10359

Hafsa-Naeem:


dbscripts/xml/install.xml
```diff
@@ -41,6 +41,7 @@
 	<migration class="PKP\migration\install\InstitutionsMigration" />
 	<migration class="PKP\migration\install\AffiliationsMigration" />
 	<migration class="PKP\migration\install\RorsMigration" />
+	<migration class="PKP\migration\upgrade\v3_5_0\I10359_DateConsideredToReviewAssignments" />
```
Vitaliy-1, line 44: We don't need this here. Script for review assignments installation is here: https://github.com/pkp/pkp-lib/blob/ba0213f2553113b6635e901ef8ebd3dce245d0ba/classes/migration/install/ReviewsMigration.php#L48

Vitaliy-1 (changes requested):
Thanks, I've added a comment. You'll need just add a new column to the script that creates the `review_assignments` table and don't forget about OMP and OPS.

PR #4679, opened 2025-02-20
pkp/pkp-lib#10076 Show declined/cancelled reviews to editors & update submodule reference

Hafsa-Naeem:


Vitaliy-1 (approved):
Looks good! Can you also make a PR to OMP to override this method?

PR #4731, opened 2025-03-11
pkp/pkp-lib#10835 Submodule update (stable 3.5)

taslangraham:
For pkp/pkp-lib#10835


taslangraham:
Submodules were updated elsewhere

PR #4734, opened 2025-03-12
pkp/pkp-lib#11070 [main] Move Issues link below all the displayed "Dashboards" for users with multiple roles

blesildaramirez:
… for users with multiple roles


classes/template/TemplateManager.php
```diff
@@ -141,8 +141,16 @@ public function setupBackendPage()
                 'icon' => 'Issues'
             ];
             $index = false;
-            if(Config::getVar('features', 'enable_new_submission_listing')) {
-                $index = array_search('dashboards', array_keys($menu));
+            if (Config::getVar('features', 'enable_new_submission_listing')) {
+                $reviewAssignmentsIndex = array_search('reviewAssignments', array_keys($menu));
+                $mySubmissionsIndex = array_search('mySubmissions', array_keys($menu));
+                if ($mySubmissionsIndex) {
```
asmecher, line 147: Be careful using "truthy" tests in PHP with functions like `array_index`:
```php
if (array_index(...)) {
```
The `array_index` function can return array index `0` when something is found at the start of a list, or it might return `false` if no result was found. Both will fail an `if(array_search(...))` test.

Instead, use a strict test...
```php
if (array_search(...) !== false) {
```

asmecher (commented):
[One change](https://github.com/pkp/ojs/pull/4734/files#r1991887678) to suggest -- otherwise this looks good, thanks!

classes/template/TemplateManager.php
blesildaramirez: Thanks @asmecher , I applied this and also removed the config check for enable_new_submission_listing, based on your suggestion to OMP: https://github.com/pkp/omp/pull/1894#pullrequestreview-2679199227

asmecher (approved):

PR #4756, opened 2025-03-17
pkp/pkp-lib#10766 Remove support code for old submission lists

asmecher:


pages/workflow/WorkflowHandler.php
```diff
@@ -211,113 +191,6 @@ protected function getEditorAssignmentNotificationTypeByStageId($stageId)
...
-            InitialDecline::class,
-            Decline::class,
-            CancelReviewRound::class,
-            BackFromCopyediting::class,
-            BackFromProduction::class,
-        ];
-    }
-
     protected function getTitleAbstractForm(string $latestPublicationApiUrl, array $locales, Publication $latestPublication, Context $context): TitleAbstractForm
     {
```
jardakotesovec, line 195: I think this also can be removed

pages/workflow/index.php
```diff
@@ -24,7 +24,5 @@
     case 'externalReview':
     case 'editorial':
     case 'production':
```
jardakotesovec, line 26: I am only familiar with access&index. Not sure when the others are used.

pages/workflow/WorkflowHandler.php
asmecher: It's still used in `ChangeSubmissionLanguageMetadataForm`.

pages/workflow/index.php
asmecher: I figured we should leave those in place because they could conceivably be bookmarked or included in emails.

PR #4763, opened 2025-03-19
pkp/pkp-lib#11117 remove test mode for DOAJ

bozana:
s. https://github.com/pkp/pkp-lib/issues/11117


bozana:
Because this only removes code, and I tested it, I will merge it without review, so it can come into RC2....

PR #4776, opened 2025-03-20
[OJS][stable-3_5_0] #11151 Fix redirect function for ChangeProfileEmail and RegistrationAccessInvite Invitations

defstat:


asmecher:
Merged the pkp-lib PR.

PR #4778, opened 2025-03-20
pkp/pkp-lib#10659 upgrade crossref schema to 5.4.0

kaitlinnewson:


classes/migration/upgrade/v3_5_0/I10659_UpdateCrossrefSchema.php
```diff
@@ -0,0 +1,42 @@
...
+
+class I10659_UpdateCrossrefSchema extends \PKP\migration\Migration
+{
+    /**
+     * Run the migration.
+     */
+    public function up(): void
+    {
+        DB::table('filter_groups')
+            ->where('output_type', 'xml::schema(https://www.crossref.org/schemas/crossref4.3.6.xsd)')
```
kaitlinnewson, line 28: Realizing I should probably use `symbolic` instead of `output_type` here, so will update that when implementing any other changes

PR #4779, opened 2025-03-21
pkp/pkp-lib#11025 Include issue info on submission API data

taslangraham:
For pkp/pkp-lib#11025

### DIFF
--- a/classes/submission/maps/Schema.php
+++ b/classes/submission/maps/Schema.php
@@ -18,8 +18,11 @@ use APP\core\Application;
 use APP\decision\types\Accept;
 use APP\decision\types\SkipExternalReview;
 use APP\facades\Repo;
+use APP\issue\Issue;
+use APP\publication\Publication;
 use APP\submission\Submission;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Enumerable;
 use PKP\db\DAORegistry;
 use PKP\decision\DecisionType;
 use PKP\decision\types\BackFromCopyediting;
@@ -41,6 +44,9 @@ use PKP\submission\reviewRound\ReviewRoundDAO;
 
 class Schema extends \PKP\submission\maps\Schema
 {
+    /** @var Enumerable<int, Issue[]> Issues associated with submissions. Keyed by submission ID. **/
+    public Enumerable $submissionsIssues;
+
     /**
      * @copydoc \PKP\submission\maps\Schema::mapByProperties()
      */
@@ -70,6 +76,10 @@ class Schema extends \PKP\submission\maps\Schema
             $locales[] = $primaryLocale;
         }
 
+        if (in_array('issueToBePublished', $props)) {
+            $output['issueToBePublished'] = $this->getPropertyIssueToBePublished($submission->getData('publications'));
+        }
+
         $output = $this->schemaService->addMissingMultilingualValues($this->schemaService::SCHEMA_SUBMISSION, $output, $locales);
 
         ksort($output);
@@ -81,9 +91,9 @@ class Schema extends \PKP\submission\maps\Schema
     {
         return [
             'scheduledIn',
+            'issueToBePublished'
         ];
     }
-
 
     /**
      * Gets the Editorial decisions available to editors for a given stage of a submission
@@ -164,5 +174,68 @@ class Schema extends \PKP\submission\maps\Schema
         Hook::call('Workflow::Decisions', [&$decisionTypes, $stageId]);
 
         return $decisionTypes;
+    }
+
+    /**
+     * Get issues associated with submissions. Results are keyed by submission ID.
+     *
+     * @return Enumerable<int, Issue[]>
+     */
+    protected function getSubmissionsIssues(Enumerable $submissions): Enumerable
+    {
+        $submissionIds = $submissions->map(fn (Submission $submission) => $submission->getId())->all();
+        $publications = Repo::publication()->getCollector()->filterBySubmissionIds($submissionIds)->getMany();
+        $issueIds = $publications->map(fn (Publication $publication) => $publication->getData('issueId'))
+            ->unique()
+            ->all();
+
+        $issues = Repo::issue()->getCollector()
+            ->filterByContextIds([$this->context->getId()])
+            ->filterByIssueIds($issueIds)
+            ->getMany();
+
+        $issueIdsGroupedBySubmission = $publications->groupBy(fn (Publication $publication) => $publication->getData('submissionId'))
+            ->map(fn ($entry) => $entry->map(fn (Publication $publication) => $publication->getData('issueId'))
+                ->filter()); // Filter to remove any entry where `$publication->getData('issueId')` returned null
+
+        return $submissions->mapWithKeys(function ($submission) use ($issues, $publications, $issueIdsGroupedBySubmission) {
+            $submissionIssueIds = $issueIdsGroupedBySubmission->get($submission->getId());
+            return [$submission->getId() => $submissionIssueIds->mapWithKeys(fn ($id) => [$id => $issues->get($id)])];
+        });
+    }
+
+    /**
+     * Get details about the issue a submission will be published in.
+     *
+     * @param Enumerable<int, Publication> $publications Publications, keyed by publication ID.
+     */
+    protected function getPropertyIssueToBePublished(Enumerable $publications): ?array
+    {
+        /** @var Publication $latestScheduledPublication */
+        $latestScheduledPublication = $publications
+            ->filter(fn ($publication) => $publication->getData('status') === Submission::STATUS_SCHEDULED)
+            ->sortByDesc(fn (Publication $publication) => $publication->getData('version'))
+            ->first();
+
+        if (!$latestScheduledPublication) {
+            return null;
+        }
+
+        $submissionId = $latestScheduledPublication->getData('submissionId');
+        $issueId = $latestScheduledPublication->getData('issueId');
+        $issue = $this->submissionsIssues->get($submissionId)?->get($issueId);
+
+        return $issue ? [
+            'id' => $issueId,
+            'label' => $issue->getIssueIdentification()
+        ] : null;
+    }
+
+    /**
+     * Populate class properties specific to OJS.
+     */
+    protected function addAppSpecificData(Enumerable $submissions): void
+    {
+        $this->submissionsIssues ??= $this->getSubmissionsIssues($submissions);
     }
 }
--- a/schemas/submission.json
+++ b/schemas/submission.json
@@ -97,6 +97,18 @@
 					}
 				}
 			}
+		},
+		"issueToBePublished": {
+			"type": "object",
+			"properties": {
+				"id": {
+					"type": "integer"
+				},
+				"label": {
+					"type": "string"
+				}
+			},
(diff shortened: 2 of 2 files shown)


classes/submission/maps/Schema.php
```diff
@@ -41,6 +44,9 @@
 
 class Schema extends \PKP\submission\maps\Schema
 {
+    /** Issues associated with submissions. Keyed by submission ID. */
```
jonasraoni, line 47: /** @var  Enumerable<int, Issue[]> description Issues associated with submissions. Keyed by submission ID. **/

jonasraoni, line 47:
```suggestion
    /** @var  Enumerable<int, Issue[]> description Issues associated with submissions. Keyed by submission ID. **/
```

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
 
         return $decisionTypes;
     }
+
+    /**
+     * Get issues associated with submissions. Results are keyed by submission ID.
+     *
```
jonasraoni, line 181:
```suggestion
     * @param Enumerable<int, Issue[]> $submissions Issues associated with submissions. Keyed by submission ID. **/
     *
```

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+     * @return Enumerable<int, Issue[]>
+     */
+    protected function getSubmissionsIssues(Enumerable $submissions): Enumerable
+    {
+        $submissionIds = $submissions->map(fn (Submission $submission) => $submission->getId())->all();
+        $publications = Repo::publication()->getCollector()->filterBySubmissionIds($submissionIds)->getMany();
+        $issueIds = $publications->map(fn (Publication $publication) => $publication->getData('issueId'))
+            ->unique()
+            ->all();
```
jonasraoni, line 190: I guess the `Submission` object is supposed to have data in the `getData('publications')`, which will be a list of publications, then the `$submissionIds` can be removed and the `$publications` can be optimized.

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+            ->filterByContextIds([$this->context->getId()])
+            ->filterByIssueIds($issueIds)
+            ->getMany();
+
+        $issueIdsGroupedBySubmission = $publications->groupBy(fn (Publication $publication) => $publication->getData('submissionId'))
+            ->map(fn ($entry) => $entry->map(fn (Publication $publication) => $publication->getData('issueId')));
+
+        return $submissions->mapWithKeys(function ($submission) use (&$issues, $publications, $issueIdsGroupedBySubmission) {
+            $submissionIssueIds = $issueIdsGroupedBySubmission->get($submission->getId())->all();
+            return [$submission->getId() => $issues->filter(fn ($issue) => in_array($issue->getId(), $submissionIssueIds))];
```
jonasraoni, line 202: I guess the `$issues` is keyed by `issueId`, then instead of filtering through all issues, followed by an `in_array()`, you can just collect the right ones.

I mean something like `[$submission->getId() => $submissionIssueIds->map(fn (int $issueId) => $issues->get($issueId))`

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+
+        $issues = Repo::issue()->getCollector()
+            ->filterByContextIds([$this->context->getId()])
+            ->filterByIssueIds($issueIds)
+            ->getMany();
+
+        $issueIdsGroupedBySubmission = $publications->groupBy(fn (Publication $publication) => $publication->getData('submissionId'))
+            ->map(fn ($entry) => $entry->map(fn (Publication $publication) => $publication->getData('issueId')));
+
+        return $submissions->mapWithKeys(function ($submission) use (&$issues, $publications, $issueIdsGroupedBySubmission) {
```
jonasraoni, line 200: The reference is not needed, the variable isn't updated inside the callback

```suggestion
        return $submissions->mapWithKeys(function ($submission) use ($issues, $publications, $issueIdsGroupedBySubmission) {
```

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+            $submissionIssueIds = $issueIdsGroupedBySubmission->get($submission->getId())->all();
+            return [$submission->getId() => $issues->filter(fn ($issue) => in_array($issue->getId(), $submissionIssueIds))];
+        });
+    }
+
+    /**
+     * Get details about the issue a submission will be published in.
```
jonasraoni, line 207:
```suggestion
     * Get details about the issue a submission will be published in.
     * @param Enumerable<int, Publication> $publications Publications, keyed by publication ID. **/
```

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+    }
+
+    /**
+     * Populate class properties specific to OJS.
+     */
+    protected function addAppSpecificData(Enumerable $submissions): void
+    {
+        if (empty($this->submissionsIssues)) {
+            $this->submissionsIssues = $this->getSubmissionsIssues($submissions);
+        }
```
jonasraoni, line 238:
```suggestion
        $this->submissionsIssues ??= $this->getSubmissionsIssues($submissions);
```

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+    }
+
+    /**
+     * Get details about the issue a submission will be published in.
+     */
+    protected function getPropertyIssueToBePublished(Enumerable $publications): ?array
+    {
+        /** @var Publication $latestScheduledPublication */
+        $latestScheduledPublication = $publications
+            ->filter(fn ($publication) => $publication->getData('status') === Submission::STATUS_SCHEDULED)
```
jonasraoni, line 213: I guess you just need the publications which are going to be published in the future, right? Otherwise, I think the `STATUS_PUBLISHED` should be included.

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+            ->sortByDesc(fn (Publication $publication) => $publication->getData('version'))
+            ->first();
+
+        if ($latestScheduledPublication) {
```
jonasraoni, line 217: I think this type of early return is better to read

```suggestion
        if (!$latestScheduledPublication) {
            return null;
        }
```

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+                'id' => $issueId,
+                'label' => $issue->getIssueIdentification()
+            ] : null;
+        }
+
+        return null;
+    }
+
+    /**
+     * Populate class properties specific to OJS.
```
jonasraoni, line 232:
```suggestion
     * Populate class properties specific to OJS.
     * @param Enumerable<int, Submission> $submissions Submissions, keyed by submission ID.
```

classes/submission/maps/Schema.php
```diff
@@ -165,4 +175,66 @@ protected function getAvailableEditorialDecisions(int $stageId, Submission $subm
...
+        /** @var Publication $latestScheduledPublication */
+        $latestScheduledPublication = $publications
+            ->filter(fn ($publication) => $publication->getData('status') === Submission::STATUS_SCHEDULED)
+            ->sortByDesc(fn (Publication $publication) => $publication->getData('version'))
+            ->first();
+
+        if ($latestScheduledPublication) {
+            $submissionId = $latestScheduledPublication->getData('submissionId');
+            $issueId = $latestScheduledPublication->getData('issueId');
+            $issue = $this->submissionsIssues->get($submissionId, collect())->get($issueId);
```
jonasraoni, line 220: Perhaps: 
```suggestion
            $issue = $this->submissionsIssues->get($submissionId)?->get($issueId);
```

jonasraoni (changes requested):
As a reader it's not clear for me that the `$this->submissionsIssues` will be filled implicitly, some time before the `getPropertyIssueToBePublished()` is called, so I decided to watch better this `Schema` class...

---

I see the purpose of most of the code in this PR, and also the PR for the `pkp-lib`, is mostly to pre-load the list of issues, which is good, but I think there's a simpler alternative.

Looks like `$this->collection` will be filled whenever a list of submissions is going to be mapped, so you can probably drop most of the code, and use a generic `Issue` loader on the `getPropertyIssueToBePublished()`, e.g.:

```php
    private function getIssue(int $issueId): ?Issue
    {
        if ($this->cachedIssues === null) {
            if (!$this->collection) {
                $this->cachedIssues = collect([$issueId => Repo::issue()->get($issueId)]);
            } else {
                //code from the `getSubmissionsIssues()`
                $this->cachedIssues = ...;
            }
        }
        
        return $this->cachedIssues->get($issueId);
    }
```

classes/submission/maps/Schema.php
taslangraham: No, not required to be keyed by submission  ID

classes/submission/maps/Schema.php
taslangraham: we are only interested in the scheduled publications

classes/submission/maps/Schema.php
taslangraham: $submissions won't necessarily be keyed by ID, example when this method is called from `map` method in schema, like: `$this->addAppSpecificData(collect([$item]));`, that won't be keyed by submission ID

taslangraham:
> As a reader it's not clear for me that the `$this->submissionsIssues` will be filled implicitly, some time before the `getPropertyIssueToBePublished()` is called, so I decided to watch better this `Schema` class...
> 
> I see the purpose of most of the code in this PR, and also the PR for the `pkp-lib`, is mostly to pre-load the list of issues, which is good, but I think there's a simpler alternative.
> 
> Looks like `$this->collection` will be filled whenever a list of submissions is going to be mapped, so you can probably drop most of the code, and use a generic `Issue` loader on the `getPropertyIssueToBePublished()`, e.g.:
> 
> ```
>     private function getIssue(int $issueId): ?Issue
>     {
>         if ($this->cachedIssues === null) {
>             if (!$this->collection) {
>                 $this->cachedIssues = collect([$issueId => Repo::issue()->get($issueId)]);
>             } else {
>                 //code from the `getSubmissionsIssues()`
>                 $this->cachedIssues = ...;
>             }
>         }
>         
>         return $this->cachedIssues->get($issueId);
>     }
> ```

I think the suggested code kind of drifts away from the existing pattern of how data is being pre-loaded before being used in a `getPropertyXYZ` method throughout the schema. Best if this PR keeps consistent with how things are already being done.

PR #4785, opened 2025-03-25
pkp/pkp-lib#11125 Fix: Copyediting Stage, Discussion tags do not populate  [main]

taslangraham:
For https://github.com/pkp/pkp-lib/issues/11125

### DIFF
--- a/dbscripts/xml/upgrade.xml
+++ b/dbscripts/xml/upgrade.xml
@@ -172,6 +172,7 @@
 		<migration class="PKP\migration\upgrade\v3_4_0\I7624_StrftimeDeprecation"/>
 		<migration class="PKP\migration\upgrade\v3_5_0\I11238_PrepareDBForStructuredCitations"/>
 		<migration class="APP\migration\upgrade\v3_5_0\I10659_UpdateCrossrefSchema"/>
+		<migration class="APP\migration\upgrade\v3_5_0\I11125_UpdateEmailTemplateVariables"/>
 		<code function="downloadIPGeoDB" />
 		<note file="docs/release-notes/README-3.5.0" />
 	</upgrade>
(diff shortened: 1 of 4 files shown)


dbscripts/xml/upgrade.xml
```diff
@@ -124,6 +124,7 @@
 
 	<upgrade minversion="3.4.0.0" maxversion="3.4.9.9">
 		<migration class="APP\migration\upgrade\v3_4_0\I9813_QuickSubmitSubmissionProgressType"/>
+		<migration class="APP\migration\upgrade\v3_4_0\I11125_UpdateEmailTemplateVariables"/>
```
jonasraoni, line 127: I think it makes sense to fix the problem at the origin (I didn't inspect the issue very well, but according to your comment it's here: `classes/migration/upgrade/v3_4_0/I7264_UpdateEmailTemplates.php`). I mean, if we shouldn't rename the variable there, just remove it.

Then we can apply this migration only to the affected installations, so you should create an `<upgrade>` entry to match the affected range (minversion = when the defect was created, maxversion = current release).

classes/migration/upgrade/v3_4_0/I11125_UpdateEmailTemplateVariables.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+/**
+ * @file classes/migration/upgrade/v3_4_0/I11125_UpdateEmailTemplateVariables.php
+ *
+ * Copyright (c) 2014-2025 Simon Fraser University
+ * Copyright (c) 2000-2025 John Willinsky
```
jonasraoni, line 7:
```suggestion
 * Copyright (c) 2025 Simon Fraser University
 * Copyright (c) 2025 John Willinsky
```

dbscripts/xml/upgrade.xml
asmecher: Agreed, we shouldn't be referencing 3.4.0 migrations in the 3.4.0 to 3.5.0 upgrade.

PR #4791, opened 2025-03-28
pkp/pkp-lib#11185 Submodule update [stable 3.5]

taslangraham:
For pkp/pkp-lib#11185


taslangraham:
submodules were updated elsewhere

PR #4799, opened 2025-04-01
create no-embbed option for exporting issues

Godoy0722:
*Description*: I added a `--no-embbed` option for the CLI and the system interface. It reduces the memory usage, as for faster responses and lighter XML files. It solves the issue [#9769](https://github.com/pkp/pkp-lib/issues/9769).

### DIFF
--- a/plugins/importexport/native/NativeImportExportPlugin.inc.php
+++ b/plugins/importexport/native/NativeImportExportPlugin.inc.php
@@ -196,7 +196,8 @@ class NativeImportExportPlugin extends ImportExportPlugin {
 				$exportXml = $this->exportSubmissions(
 					(array) $request->getUserVar('selectedSubmissions'),
 					$request->getContext(),
-					$request->getUser()
+					$request->getUser(),
+					$this->getExportOptionsFromRequest($request)
 				);
 				import('lib.pkp.classes.file.FileManager');
 				$fileManager = new FileManager();
@@ -209,7 +210,8 @@ class NativeImportExportPlugin extends ImportExportPlugin {
 				$exportXml = $this->exportIssues(
 					(array) $request->getUserVar('selectedIssues'),
 					$request->getContext(),
-					$request->getUser()
+					$request->getUser(),
+					$this->getExportOptionsFromRequest($request)
 				);
 				import('lib.pkp.classes.file.FileManager');
 				$fileManager = new FileManager();
@@ -222,6 +224,17 @@ class NativeImportExportPlugin extends ImportExportPlugin {
 				$dispatcher = $request->getDispatcher();
 				$dispatcher->handle404();
 		}
+	}
+
+	/**
+	 * Get export options from the request object
+	 * @param $request PKPRequest
+	 * @return array
+	 */
+	function getExportOptionsFromRequest($request) {
+		$opts = array();
+		if ($request->getUserVar('no-embed')) $opts['no-embed'] = true;
+		return $opts;
 	}
 
 	/**
--- a/plugins/importexport/native/filter/IssueGalleyNativeXmlFilter.inc.php
+++ b/plugins/importexport/native/filter/IssueGalleyNativeXmlFilter.inc.php
@@ -123,10 +123,11 @@ class IssueGalleyNativeXmlFilter extends NativeExportFilter {
 			$issueFileNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'date_uploaded', strftime('%Y-%m-%d', strtotime($issueFile->getDateUploaded()))));
 			$issueFileNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'date_modified', strftime('%Y-%m-%d', strtotime($issueFile->getDateModified()))));
 
-
-			$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($filePath)));
-			$embedNode->setAttribute('encoding', 'base64');
-			$issueFileNode->appendChild($embedNode);
+			if (empty($this->opts['no-embed'])) {
+				$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($filePath)));
+				$embedNode->setAttribute('encoding', 'base64');
+				$issueFileNode->appendChild($embedNode);
+			}
 
 			$issueGalleyNode->appendChild($issueFileNode);
 			return true;
--- a/plugins/importexport/native/filter/NativeFilterHelper.inc.php
+++ b/plugins/importexport/native/filter/NativeFilterHelper.inc.php
@@ -81,9 +81,12 @@ class NativeFilterHelper {
 				$coverNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'cover_image', htmlspecialchars($coverImageName, ENT_COMPAT, 'UTF-8')));
 				$coverNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'cover_image_alt_text', htmlspecialchars($coverImage['altText'], ENT_COMPAT, 'UTF-8')));
 
-				$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($filePath)));
-				$embedNode->setAttribute('encoding', 'base64');
-				$coverNode->appendChild($embedNode);
+				// Only embed if the no-embed option was not specified
+				if (empty($filter->opts['no-embed'])) {
+					$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($filePath)));
+					$embedNode->setAttribute('encoding', 'base64');
+					$coverNode->appendChild($embedNode);
+				}
 				$coversNode->appendChild($coverNode);
 			}
 		}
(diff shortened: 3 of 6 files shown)


plugins/importexport/native/NativeImportExportPlugin.inc.php
```diff
@@ -224,6 +226,17 @@ function display($args, $request) {
 		}
 	}
 
+	/**
+	 * Get export options from the request object
+	 * @param $request PKPRequest
+	 * @return array
+	 */
+	function getExportOptionsFromRequest($request) {
```
jonasraoni, line 234:
```suggestion
	private function getExportOptionsFromRequest($request) {
```

plugins/importexport/native/filter/IssueGalleyNativeXmlFilter.inc.php
```diff
@@ -123,10 +123,11 @@ function addFile($doc, $issueGalleyNode, $issueGalley) {
 			$issueFileNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'date_uploaded', strftime('%Y-%m-%d', strtotime($issueFile->getDateUploaded()))));
 			$issueFileNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'date_modified', strftime('%Y-%m-%d', strtotime($issueFile->getDateModified()))));
 
-
-			$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($filePath)));
-			$embedNode->setAttribute('encoding', 'base64');
-			$issueFileNode->appendChild($embedNode);
+			if (empty($this->opts['no-embed'])) {
```
jonasraoni, line 126: Wherever you added support for the no-embed, there should be also the URL (I didn't check, but I assume it's possible), just to keep the standard.

plugins/importexport/native/filter/NativeFilterHelper.inc.php
```diff
@@ -116,9 +119,19 @@ function createIssueCoversNode($filter, $doc, $object) {
...
+				if (empty($filter->opts['no-embed'])) {
+					$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($filePath)));
+					$embedNode->setAttribute('encoding', 'base64');
+					$coverNode->appendChild($embedNode);
+				} else if (!empty($filter->opts['use-file-urls'])) {
+					// Use the file URL in the XML
+					$request = Application::get()->getRequest();
+					$baseUrl = $request->getBaseUrl();
+					$context = $deployment->getContext();
+					$fileUrl = $baseUrl . '/' . $context->getPath() . '/public/' . $coverImage;
```
jonasraoni, line 132: This `public` path isn't standard, you need to use the class `PublicFileManager`.

plugins/importexport/native/templates/index.tpl
```diff
@@ -111,16 +114,30 @@
...
 			{rdelim});
 		</script>
 		<form id="exportIssuesXmlForm" class="pkp_form" action="{plugin_url path="exportIssues"}" method="post">
 			{csrf}
 			{fbvFormArea id="issuesXmlForm"}
 				{capture assign="issuesListGridUrl"}{url router=$smarty.const.ROUTE_COMPONENT component="grid.issues.ExportableIssuesListGridHandler" op="fetchGrid" escape=false}{/capture}
 				{load_url_in_div id="issuesListGridContainer" url=$issuesListGridUrl}
-				{fbvFormButtons submitText="plugins.importexport.native.exportIssues" hideCancel="true"}
+				{fbvFormSection list="true"}
+					{fbvElement type="checkbox" id="validation" label="plugins.importexport.common.validation" checked=$validation|default:true}
```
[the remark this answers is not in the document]
Godoy0722: It's used in the UI interface to give the user an option to embed or not.

Godoy0722:
@jonasraoni I just made the requested updates. I'll also take a look at OMP and OPS and will open PR forms for them if it's possible to change it there. I also see that my validation was not used anywhere - I forgot to implement the validation, but it's not required, so I removed it.

plugins/importexport/native/filter/NativeFilterHelper.inc.php
```diff
@@ -224,6 +258,66 @@ function parsePublicationCover($filter, $node, $object) {
...
+						}
+
+						$fileContents = file_get_contents($imageUrl);
+						if ($fileContents === false) {
+							$deployment->addWarning(ASSOC_TYPE_PUBLICATION, $object->getId(), __('plugins.importexport.common.error.failedDownloadingFile', array('url' => $imageUrl)));
+							break;
+						}
+
+						file_put_contents($filePath, $fileContents);
+						break;
```
jonasraoni, line 320: This whole piece can be removed, once you copy the implementation for the `<href>`

```suggestion
```

plugins/importexport/native/filter/NativeFilterHelper.inc.php
```diff
@@ -284,11 +378,93 @@ function parseIssueCover($filter, $node, $object) {
 
 						file_put_contents($filePath, base64_decode($n->textContent));
 						break;
+					case 'href':
```
jonasraoni, line 381: This is similar to the previous modified code block, so the same comments can be applied

plugins/importexport/native/filter/NativeFilterHelper.inc.php
```diff
@@ -284,11 +378,93 @@ function parseIssueCover($filter, $node, $object) {
...
+				return 'image/png';
+			case 'gif':
+				return 'image/gif';
+			case 'webp':
+				return 'image/webp';
+			default:
+				return 'application/octet-stream';
+		}
+	}
+
```
jonasraoni, line 469: We have such method already in the codebase, but before using it, check if the application doesn't provide it through a `getMimeType()` (or something like that) method.

```suggestion
```

plugins/importexport/native/locale/en_US/locale.po
```diff
@@ -77,6 +77,15 @@ msgstr "A cover image with an invalid file extension was specified."
...
 
+msgid "plugins.importexport.common.error.missingHref"
+msgstr "The href element is missing the required href attribute."
+
+msgid "plugins.importexport.common.error.missingFileUrl"
+msgstr "The file_url element does not contain a URL."
+
+msgid "plugins.importexport.common.error.failedDownloadingFile"
+msgstr "Failed to download file from URL: {$url}"
+
```
jonasraoni, line 88: This can be removed once you copy the other implementation.

```suggestion
```

plugins/importexport/native/locale/en_US/locale.po
```diff
@@ -125,3 +134,27 @@ msgid "plugins.importexport.native.import.error.publishedDateMissing"
 msgstr ""
 "The article \"{$articleTitle}\" is contained within an issue, but has no "
 "published date."
+
+msgid "plugins.importexport.common.export"
+msgstr "Export"
```
jonasraoni, line 139: Generic notes for the locale keys:
- Check if generic texts already exist on the localization files and re-use them, this one does exist for example as `common.export`.
- In general, try to not create new locale keys on old branches, these are probably not going to be translated (due to problems in how we manage the translations across all the branches, the old branches are currently left to rot).
- If the locale key is going to be re-used in other applications, create them on the `pkp-lib` submodule, then their translation will be shared.

plugins/importexport/native/filter/IssueGalleyNativeXmlFilter.inc.php
```diff
@@ -120,13 +120,40 @@ function addFile($doc, $issueGalleyNode, $issueGalley) {
...
+				$dateModified = substr($dateModified, 0, 10); // Extract YYYY-MM-DD
+			}
 
-			$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($filePath)));
-			$embedNode->setAttribute('encoding', 'base64');
-			$issueFileNode->appendChild($embedNode);
+			$issueFileNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'date_uploaded', $dateUploaded));
+			$issueFileNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'date_modified', $dateModified));
+
+			if (!empty($this->opts['use-file-urls'])) {
```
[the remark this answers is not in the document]
Godoy0722: Just a message to tell you I agree with you.

plugins/importexport/native/filter/NativeFilterHelper.inc.php
```diff
@@ -81,9 +81,26 @@ function createPublicationCoversNode($filter, $doc, $object) {
 				$coverNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'cover_image', htmlspecialchars($coverImageName, ENT_COMPAT, 'UTF-8')));
 				$coverNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'cover_image_alt_text', htmlspecialchars($coverImage['altText'], ENT_COMPAT, 'UTF-8')));
 
-				$embedNode = $doc->createElementNS($deployment->getNamespace(), 'embed', base64_encode(file_get_contents($filePath)));
-				$embedNode->setAttribute('encoding', 'base64');
-				$coverNode->appendChild($embedNode);
+				if (!empty($filter->opts['use-file-urls'])) {
```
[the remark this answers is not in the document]
Godoy0722: Got it! I'll take a look and update these as well.

PR #4806, opened 2025-04-03
pkp/pkp-lib#10263 Relax editing metadata on published/posted materials

Hafsa-Naeem:
for [issue##10263](https://github.com/pkp/pkp-lib/issues/10263)

### DIFF
--- a/locale/en/submission.po
+++ b/locale/en/submission.po
@@ -254,9 +254,6 @@ msgstr "Date Published"
 msgid "publication.editDisabled"
 msgstr "This version has been published and can not be edited."
 
-msgid "publication.editorEditWarning"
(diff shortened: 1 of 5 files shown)


locale/en/submission.po
```diff
@@ -254,6 +254,9 @@ msgstr "Date Published"
 msgid "publication.editDisabled"
 msgstr "This version has been published and can not be edited."
 
+msgid "publication.editorEditWarning"
```
asmecher, line 257: This locale key is the same across all 3 apps; it should be moved to the locale files in `pkp-lib`.

asmecher (changes requested):
Just one tiny change to request here -- thanks, @Hafsa-Naeem!

Vitaliy-1 (approved):

PR #4827, opened 2025-04-11
pkp/pkp-lib#11260 fix context abbrev in publication URN to be lower case, as it is for all other objects

bozana:
s. https://github.com/pkp/pkp-lib/issues/11260

### DIFF
--- a/plugins/pubIds/urn/URNPubIdPlugin.php
+++ b/plugins/pubIds/urn/URNPubIdPlugin.php
@@ -23,6 +23,7 @@ use APP\plugins\pubIds\urn\classes\form\FieldPubIdUrn;
 use APP\plugins\pubIds\urn\classes\form\FieldTextUrn;
 use APP\plugins\pubIds\urn\classes\form\URNSettingsForm;
 use APP\template\TemplateManager;
+use Illuminate\Support\Str;
 use PKP\components\forms\FormComponent;
 use PKP\components\forms\publication\PKPPublicationIdentifiersForm;
 use PKP\linkAction\LinkAction;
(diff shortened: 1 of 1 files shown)


plugins/pubIds/urn/URNPubIdPlugin.php
```diff
@@ -353,7 +354,7 @@ public function addPublicationFormFields(string $hookName, FormComponent $form):
                 'value' => $form->publication->getData('pub-id::other::urn'),
                 'prefix' => $prefix,
                 'pattern' => $pattern,
-                'contextInitials' => $form->submissionContext->getData('acronym', $form->submissionContext->getData('primaryLocale')) ?? '',
+                'contextInitials' => preg_replace('/[^-._;()\/A-Za-z0-9]/', '', Str::lower($form->submissionContext->getData('acronym', $form->submissionContext->getData('primaryLocale')) ?? '')),
```
jonasraoni, line 357: Do you know the purpose of this regular expression?
As it allows some special characters, it's a bit curious for me... Also it will remove a lot of valid letters (Á, Ç, Я, 火, etc).

jonasraoni: I don't know how the `contextInitials` will be used down the road by the front end, and without this context it's difficult to say what should be done... Without this context, I'd say it should just leave characters and digits or perhaps be removed.

jonasraoni (changes requested):
I'd need more context to review properly, but I think something needs to be changed anyway.

plugins/pubIds/urn/URNPubIdPlugin.php
bozana: The issues where this was introduced (for DOIs): 
https://github.com/pkp/pkp-lib/issues/5505
https://github.com/pkp/pkp-lib/issues/7595

bozana: @jonasraoni, OK so?

jonasraoni: Thanks, I'll approve the PR.

I still have lazy doubts (I could investigate by myself haha), which could be handled on another issue:
- Was the `URNPubIdPlugin` created only to generate DOIs for Crossref? If not, then this regular expression should be in another place.
- Is it ok if the regular expression leaves the `contextInitials` empty? If not, then we need to input something there.

jonasraoni (approved):

plugins/pubIds/urn/URNPubIdPlugin.php
bozana: URN plugin is slightly different from DOIs and Crossref, but we treat the pattern defined by user for URN suffix in the same way as for DOIs. That means if the user defines that URN suffix should contain context initials (%j) it should behave in the same way as that (%j) in DOIs. URNs are however slightly different, e.g. they also consider if a check number needs to be added to the URN.
I suppose that the JMs are responsible to have the URN suffix patterns and the data needed for it set up properly. So if they for example use the default pattern that contains %j, and do not enter anything for contextInitials, then this is how they want their URNs to look like.

PR #4839, opened 2025-04-16
structured citations

GaziYucel:
Issue https://github.com/pkp/pkp-lib/issues/10692

@bozana: this PR is ready to be reviewed. Thanks.


templates/frontend/objects/article_details.tpl
```diff
@@ -245,19 +242,18 @@
 			{/if}
 
 			{* References *}
-			{if $parsedCitations || (string) $publication->getData('citationsRaw')}
+			{if $citations}
```
jardakotesovec: Is there way to avoid breaking changes in the template? This would affect all existing themes and customisations..

GaziYucel: Yes, that's possible. I will implement unchanging templates and check if I change templates elsewhere.

asmecher: Note that I added the `(string)` cast in order to account for [this optimization](https://github.com/pkp/pkp-lib/commit/c5c583d415163dbadf60926aa84f0dd2c9776fd6#diff-3a820da4ea9ffdae5c0c30168415095762f702c268ff91a2385b3e7f31ff41aeR181-R187), which causes the raw citations only to be fetched from the DB when necessary. This would already have to be rolled out to other themes.

jardakotesovec: @asmecher is not it something we would want to avoid in stable release? There will be lots of instances with their own article_details.tpl..  So even if we update all our themes - its still might be tricky surprise. And even if they do rely on our theme without customising article_details.tpl - I am not sure how to make obvious that they have to also update theme plugin

GaziYucel: I removed the template changes. The template and the backend php are as in upstream stable-3_5_0.

asmecher: > is not it something we would want to avoid in stable release?

Yes, usually -- but I took a liberty here, given that uptake on `stable-3_5_0` won't be high yet and the performance impact was pretty big.

PR #4849, opened 2025-04-22
change in accept invitation local value

ipula:


jardakotesovec:
Cherry picked the locale change to avoid the submodules conflict

PR #4857, opened 2025-04-29
submodule update ##kaitlinnewson/11337-3_5_0##

kaitlinnewson:


taslangraham (approved):

classes/issue/IssueAction.php
```diff
@@ -66,7 +66,7 @@ public function subscriptionRequired(Issue $issue, Journal $journal): bool
      */
     public function allowedIssuePrePublicationAccess($journal, $user)
     {
-        /** @var RoleDAO */
+        /** @var RoleDAO $roleDao */
```
jonasraoni, line 69:
```suggestion
        /** @var RoleDAO */
```

According to the [phpdoc documentation](https://docs.phpdoc.org/guide/references/phpdoc/tags/var.html#var), looks like this syntax is needed only when declaring many variables, e.g.:

```php
/**
  * @var string $name        Should contain a description
  * @var string $description Should contain a description
  */
protected $name, $description = 'Default description';
```

Or in dubious cases, like:
```php
/** @var \Sqlite3 $sqlite */
foreach ($connections as $sqlite);
```

jonasraoni (approved):

classes/issue/IssueAction.php
kaitlinnewson: Some IDEs (in my case, PHPstorm) need this for code completion and linking to references. In the docs linked it says:

"The @var tag MUST contain the name of the element it documents. An exception to this is when a declaration only refers to a single property or constant. In that case, the name of the property or constant MAY be omitted."

So I think it's okay to keep these for variables.

PR #4875, opened 2025-05-12
pkp/pkp-lib#9295 Continuous publication

touhidurabir:
for pkp/pkp-lib#9295 .


classes/components/forms/context/AccessForm.php
```diff
@@ -51,15 +51,15 @@ public function __construct($action, $locales, $context)
...
-            'label' => __('manager.distribution.publishingMode'),
-            'type' => 'radio',
-            'options' => [
-                ['value' => Journal::PUBLISHING_MODE_OPEN, 'label' => __('manager.distribution.publishingMode.openAccess')],
-                ['value' => Journal::PUBLISHING_MODE_SUBSCRIPTION, 'label' => __('manager.distribution.publishingMode.subscription')],
-                ['value' => Journal::PUBLISHING_MODE_NONE, 'label' => __('manager.distribution.publishingMode.none')],
-            ],
-            'value' => $context->getData('publishingMode'),
-        ]))
+                'label' => __('manager.distribution.publishingMode'),
```
asmecher, line 54: (I think these are accidental whitespace changes and can be removed.)

classes/core/Application.php
```diff
@@ -245,6 +245,14 @@ public function hasCustomizableReviewerRecommendation(): bool
         return true;
     }
 
+    /**
+     * Define if the application has continuous publication functionality
+     */
+    public function hasContinuousPublication(): bool
```
asmecher, line 251: I don't think this function is necessary here and in pkp-lib -- I don't see it used anywhere in the PRs.

classes/journal/enums/JournalContentOption.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        $context ??= Application::get()->getRequest()->getContext();
+        $options = static::getDefaultOptions();
+
+        if (!$context) {
+            return $options;
+        }
+
+        if (!static::getIssueExists($context)) {
+            array_shift($options);
```
asmecher, line 54: I think this behaviour should be removed and the issue option should always be available. Consider a newly installed OJS, where there won't be any issues yet; the editor won't be able to choose issue-based publication until after they've published an issue, which will be confusing.

classes/publication/Repository.php
```diff
@@ -92,7 +92,12 @@ public function validate($publication, array $props, Submission $submission, Con
             }
         }
 
-        // Ensure that the issueId exists
+        // If no issue is selected, it will be considered as continuous publication
+        if (!isset($props['issueId'])) {
```
asmecher, line 96: I think this stanza can be removed. The next `if` statement already handles it.

classes/publication/Repository.php
```diff
@@ -165,6 +172,16 @@ public function version(Publication $publication): int
...
     protected function setStatusOnPublish(Publication $publication)
     {
+        if ($publication->getData('continuousPublication')) {
+            $publication->setData('status', Submission::STATUS_PUBLISHED);
+            
+            if (!$publication->getData('datePublished')) {
+                $publication->setData('datePublished', Core::getCurrentDate());
+            }
+
+            return;
```
asmecher, line 182: Early return behaviour will make it likely that code added below this function will be inadvertently missed. It's better to avoid early returns in this kind of situation.

classes/search/ArticleSearchDAO.php
```diff
@@ -78,10 +78,10 @@ public function getPhraseResults($journal, $phrase, $publishedFrom = null, $publ
         }
 
         if (!empty($journal)) {
-            $sqlWhere .= ' AND i.journal_id = ?';
+            $sqlWhere .= ' AND (i.journal_id = ? OR i.journal_id is NULL)';
```
asmecher, line 81: I think this might leak publications between journals, because we're no longer checking the connection between publication and journal. You might want to verify the right journal by using `publications.submission_id` > `submissions.context_id` > `journals.journal_id` instead.

plugins/themes/default/styles/components/categotyHeader.less
```diff
@@ -0,0 +1,29 @@
+
+.categotyHeader{
```
asmecher, line 2: Typo in filename and content: `categoty` should be `category` (also elsewhere)

schemas/context.json
```diff
@@ -384,6 +384,12 @@
 				"nullable"
 			],
 			"defaultLocaleKey": "default.submission.step.reviewerSuggestions"
+		},
+		"continuousPublication": {
```
asmecher, line 388: I think rather than naming the workflow ("continuous publication") we should describe the entity. In this case, what `continuousPublication` means is that the submission is published, regardless of its status being scheduled or not against an issue. To match the `issues` table, we should do this by adding a `published` column to the `publications` table and use it instead. A first-class column will perform better (and reduce the joins), and it'll be less confusing than a `continuousPublication` setting.

classes/components/forms/context/AccessForm.php
touhidurabir: Yes, added form logic initially there but it later removed but some formatting changes were left .

classes/search/ArticleSearchDAO.php
touhidurabir: that correct and this is a silly mistake as I was experimenting with different query . we can move this issue relation to journal as 
```sql
( (i.published = 1 AND i.journal_id = s.context_id)
                    OR ps.publication_id IS NOT NULL
                    OR p.issue_id IS NULL )
```
and make it based on submission's context_id like `s.context_id = ?` .

classes/journal/enums/JournalContentOption.php
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Get the content options for a journal.
+     */
+    public static function getOptions(): array
+    {
+        return [
+            [
+                'value' => static::ISSUE_TOC->value,
+                'label' => __('manager.setup.journalContentOrganization.option.issue_toc'),
```
asmecher: We tend to use `lowerCamelCase` for locale keys, not `snake_case`.

asmecher (changes requested):

classes/components/forms/publication/AssignToIssueForm.php
```diff
@@ -47,7 +48,7 @@ public function __construct($action, $publication, $publicationContext)
             ->getMany();
 
         if ($unpublishedIssues->count() > 0) {
-            $issueOptions[] = ['value' => '', 'label' => '--- ' . __('editor.issues.futureIssues') . ' ---'];
+            $issueOptions[] = ['value' => '-1', 'label' => '--- ' . __('editor.issues.futureIssues') . ' ---'];
```
jardakotesovec: Curious, why this change?

locale/en/submission.po
```diff
@@ -202,12 +202,35 @@ msgstr ""
...
 msgstr ""
 "All publication requirements have been met. This will be published "
 "immediately because the date of publication has been set to "
 "{$datePublished}. Are you sure you want to publish this?"
 
+msgid "publication.publish.continuousPublication.description"
+msgstr "Allow this publication to be published regardless of the issue being published."
+
+msgid "publication.publish.continuousPublication.label"
+msgstr "Publish in continuous publication"
```
asmecher: Would "publish immediately" be a clearer label?

classes/components/forms/publication/AssignToIssueForm.php
asmecher: These are the constant values from [this PR](https://github.com/pkp/quickSubmit/pull/106/files), right? Would it make sense to move that code into the OJS repo, and refer to the constants instead of coding them as magic numbers?

classes/components/forms/publication/AssignToIssueForm.php
```diff
@@ -71,10 +72,39 @@ public function __construct($action, $publication, $publicationContext)
...
+                        'value' => true,
+                        'label' => __('publication.publish.issuelessPublication.label'),
+                    ],
+                ],
+                'value' => (bool) $publication->getData('published'),
+                'showWhen' => ['issueId', ''],
+            ]));
+        
+        foreach ($unpublishedIssues as $issue) {
+            $this->addField(new FieldOptions("continuousPublication", [
```
asmecher: (preferred to use `'` instead of `"` unless you want string interpolation)

classes/migration/upgrade/v3_6_0/I9295_AddPublishedColunmToPublications.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+/*
+ * @file classes/migration/upgrade/v3_6_0/I9295_AddPublishedColunmToPublications.php
```
asmecher: *column (here and elsewhere)

classes/migration/install/OJSMigration.php
```diff
@@ -219,6 +219,7 @@ public function up(): void
 
             $table->bigInteger('access_status')->default(0)->nullable();
             $table->date('date_published')->nullable();
+            $table->boolean('published')->default(false);
```
asmecher: We should generally be adding `comment` calls to add self-documentation to database columns. (This is published via SchemaSpy to the online documentation.) Don't worry about this for upgrades -- just new installs.

PR #4877, opened 2025-05-14
pkp/pkp-lib#11390 Allow nullable reviewer recommendation at save for later

touhidurabir:
for pkp/pkp-lib#11390 , [TEST ONLY] .

### DIFF
--- a/templates/reviewer/review/reviewerRecommendations.tpl
+++ b/templates/reviewer/review/reviewerRecommendations.tpl
@@ -25,7 +25,7 @@
 		required=$required|default:true
 		disabled=$readOnly
 		translate=false
-		defaultValue=""
+		defaultValue=" "
 		defaultLabel="common.chooseOne"|translate
 	}
 {/fbvFormSection}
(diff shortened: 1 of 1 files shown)


templates/reviewer/review/reviewerRecommendations.tpl
```diff
@@ -25,7 +25,7 @@
 		required=$required|default:true
 		disabled=$readOnly
 		translate=false
-		defaultValue=""
+		defaultValue=" "
```
Vitaliy-1, line 28: This approach will skip front-end input validation altogether but we need to skip that only for save for later button.

Looks like associated JS handler can be adjusted: https://github.com/pkp/pkp-lib/blob/e6cfd21cc0bbbcc82d95bdab8121bab8854fe439/js/controllers/form/reviewer/ReviewerReviewStep3FormHandler.js#L69
At the first glance I don't see an element with ID `recommendation` but there is `reviewerRecommendationId`. And this approach deals with required attribute depending on which button is pressed.

Vitaliy-1 (changes requested):
Thanks. Think we should adjust associated JS handler in this case.

templates/reviewer/review/reviewerRecommendations.tpl
touhidurabir: Thanks for pointing to the right direction .

touhidurabir:
Let the submodule get updated later .

PR #4886, opened 2025-05-21
pkp/pkp-lib#4325 update dtd and add keywords and references to pubmed…

kaitlinnewson:
… export

### DIFF
--- a/plugins/importexport/pubmed/filter/ArticlePubMedXmlFilter.php
+++ b/plugins/importexport/pubmed/filter/ArticlePubMedXmlFilter.php
@@ -15,6 +15,7 @@
 namespace APP\plugins\importexport\pubmed\filter;
 
 use APP\author\Author;
+use APP\core\Application;
 use APP\decision\Decision;
 use APP\facades\Repo;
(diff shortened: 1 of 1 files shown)


plugins/importexport/pubmed/filter/ArticlePubMedXmlFilter.php
```diff
@@ -135,6 +138,43 @@ public function &process(&$submissions): DOMDocument
...
+                    $keywordNode->setAttribute('Name', 'value');
+                    $objectNode->appendChild($keywordNode);
+                    $objectListNode->appendChild($objectNode);
+                }
+                $articleNode->appendChild($objectListNode);
+            }
+
+            // References
+            $citationDao = DAORegistry::getDAO('CitationDAO'); /** @var CitationDAO $citationDao */
+            $rawCitations = $citationDao->getRawCitationsByPublicationId($publication->getId());
```
bozana, line 165: Also here, I think now, with 3.5, you can use:
$citations = $publication->getData('citations')

PR #4896, opened 2025-05-26
pkp/pkp-lib#11296, pkp/pkp-lib#11366, pkp/pkp-lib#11387 [Versioning]

blesildaramirez:


cypress/tests/data/60-content/AmwandengaSubmission.cy.js
```diff
@@ -458,21 +458,22 @@ describe('Data suite: Amwandenga', function() {
...
-		cy.get('button').contains('Create New Version').click();
-		cy.contains('Are you sure you want to create a new version?');
-		cy.get('div[role=dialog]:contains("Create New Version")').get('button').contains('Yes').click();
+		cy.get(`[data-cy="active-modal"] nav a:contains('Create New Version')`).click();
+		cy.assignPublicationStage('AO', 'true');
+
+		// wait for the new publication data to appear on the side menu
+		cy.get('body').then($body =>
+			$body.find('[data-cy="dialog"]').length === 0
+		);
```
jardakotesovec: Curious, is this one necessary? I would hope that cypress will wait/retry until the Author Original 1.0 shows up? But not sure.

blesildaramirez: You're right, this is not needed so I removed it. When I was testing this locally, I had to implement this and as I go along I had another workaround, so this is not needed anymore. Thanks for catching this

PR #4898, opened 2025-05-26
pkp/pkp-lib#10962 Map contextAcronym to journalAcronym in the email template installer

Hafsa-Naeem:
fixes [#10962](https://github.com/pkp/pkp-lib/issues/10962)


dbscripts/xml/upgrade.xml
```diff
@@ -125,6 +125,7 @@
 	<upgrade minversion="3.4.0.0" maxversion="3.4.9.9">
 		<migration class="APP\migration\upgrade\v3_4_0\I9813_QuickSubmitSubmissionProgressType"/>
 		<migration class="APP\migration\upgrade\v3_5_0\I11241_MissingDecisionConstantsUpdate"/>
+		<migration class="APP\migration\upgrade\v3_4_0\I7264_UpdateEmailTemplates"/>
```
Vitaliy-1: I'm worrying about running the whole script the second time. 
E.g., we can extend `PKP\migration\upgrade\v3_4_0\I7264_UpdateEmailTemplates` and override `up()` and `down()` method only for iterating through email templates and replacing needed variable.

Vitaliy-1: The new migration class will be added for the `3.4.0.0` - `3.4.9.9` instead of the old one.

Vitaliy-1 (changes requested):
Thanks! I left a comment

Vitaliy-1 (changes requested):
Thanks! Can you double check if it also should be run for future 3.5+ upgrades? 
3.5.0.0 -> 3.5.0.2

classes/migration/upgrade/v3_4_0/I10962_UpdateEmailTemplateVariables.php
```diff
@@ -0,0 +1,33 @@
...
+    public function up(): void
+    {
+        $map = $this->oldNewVariablesMap();
+        $this->renameTemplateVariables($map);
+    }
+
+    protected function oldNewVariablesMap(): array
+    {
+        return [
+            'COPYEDIT_REQUEST' => ['contextAcronym' => 'journalAcronym'],
```
touhidurabir: is this conflicting with https://github.com/pkp/pkp-lib/issues/11125 ? so as part of `11125` , if I understand correctly , changes from `journalAcronym` to `contextAcronym` and here we are again changing from `contextAcronym` to `journalAcronym` . @taslangraham can you please confirm as you worked on `11125` ?

classes/migration/upgrade/v3_4_0/I10962_UpdateEmailTemplateVariables.php
```diff
@@ -0,0 +1,33 @@
...
+        ];
```
touhidurabir: @Hafsa-Naeem probably we should add one more following entry 
```php
'USER_ROLE_ASSIGNMENT_INVITATION' => ['contextName' => 'journalName',],
```
to handle the missing variable issue describe at https://github.com/pkp/pkp-lib/issues/11603#issuecomment-3090056452 .

classes/migration/upgrade/v3_4_0/I10962_UpdateEmailTemplateVariables.php
Hafsa-Naeem: Thanks for catching that overlap with #11125. These two PRs are targeting different templates, #11125 flips journalAcronym→contextAcronym for the invitation emails, while this remaps contextAcronym→journalAcronym for `COPYEDIT_REQUEST`. Before we move on, could you confirm which direction we want to standardize on (and for which templates)? Should we keep them scoped per-PR, or unify the mapping across all email templates? Let me know and I’ll align this PR accordingly.
@Vitaliy-1 @touhidurabir  what’s your recommendation here?

Hafsa-Naeem: Thanks for catching that overlap with #11125. These two PRs are targeting different templates, #11125 flips journalAcronym -> contextAcronym for the invitation emails, while this remaps contextAcronym -> journalAcronym for `COPYEDIT_REQUEST`. Before we move on, could you confirm which direction we want to standardize on (and for which templates)? Should we keep them scoped per-PR, or unify the mapping across all email templates? Let me know and I’ll align this PR accordingly.
@Vitaliy-1 @touhidurabir  what’s your recommendation here?

touhidurabir: @Hafsa-Naeem if I am mistaken , `11125` also made this change for the `COPYEDIT_REQUEST` , please check the change at https://github.com/pkp/ojs/pull/4784/files#diff-9ca12b8ed31cbaf9bfd01decb299455b2024df093dfacad1048f2a63ab6b8456 for 3.5 branch PR . thats where I am mostly confused . And as the PRs of `11125` got merged , is this issue still causing problem ?

taslangraham: Hey @touhidurabir @Hafsa-Naeem PRs for https://github.com/pkp/pkp-lib/issues/11125 did change `journalAcronym` to `contextAcronym` in the `COPYEDIT_REQUEST` template.  This PR would reverse that. I agree with @Hafsa-Naeem that we need to agree on a standard approach to these variables. My vote is to use names like `contextAcronym` instead of `journalAcronym`, that way we'd have a standard approach making it easier to maintain than having app-specific variations.

asmecher: Agreed, let's standardize on `contextAcronym` rather than `journalAcronym` (and e.g. `contextName` rather than `journalName`). This will mean consistent naming across the applications, and less application-specific code.

Hafsa-Naeem: yes that makes sense. so now that we've agreed to standardize on contextAcronym everywhere, this PR's purpose (mapping it back to journalAcronym) is now obsolete.I suggest we close it?

asmecher: @Hafsa-Naeem, the [original bug report](https://github.com/pkp/pkp-lib/issues/10962) is still reproducible, no?

Hafsa-Naeem: @asmecher  Yes, it’s working, `{$contextAcronym}` is correctly replaced in the `COPYEDIT_REQUEST` subject when using "Notify" from the participants menu.

But, I found another bug in Notify -> Assign Editor (Production) where `{$journalAcronym}` appears in the subject. I reproduced this locally and added a fix for it in my latest commit.

@taslangraham I also noticed that [I7264_UpdateEmailTemplates](https://github.com/pkp/ojs/blob/stable-3_5_0/classes/migration/upgrade/v3_4_0/I7264_UpdateEmailTemplates.php) converts PKP’s `context` variables back to `journal`, undoing the standardization. Since we’ve agreed to use `context` everywhere, this reversal should be removed or changed to a pass-through so PKP’s mappings remain consistent.

classes/mail/variables/ContextEmailVariable.php
```diff
@@ -20,9 +20,9 @@
 
 class ContextEmailVariable extends PKPContextEmailVariable
 {
-    public const CONTEXT_NAME = 'journalName';
-    public const CONTEXT_URL = 'journalUrl';
-    public const CONTEXT_SIGNATURE = 'journalSignature';
+    public const CONTEXT_NAME = 'contextName';
+    public const CONTEXT_URL = 'contextUrl';
+    public const CONTEXT_SIGNATURE = 'contextSignature';
```
touhidurabir: do we even need these const to define here as this extends the `PKPContextEmailVariable` which already have these defined there with `context` prefix .

classes/migration/upgrade/v3_4_0/I10962_UpdateEmailTemplateVariables.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    public function up(): void
+    {
+        DB::transaction(function () {
+            $map = [
+                'journalAcronym' => 'contextAcronym',
+                'journalName' => 'contextName',
+                'journalUrl' => 'contextUrl',
+                'journalSignature' => 'contextSignature',
+            ];
```
touhidurabir: so as this need to be ported to OMP/OPS (if I am not mistaken), how about organise the in the shared lib as only stuff that going to change is these old to new mapping ,

touhidurabir (commented):
@Hafsa-Naeem I have few a small comment . I also tested your PR in following cases using the Role invitation mechanism e.g. template `USER_ROLE_ASSIGNMENT_INVITATION ` as :-

1. upgrade from 3.3.0 using our dataset
2. upgrade from 3.4.0 using our dataset
3. upgrade from 3.5.0-0 (I have a local backup)

in all cases the it worked properly and the `{contextName}` has been properly replaced with proper information . 

So I think now we need to forward port these for `OMP` and `OPS` .  So perhaps we need to put the `I10962_UpdateEmailTemplateVariables.php` migration in shared lib and  extend in the app level to have the variable mapping .

dbscripts/xml/upgrade.xml
```diff
@@ -125,6 +125,11 @@
 	<upgrade minversion="3.4.0.0" maxversion="3.4.9.9">
 		<migration class="APP\migration\upgrade\v3_4_0\I9813_QuickSubmitSubmissionProgressType"/>
 		<migration class="APP\migration\upgrade\v3_5_0\I11241_MissingDecisionConstantsUpdate"/>
+		<migration class="PKP\migration\upgrade\v3_4_0\I10962_UpdateEmailTemplateVariables"/>
+	</upgrade>
+
+	<upgrade minversion="3.5.0.0" maxversion="3.5.9.9">
+		<migration class="PKP\migration\upgrade\v3_4_0\I10962_UpdateEmailTemplateVariables"/>
```
taslangraham: Should we be referencing 3.4 migrations in 3.5 upgrades?

touhidurabir: Good catch @taslangraham . @Hafsa-Naeem we should have the migration moved to 3_5_0 location . and we can have the 3.5.0 one in `3.4.0.0 - 3.4.9.9` block .

PR #4918, opened 2025-06-09
pkp/pkp-lib#11447 NLM title setting for pubmed export plugin

kaitlinnewson:


plugins/importexport/pubmed/PubMedExportPlugin.php
```diff
@@ -108,44 +110,72 @@ public function display($args, $request)
                 ]);
                 $templateMgr->display($this->getTemplateResource('index.tpl'));
                 break;
+            case 'settings':
```
bozana: You could also implement the function manage() as for example here: https://github.com/pkp/ojs/blob/stable-3_5_0/classes/plugins/PubObjectsExportPlugin.php#L112, so there would be no need to have this additional verb.
In index.tpl you would then use verb='index' -- the SettingsPluginGridHandler calls then the plugin manage function with the verb=index.
In settingsForm.tpl you would then use verb='save'.

But this way it works too, so... however you feel...

bozana:
Hi @kaitlinnewson, this looks good to me. Only one comment -- how other plugins work -- but because it also works this way no need to change -- as you feel/would like. It also does not load long for me :thinking:

PR #4928, opened 2025-06-12
pkp/pkp-lib#11474 Allow search by role in Users & Roles search (stable3.5)

taslangraham:
For pkp/pkp-lib#11474


Hafsa-Naeem (commented):
The test failure in `Subscriptions.cy.js` looks unrelated to the changes in this PR. Could you take a quick look or re-run the suite? Looks like a test selector/fixture issue rather than something in your PR.

PR #4963, opened 2025-06-27
pkp/pkp-lib#8920 Implement Laravel Scout and Opensearch

asmecher:
See https://github.com/pkp/pkp-lib/discussions/7968#discussioncomment-13645386 for details.


classes/search/SubmissionSearchResult.php
```diff
@@ -0,0 +1,52 @@
...
+    public function newCollection(array $models = [])
+    {
+        $collection = parent::newCollection($models);
+
+        $issueAction = new IssueAction();
+        $issueCache = [];
+        $issueAvailabilityCache = [];
+
+        $request = Application::get()->getRequest();
+        $user = $request->getUser();
```
jonasraoni, line 33: In generic code I think we should start coding without expecting that a request/context/user is available, the code should expect/deal with its absence, and if required, it could be provided through a more direct way (builders, constructors, etc).

classes/search/SubmissionSearchResult.php
```diff
@@ -0,0 +1,52 @@
...
+use APP\core\Application;
+use APP\facades\Repo;
+use APP\issue\IssueAction;
+
+class SubmissionSearchResult extends \PKP\search\SubmissionSearchResult
+{
+    /**
+     * @see Illuminate\Database\Eloquent\HasCollection.
+     */
+    public function newCollection(array $models = [])
```
jonasraoni, line 24: Would be nice to type the `$models` or the `$collection` below.

classes/search/SubmissionSearchResult.php
```diff
@@ -0,0 +1,52 @@
...
+        $issueCache = [];
+        $issueAvailabilityCache = [];
+
+        $request = Application::get()->getRequest();
+        $user = $request->getUser();
+
+        foreach ($collection as $result) {
+            $submissionId = $result['submission']->getId();
+            $issueId = $result['currentPublication']->getData('issueId');
+            $issue = $issueId ? $issueCache[$issueId] ?? ($issueCache[$issueId] = Repo::issue()->get($issueId)) : null;
```
jonasraoni:
```suggestion
            $issue = $issueId ? ($issueCache[$issueId] ??= Repo::issue()->get($issueId)) : null;
```

pages/search/SearchHandler.php
```diff
@@ -50,130 +52,59 @@ public function index($args, $request)
...
 
-        // Get and transform active filters.
-        $articleSearch = new ArticleSearch();
-        $searchFilters = $articleSearch->getSearchFilters($request);
-        $keywords = $articleSearch->getKeywordsFromSearchFilters($searchFilters);
+        $context = $request->getContext();
+        $contextId = $context?->getId();
+        if (!$context) {
+            $contextId = (int) $request->getUserVar('searchContext');
+        }
```
jonasraoni:
```suggestion
        $contextId = $context?->getId() ?? (int) $request->getUserVar('searchContext');
```

pages/search/SearchHandler.php
```diff
@@ -50,130 +52,59 @@ public function index($args, $request)
...
+            ->where('publishedFrom', $dateFrom)
+            ->where('publishedTo', $dateTo)
+            ->whereIn('categoryIds', $request->getUserVar('categoryIds'))
+            ->whereIn('sectionIds', $request->getUserVar('sectionIds'))
+            ->paginate($rangeInfo->getCount(), 'submissions', $rangeInfo->getPage());
 
         $this->setupTemplate($request);
 
         $templateMgr = TemplateManager::getManager($request);
         $templateMgr->setCacheability(TemplateManager::CACHEABILITY_NO_STORE);
```
jonasraoni: This is the default value

```suggestion
```

pages/search/SearchHandler.php
```diff
@@ -50,130 +52,59 @@ public function index($args, $request)
...
 
-        $this->_assignSearchFilters($request, $templateMgr, $searchFilters);
-        [$orderBy, $orderDir] = $articleSearch->getResultSetOrdering($request);
+        // Assign the year range.
+        $collector = Repo::publication()->getCollector();
+        if ($contextId) {
+            $collector->filterByContextIds([$contextId]);
+        }
```
jonasraoni: Some lines below there's the following pattern:

```suggestion
        $collector->filterByContextIds($contextId ? [$contextId] : null);
```

I like having less indentation/lines, but as it leads to a useless method call, I'd go with another variation (which I guess doesn't suit you):

```suggestion
        $contextId && $collector->filterByContextIds([$contextId]);
```

Just curious if you'd like to have a pattern regarding code appearance/correctness.

pages/search/SearchHandler.php
```diff
@@ -50,130 +52,59 @@ public function index($args, $request)
...
-    }
-
     /**
      * Show the search form
-     *
-     * @param array $args
-     * @param \APP\core\Request $request
      */
-    public function search($args, $request)
+    public function search(array $args, PKPRequest $request)
```
jonasraoni:
```suggestion
    public function search(array $args, PKPRequest $request): void
```

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.php
```diff
@@ -51,6 +52,44 @@ public function register($category, $path, $mainContextId = null): bool
...
 
+    /**
+     * Identify similarity terms for a given submission.
+     *
+     * @return null|array An array of string keywords or null
+     * if some kind of error occurred.
+     *
+     * @hook RecommendBySimilarityPlugin::getSimilarityTerms [$submissionId, &$searchTerms]
+     */
+    public function getSimilarityTerms(int $submissionId)
```
jonasraoni:
```suggestion
    public function getSimilarityTerms(int $submissionId): array
```

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.php
```diff
@@ -51,6 +52,44 @@ public function register($category, $path, $mainContextId = null): bool
...
+    {
+        // Check whether a search plugin provides terms for a similarity search.
+        $searchTerms = [];
+        $result = Hook::run('RecommendBySimilarityPlugin::getSimilarityTerms', [$submissionId, &$searchTerms]);
+
+        // If no plugin implements the hook then use the subject keywords
+        // of the submission for a similarity search.
+        if ($result === false) {
```
jonasraoni:
```suggestion
        if ($result == Hook::ABORT) {
          return $searchTerms;
        }
```

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.php
```diff
@@ -51,6 +52,44 @@ public function register($category, $path, $mainContextId = null): bool
         return true;
     }
 
+    /**
+     * Identify similarity terms for a given submission.
+     *
+     * @return null|array An array of string keywords or null
+     * if some kind of error occurred.
+     *
+     * @hook RecommendBySimilarityPlugin::getSimilarityTerms [$submissionId, &$searchTerms]
```
jonasraoni, line 58:
```suggestion
     * @hook RecommendBySimilarityPlugin::getSimilarityTerms [$submissionId, &$searchTerms]
     * @return array<...>
```

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.php
```diff
@@ -51,6 +52,44 @@ public function register($category, $path, $mainContextId = null): bool
...
+        // Check whether a search plugin provides terms for a similarity search.
+        $searchTerms = [];
+        $result = Hook::run('RecommendBySimilarityPlugin::getSimilarityTerms', [$submissionId, &$searchTerms]);
+
+        // If no plugin implements the hook then use the subject keywords
+        // of the submission for a similarity search.
+        if ($result === false) {
+            // Retrieve the article.
+            $submission = Repo::submission()->get($submissionId);
+            if ($submission->getData('status') === Submission::STATUS_PUBLISHED) {
```
jonasraoni: Due to the data type issue in MySQL, better to avoid strict comparison (unless this is addressed internally with coercions)

```suggestion
            if ($submission->getData('status') != Submission::STATUS_PUBLISHED) {
                return $searchTerms;
            }
```

plugins/generic/recommendBySimilarity/RecommendBySimilarityPlugin.php
```diff
@@ -51,6 +52,44 @@ public function register($category, $path, $mainContextId = null): bool
...
+                // Retrieve keywords (if any).
+                $allSearchTerms = array_filter(
+                    Repo::controlledVocab()->getBySymbolic(
+                        ControlledVocab::CONTROLLED_VOCAB_SUBMISSION_KEYWORD,
+                        Application::ASSOC_TYPE_PUBLICATION,
+                        $submission->getCurrentPublication()->getId(),
+                        [Locale::getLocale(), $submission->getData('locale'), Locale::getPrimaryLocale()]
+                    )
+                );
+                foreach ($allSearchTerms as $locale => $localeSearchTerms) {
```
jonasraoni:
```suggestion
                foreach ($allSearchTerms as $localeSearchTerms) {
```

tools/rebuildSearchIndex.php
```diff
@@ -66,9 +67,19 @@ public function execute(): void
...
         Hook::add('Request::getBaseUrl', $this->callbackBaseUrl(...));
 
+        $searchEngine = app(\Laravel\Scout\EngineManager::class)->engine();
+        $searchEngine->flush(Config::getVar('search_index_name', 'submissions'));
+
         // Let the search implementation re-build the index.
-        $articleSearchIndex = Application::getSubmissionSearchIndex();
-        $articleSearchIndex->rebuildIndex(true, $journal, $switches);
+        $submissions = Repo::submission()->getCollector()
+            ->filterByContextIds($journal ? [$journal->getId()] : [Application::SITE_CONTEXT_ID_ALL])
```
jonasraoni:
```suggestion
            ->filterByContextIds([$journal?->getId() ?? Application::SITE_CONTEXT_ID_ALL])
```

tools/rebuildSearchIndex.php
```diff
@@ -66,9 +67,19 @@ public function execute(): void
...
+            ->getIds()->chunk(100)->each(function ($submissionIds) use ($searchEngine) {
```
jonasraoni:
```suggestion
            ->getIds()
            ->chunk(100)
            ->each(function (array $submissionIds) use ($searchEngine) {
```

tools/rebuildSearchIndex.php
```diff
@@ -66,9 +67,19 @@ public function execute(): void
...
+                $submissions = Repo::submission()->getCollector()
+                    ->filterByContextIds([Application::SITE_CONTEXT_ID_ALL])
+                    ->filterBySubmissionIds($submissionIds->toArray())
+                    ->getMany();
+                $searchEngine->update($submissions);
```
jonasraoni, line 84: This should go into a job, besides the response delay (e.g. external search engine), it will deal better with retries.

tools/rebuildSearchIndex.php
```diff
@@ -66,9 +67,19 @@ public function execute(): void
         // useful URLs to journal content.
         Hook::add('Request::getBaseUrl', $this->callbackBaseUrl(...));
 
+        $searchEngine = app(\Laravel\Scout\EngineManager::class)->engine();
+        $searchEngine->flush(Config::getVar('search_index_name', 'submissions'));
+
         // Let the search implementation re-build the index.
-        $articleSearchIndex = Application::getSubmissionSearchIndex();
-        $articleSearchIndex->rebuildIndex(true, $journal, $switches);
+        $submissions = Repo::submission()->getCollector()
```
jonasraoni, line 75: I think it makes sense to sort by context, then submission ID, to make it easier to understand/resume the indexation process.

Instead of retrieving all submission IDs, we can assume we don't know how big the database is and use something like the `orderedChunkById()`.

tools/rebuildSearchIndex.php
```diff
@@ -66,9 +67,19 @@ public function execute(): void
...
+            ->filterByContextIds($journal ? [$journal->getId()] : [Application::SITE_CONTEXT_ID_ALL])
+            ->getIds()->chunk(100)->each(function ($submissionIds) use ($searchEngine) {
+                $submissions = Repo::submission()->getCollector()
+                    ->filterByContextIds([Application::SITE_CONTEXT_ID_ALL])
+                    ->filterBySubmissionIds($submissionIds->toArray())
```
jonasraoni: The `toArray()` is a bad name choice, there's another internal loop, attempting to do extra magic.
```suggestion
                    ->filterBySubmissionIds($submissionIds->all())
```

jonasraoni (approved):
Great to see the `submission_search_*` tables going away, now there will be just the `metrics_` to worry about (regarding performance/size).

Unrelated, but something to discuss:
For a soft landing, the removed classes/hooks should probably be kept alive, with `@deprecated` notes, and their behavior should be redirected to the newer structure.

OR... As it's harder and we have few hands, perhaps we can take an aggressive approach and stop worrying with back compatibility until we get a homogeneous/stable codebase (basically, the same discussion that I've opened regarding implementing the `down()` method at migrations).

classes/search/ArticleSearch.php
```diff
@@ -1,439 +0,0 @@
...
-     * @return null|array An array of string keywords or null
-     * if some kind of error occurred.
-     *
-     * @hook ArticleSearch::getSimilarityTerms [[$submissionId, &$searchTerms]]
-     */
-    public function getSimilarityTerms($submissionId)
-    {
-        // Check whether a search plugin provides terms for a similarity search.
-        $searchTerms = [];
-        $result = Hook::call('ArticleSearch::getSimilarityTerms', [$submissionId, &$searchTerms]);
```
jonasraoni, line 333: I think not all removed hooks were added to the `Hook::addUnsupportedHooks()`

classes/search/SubmissionSearchResult.php
asmecher: Agreed, but this is a carried-over FIXME from the previous implementation and would require cleanup of `IssueAction`; I'm trying to keep the scope limited here.

tools/rebuildSearchIndex.php
asmecher: Our `Collector` classes don't support multiple sort ordering, unfortunately.

jonasraoni: It's still ok to loop through the contexts, then through their submissions, not a big deal though.

pages/search/SearchHandler.php
```diff
@@ -40,140 +42,60 @@ public function authorize($request, &$args, $roleAssignments)
...
-                        0,
-                        empty($month) ? $defaultMonth : $month,
-                        empty($day) ? $defaultDay : $day,
-                        $year
-                    )
-                );
-                $hasActiveFilters = true;
-            }
-            $templateMgr->assign([
-                "date{$fromTo}Month" => $month,
```
jardakotesovec, line 105: @asmecher Hi Alec, these are need it in template. Its used to correctly select the filter from the url, and also create accurate links for pagination, so the filters does not get lost when paginating.

Tribunal33: I'm not sure if this was intended.  The list of results appears to be correct and the UI resets itself.  I would recommend leaving it as is (deleted) and establishing some requirements as identified in my comments here : https://github.com/pkp/pkp-lib/issues/8920#issuecomment-3212386646.  It's a feature not a bug :).

Tribunal33: @jardakotesovec I have updated my comment to include more requirements for this change to account for the missing advanced feature inputs.

jardakotesovec: @Tribunal33 Its definitely bug..  maybe quick catch up to go through it?

PR #4980, opened 2025-07-09
pkp/pkp-lib#11327 Implement public facing UI for user comments

taslangraham:
For WIP: pkp/pkp-lib#11327


js/load_frontend.js
```diff
@@ -1,5 +1,24 @@
...
+import * as useLocalize from '@/composables/useLocalize.js';
+import {createApp} from 'vue';
 
+import {createPinia} from 'pinia';
+import VueRegistry from '../lib/pkp/js/classes/VueRegistry';
+
+const app = createApp();
+
+const pinia = createPinia();
+app.use(pinia)
```
taslangraham: Mounting pinia because  using `useFetchPaginated` triggers use of it

templates/frontend/objects/article_details.tpl
```diff
@@ -70,6 +70,21 @@
...
+		const el = document.getElementById('scrollToComments');
+		if (el) {
+			el.onclick = function(event) {
+				event.preventDefault();
+				// Element with ID 'pkpUserCommentsContainer' is defined in the pkpUserComment component.
+				document.getElementById('pkpUserCommentsContainer')?.scrollIntoView({ behavior: 'smooth' });
+			};
+		}
+	}
+	document.addEventListener('DOMContentLoaded', attachScrollToComments);
```
ewhanson: I'm not sure how necessary it is given this is just for a Vue component in an isolated page, but you could cleanup the event listener in a `beforeunload` event.

pages/article/ArticleHandler.php
```diff
@@ -219,6 +221,68 @@ public function view($args, $request)
...
+
+        // get submission from publication
+        $submission = Repo::submission()->get($publication->getData('submissionId'));
+        // get all published  publication ids for this submission
+        $publishedPublicationIds = collect();
+        foreach ($submission->getPublishedPublications() as $publishedPublication) {
+            $publishedPublicationIds->add($publishedPublication->getId());
+        }
+
+        $templateMgr->setLocaleKeys([
```
jardakotesovec: This is good..  in this case I would suggest to follow one of the patterns we used to have. If you for example look at the ListPanel.php. You can see that class encapsulated all the ListPanel data requirements.  And I think in this case it could also be useful, rather than having these requirements directly in the handler.

I would suggest have getConfig similarly like listpanel, but also have  getLocaleKeys, which you can than pass to setLocaleKeys.   

So the mechanism would be same. Just the commenting component requirements would be encapsulated in the class.

pages/article/ArticleHandler.php
```diff
@@ -219,6 +221,68 @@ public function view($args, $request)
...
+            ->groupBy('publication_id')
+            ->pluck('count', 'publication_id')
+            ->toArray();
+
+        $allCommentsCount = array_sum($commentsCountPerPublication);
+
+        $templateMgr->assign(
+            'userCommentsInitConfig',
+            [
+                'publicationIds' => $publishedPublicationIds->sortDesc()->values(),
```
jardakotesovec: Even on the backend we are trying to reduce number of the details that we pass to the component on the initial render and rely more on the API. Last piece of data were forms - and we are already creating some new ones client side to basically have zero dependency on the data outside of the API. As result we are making more APIs that are useful for 3rd party. 

On the reader ui - as we will soon start working on the reader ui API. I was hoping to go in same direction. Where we would rely much less (as little as possible) on having custom data fetching code in handlers to provide data needed.

I know that for smarty/blade we won't be directly calling API, but hoping would could still rely on the same mappings (kind of sharing the same functions with the API) and basically making possible if someone wants to build own UI in react or something. They would be able to get everything from the API. And be able to replicate our reader ui..  

That obviously makes some extra work and considerations to thoughtfully evolve the API.

Tagging @ewhanson @asmecher - can you provide your feedback, whether thats worthy goal. And how strict we are going to be on it. At this point its somewhat my aspiration... but your guidance on this is also essential to decide whether it would be team wide consistent effort.

templates/frontend/objects/article_details.tpl
```diff
@@ -432,6 +447,22 @@
 				</div>
 			{/if}
 
+			<div class="item">
+				<section class="sub_item">
```
jardakotesovec: I would suggest to  make this as vue component as well. Including the scrolling action.

Encapsulating it to the vue.js component will make it easier to spread it to other themes, which has often its own templates.

jardakotesovec (changes requested):

templates/frontend/objects/article_details.tpl
taslangraham: reimplemented in vue component without need for event listener.

templates/frontend/objects/article_details.tpl
```diff
@@ -432,6 +432,13 @@
 				</div>
 			{/if}
 
+			<div class="item">
+				<section class="sub_item" data-vue-root>
+                    <h2 class="label"> {translate key="userComment.comments"} </h2>
+                    <pkp-scroll-to-comments v-bind='{$scrollToCommentsInitConfig|json_encode}'></pkp-scroll-to-comments>
```
jardakotesovec: Should not be the public commenting functionality behind some switch? I don't think everyone will want to use that.

taslangraham: My understanding is that it will be an ever present feature. I haven't seen anything in the requirements that suggest it should be configurable.

taslangraham: Following further discussions, we've decided that this should be behind a flag/setting. Thanks for highlighting this @jardakotesovec.

PR #4982, opened 2025-07-09
pkp/pkp-lib#10553 Different DOIs for different article versions

bozana:


locale/en/manager.po
```diff
@@ -1965,6 +1965,17 @@ msgstr "Assign DOIs to all published journal objects that have not been assigned
 msgid "doi.manager.settings.doiAssignJournalWide.confirm"
 msgstr "Are you sure you wish to assign DOIs to all published objects that have not been assigned DOIs?"
 
+msgid "doi.manager.settings.doiVersioning.description"
+msgstr ""
+"Assign a new DOI to each publication version? Most journals will "
+"want to use the default option (no), as it was till now."
```
ewhanson:
```suggestion
"want to use the default option (no), as it was until now."
```

locale/en/manager.po
```diff
@@ -1965,6 +1965,17 @@ msgstr "Assign DOIs to all published journal objects that have not been assigned
 msgid "doi.manager.settings.doiAssignJournalWide.confirm"
 msgstr "Are you sure you wish to assign DOIs to all published objects that have not been assigned DOIs?"
 
+msgid "doi.manager.settings.doiVersioning.description"
+msgstr ""
+"Assign a new DOI to each publication version? Most journals will "
```
ewhanson: I wonder if this could be something like `Most users will` rather than `Most journals/presses/etc` so it can be in pkp-lib instead. What do you think?

ewhanson (changes requested):

locale/en/manager.po
bozana: Yes, that make sense for OJS and OMP. OPS has a different text and 'yes' as default option. I think if there is a OPS specific translation, it will be taken, so yes, I will move this to pkp-lib.

bozana: Hmmm... maybe to keep them separate in the applications -- but with the same text for OJS and OMP -- I am not sure if this application specific overriding functions as I thought it would...

PR #5007, opened 2025-07-21
[OJS][stable-3_3_0] #11107 Fix issueGalley import

defstat:


plugins/importexport/native/filter/NativeXmlIssueGalleyFilter.inc.php
```diff
@@ -84,33 +84,43 @@ function handleElement($node) {
...
+							case 'file_size': $issueFile->setFileSize($o->textContent); break;
+							case 'content_type': $issueFile->setContentType((int)$o->textContent); break;
+							case 'original_file_name': $issueFile->setOriginalFileName($o->textContent); break;
+							case 'date_uploaded': $issueFile->setDateUploaded($o->textContent); break;
+							case 'date_modified': $issueFile->setDateModified($o->textContent); break;
+							case 'embed':
+								import('classes.file.IssueFileManager');
+								$issueFileManager = new IssueFileManager($issue->getId());
+								$filePath = $issueFileManager->getFilesDir() . $issueFileManager->contentTypeToPath($issueFile->getContentType()) . '/' . $issueFile->getServerFileName();
+								$issueFileManager->writeFile($filePath, base64_decode($o->textContent));
```
kaitlinnewson: Not critical as I see it's not part of your changes, but this is triggering a PHP notice for "Only variables should be passed by reference" on import due to the second parameter.

kaitlinnewson (approved):

plugins/importexport/native/filter/filterConfig.xml
```diff
@@ -109,7 +109,7 @@
 			displayName="plugins.importexport.native.displayName"
 			description="plugins.importexport.native.description"
 			inputType="xml::schema(plugins/importexport/native/native.xsd)"
-			outputType="class::classes.publication.Publication" />
+			outputType="class::classes.publication.Publication[]" />
```
kaitlinnewson, line 112: I think this would also require a database migration because it's written to the `filter_groups` table.

plugins/importexport/native/filter/NativeXmlIssueGalleyFilter.inc.php
```diff
@@ -84,33 +84,45 @@ function handleElement($node) {
...
+								case 'content_type': $issueFile->setContentType((int)$o->textContent); break;
+								case 'original_file_name': $issueFile->setOriginalFileName($o->textContent); break;
+								case 'date_uploaded': $issueFile->setDateUploaded($o->textContent); break;
+								case 'date_modified': $issueFile->setDateModified($o->textContent); break;
+								case 'embed':
+									import('classes.file.IssueFileManager');
+									$issueFileManager = new IssueFileManager($issue->getId());
+									$filePath = $issueFileManager->getFilesDir() . $issueFileManager->contentTypeToPath($issueFile->getContentType()) . '/' . $issueFile->getServerFileName();
+
+									$context = base64_decode($o->textContent);
```
kaitlinnewson, line 107: Should this be named something like `$content` or `$contents`?

plugins/importexport/native/filter/NativeXmlIssueGalleyFilter.inc.php
```diff
@@ -84,33 +84,45 @@ function handleElement($node) {
...
+								case 'file_type': $issueFile->setFileType($o->textContent); break;
+								case 'file_size': $issueFile->setFileSize($o->textContent); break;
+								case 'content_type': $issueFile->setContentType((int)$o->textContent); break;
+								case 'original_file_name': $issueFile->setOriginalFileName($o->textContent); break;
+								case 'date_uploaded': $issueFile->setDateUploaded($o->textContent); break;
+								case 'date_modified': $issueFile->setDateModified($o->textContent); break;
+								case 'embed':
+									import('classes.file.IssueFileManager');
+									$issueFileManager = new IssueFileManager($issue->getId());
+									$filePath = $issueFileManager->getFilesDir() . $issueFileManager->contentTypeToPath($issueFile->getContentType()) . '/' . $issueFile->getServerFileName();
```
jonasraoni, line 105: Considering the [line 95](https://github.com/pkp/ojs/pull/5007/files#diff-96d86d5f53095906d5da68c6487f7e4a801e2d2b715b16daea894e9264b80c1bR95) (filename coming from user input), please confirm:
- There's no room for path injections (e.g. `../../..`) 
- Empty value works fine

jonasraoni: Also, I remember file names are being hashed, so it's worth to see how the issue files are saved when uploaded by the UI and replicate the behavior here.

plugins/importexport/native/filter/NativeXmlIssueGalleyFilter.inc.php
```diff
@@ -84,33 +84,45 @@ function handleElement($node) {
...
+					$issueFileDao = DAORegistry::getDAO('IssueFileDAO'); /* @var $issueFileDao IssueFileDAO */
+					$issueFile = $issueFileDao->newDataObject();
+					$issueFile->setIssueId($issue->getId());
+
+					for ($o = $n->firstChild; $o !== null; $o=$o->nextSibling) {
+						if ($o instanceof DOMElement) {
+							switch($o->tagName) {
+								case 'file_name': $issueFile->setServerFileName($o->textContent); break;
+								case 'file_type': $issueFile->setFileType($o->textContent); break;
+								case 'file_size': $issueFile->setFileSize($o->textContent); break;
```
jonasraoni, line 97: I think this line can be removed, the attribute can be set later, once the file is stored in the disk.

```suggestion
```

plugins/importexport/native/filter/NativeXmlIssueGalleyFilter.inc.php
```diff
@@ -84,33 +84,45 @@ function handleElement($node) {
...
+								case 'content_type': $issueFile->setContentType((int)$o->textContent); break;
```
jonasraoni, line 98: The `(int)` looks suspicious here. Perhaps this attribute and the file type can also be extracted from the file itself.

plugins/importexport/native/filter/NativeXmlIssueGalleyFilter.inc.php
```diff
@@ -84,33 +84,45 @@ function handleElement($node) {
...
+								case 'original_file_name': $issueFile->setOriginalFileName($o->textContent); break;
+								case 'date_uploaded': $issueFile->setDateUploaded($o->textContent); break;
+								case 'date_modified': $issueFile->setDateModified($o->textContent); break;
+								case 'embed':
```
jonasraoni, line 102: Does the schema requires a specific ordering of elements?
If not, and if the `<embed>` is placed before the other elements, the insert will be attempted without having all information available.

plugins/importexport/native/filter/NativeXmlIssueGalleyFilter.inc.php
jonasraoni: Also good to check if the decoding was successful.

plugins/importexport/native/filter/NativeXmlIssueGalleyFilter.inc.php
```diff
@@ -84,33 +84,45 @@ function handleElement($node) {
...
+
+					$issueFileId = $issueFileDao->insertObject($issueFile);
+
+					if (!$issueFileId) {
+						throw new Exception('Failed to insert issue galley file into the database.');
+					}
+
+					$issueGalley->setFileId($issueFileId);
+				} catch (Exception $e) {
 					$deployment->addWarning(ASSOC_TYPE_ISSUE_GALLEY, 0, __('plugins.importexport.common.error.import.issueGalleyFileMissing'));
```
jonasraoni, line 122: Isn't this important enough to be considered as an error instead of a warning?

jonasraoni (approved):
Just left some notes/doubts that I had.

PR #5031, opened 2025-07-30
pkp/pkp-lib Add missing REVIEW_REMIND default data, localization key and UI variable definition

Hafsa-Naeem:
for[ #10045](https://github.com/pkp/pkp-lib/issues/10045)


classes/migration/upgrade/v3_5_0/I10045_AddReviewRemindDefaultData.php
```diff
@@ -0,0 +1,78 @@
...
+
+use Exception;
+use Illuminate\Support\Facades\DB;
+use PKP\db\XMLDAO;
+use PKP\facades\Locale;
+use PKP\migration\Migration;
+
+class I10045_AddReviewRemindDefaultData extends Migration
+{
+    public function up(): void
```
Vitaliy-1: Looks good but is it possible just to extend any other relevant `InstallEmailTemplates` migration, which implements the same. E.g., `PKP\migration\upgrade\v3_5_0\InstallEmailTemplates` or `PKP\migration\upgrade\v3_4_0\InstallEmailTemplates` just to avoid code duplication? If not, we can just leave this...

Vitaliy-1: Something like
```php
cass I10045_AddReviewRemindDefaultData extends InstallEmailTemplates
{
  public getEmailTemplateKeys()
  {
    return [
        ''REVIEW_REMIND'
    ];
  } 
  
  // If there is a need of additional stuff
  public function up() 
  {
    parent::up();
    ... // something's here
  }
}
```

PR #5039, opened 2025-08-04
pkp/pkp-lib#9295 continuous publication implementation update

touhidurabir:
for pkp/pkp-lib#9295 , second batch of PRs to update the implementation to 
1. remove the previously introduced `published` column in `publications` table
2. introduce on the `status` column of `publications` table by introducing a new status `PKPSubmission::STATUS_READY_TO_PUBLISH`
3. update the publishing/scheduling workflow by combining multiple modal into one .


classes/submission/Submission.php
```diff
@@ -31,6 +31,10 @@
 
 class Submission extends PKPSubmission
 {
+    // Submission status constants
+    public const STATUS_READY_TO_PUBLISH = 6;
+    public const STATUS_READY_TO_SCHEDULE = 7;
```
asmecher:
```suggestion
    // Not assigned to an issue and not published yet
    public const STATUS_READY_TO_PUBLISH = 6; 
    
     // Assigned to future issue but not published yet
    public const STATUS_READY_TO_SCHEDULE = 7;
```

classes/submission/Submission.php
```diff
@@ -44,6 +48,23 @@ class Submission extends PKPSubmission
     // Get/set methods
     //
 
+    /**
+     * @copydoc \PKP\submission\PKPSubmission::getStatusMap()
+     */
+    public function &getStatusMap()
+	{
+		static $statusMap = null;
```
asmecher: Just syntax, but I think you can use:
```php
static $statusMap = parent::getStatusMap() + [
    ...
];
return $statusMap;
```

classes/submission/Submission.php
asmecher: I think we'll need to be very careful about `$submission->getStatus()` vs. `$publication->getStatus()`. Currently it's a bit of a mishmash -- i.e. we expect the same constants to be used in both places -- but now that you're introducing these additional constants, it won't be so clear. For example, `$publication->getStatus()` can be `STATUS_READY_TO_SCHEDULE` while `$submission->getStatus()` is `STATUS_QUEUED`.

To keep things from getting muddled, what I'll suggest is this:
- Create new `STATUS_` constants on the `PKPPublication` / `Publication` objects for all possible values (replicating those from the `PKPSubmission` class as needed)
- Add the new values to OJS's `Publication` class instead of this
- Grep for existing uses of `STATUS_` constants and change them to use `Publication::STATUS_`... where necessary. (I know this will be a lot of changes but they should be trivial.)

asmecher (changes requested):
Just a few changes, thanks!

classes/submission/Submission.php
touhidurabir: @asmecher this is one approach I actually thought about but it has one bit of drawback to have those const to pass to ui-library also but with a bit different name or a `PUBLICATION_` prefix as we can not same const name which is overwrite the submission's const . But I see your point of having less confusion at the backend and it's not a big changes at the ui-lib end .

classes/submission/Submission.php
touhidurabir: as we moving the new ones to publication , it's not needed anymore .

touhidurabir:
@asmecher I try to follow your advice at https://github.com/pkp/ojs/pull/5039#discussion_r2286269632 and updated the PRs , only for OJS now . can you please check again .

PR #5056, opened 2025-08-19
 Update marcxml record.tpl (change 786 marc field by 773 and subfields)

horaciod:
according to official documentation of the marc bibliographic manual by loc.gov The journal title and volume must to be in 773 fields with subfield "t" and "g". the 786 field is use for other purposes .

is the same as pull request #4350 for marc schema (since april 2024).


asmecher:
Thanks, @horaciod! @bozana, could you or @kaitlinnewson have a look at this one?

bozana:
Thanks @horaciod!

PR #5058, opened 2025-08-21
pkp/pkp-lib#11702 enrich task with additional data

Vitaliy-1:


docs/dev/swagger-source.json
```diff
@@ -303,6 +308,51 @@
...
+				"participants": {
+					"type": "array",
+					"items": {
+						"$ref": "#/definitions/TaskParticipant"
+					}
+				},
+				"stageId": {
+					"type": "integer"
+				},
+				"status": {
```
blesildaramirez, line 348: Can we also add the `startedBy` in the returned data? In the frontend, we need to display the full name and username for this.

docs/dev/swagger-source.json
```diff
@@ -2591,13 +2673,479 @@
...
+							"type": "object",
+							"properties": {
+								"type": {
+									"type": "integer",
+									"description": "An integer representing the type of the task. Use `1` for a discussion and `2` for a task.",
+									"required": true
+								},
+								"title": {
+									"type": "string",
+									"description": "The title of the task. Discussions do not have titles.",
```
blesildaramirez: Based on the mockup, discussions can also have a title, we should update the description text here. Like:
`The title of the task or discussion.`

image

docs/dev/swagger-source.json
```diff
@@ -2591,13 +2673,479 @@
...
+				}
+			}
+		},
+		"/submissions/{submissionId}/tasks/{taskId}/start": {
+			"put": {
+				"tags": [
+					"Submissions / Tasks"
+				],
+				"summary": "Start a task.",
+				"description": "Sets the `dateStarted` property of the task to the current date and time. Applicable only for tasks. This action cannot be undone.",
```
blesildaramirez, line 3093: We also need to set the user information of the one who started the task.

blesildaramirez (changes requested):

docs/dev/swagger-source.json
blesildaramirez: Also, for the `status`, can we map the values for "Yet to Begin", "In Progress" & "Closed"?

Vitaliy-1: Can we re-use date started and date closed for this?

blesildaramirez: That should work, I was just wondering how the value of the `status` would change. Currently, will it always return `1`? I was thinking if these values can be set? Like `2` for "In Progress" and `3` for "Closed", so it would be easier to map each sections on the frontend.

PR #5073, opened 2025-08-31
Use single format for keywords in publication object

ajnyga:


plugins/metadata/dc11/filter/Dc11SchemaArticleAdapter.php
```diff
@@ -93,12 +93,16 @@ public function &extractMetadataFromDataObject(&$article)
...
+                true
             ),
             Repo::controlledVocab()->getBySymbolic(
                 ControlledVocab::CONTROLLED_VOCAB_SUBMISSION_SUBJECT,
                 Application::ASSOC_TYPE_PUBLICATION,
-                $publication->getId()
+                $publication->getId(),
+                [],
+                true
             )
```
touhidurabir: I am thinking , as we already have these data in the publication object and can get them like `$publication->getData('')`, perhaps we can get that those like 
```
collect($keywords)
    ->map(
        fn(array $items): array => collect($items)
            ->pluck("name")
            ->all()
    )
    ->all();
```
this will not make any DB call . I think my previous approach to reload these again form DB is flawed .

plugins/reports/articles/ArticleReportPlugin.php
```diff
@@ -154,25 +154,33 @@ public function display($args, $request)
...
+                true
             );
 
             $agencies = Repo::controlledVocab()->getBySymbolic(
                 ControlledVocab::CONTROLLED_VOCAB_SUBMISSION_AGENCY,
                 Application::ASSOC_TYPE_PUBLICATION,
-                $submission->getCurrentPublication()->getId()
+                $submission->getCurrentPublication()->getId(),
+                [],
+                true
```
touhidurabir: as me as above , lets try to get the `name` only based data from the publication object . if that above code is repetitive e.g. 
```php
collect($publication->getData('keyword'))
    ->map(
        fn(array $items): array => collect($items)
            ->pluck("name")
            ->all()
    )
    ->all();
```
we can move it to vocab repository or better publication class itself .

pages/article/ArticleHandler.php
```diff
@@ -308,7 +308,7 @@ public function view($args, $request)
             'copyrightHolder' => $publication->getLocalizedData('copyrightHolder'),
             'copyrightYear' => $publication->getData('copyrightYear'),
             'pubIdPlugins' => PluginRegistry::loadCategory('pubIds', true),
-            'keywords' => $publication->getData('keywords'),
+            'keywords' => $publication->getData('keywords'), # NOTE! Not used in the default theme templates
```
touhidurabir: lets follow the recommended approach `//` for commenting .

plugins/reports/articles/ArticleReportPlugin.php
ajnyga: this approach makes sense and seems to work ok. I did not move logic to publication class at this point, but let me know if you prefer that. The snipet is quite compact though, shorter than the original one.

classes/search/ArticleSearch.php
```diff
@@ -340,11 +340,12 @@ public function getSimilarityTerms($submissionId)
                         ControlledVocab::CONTROLLED_VOCAB_SUBMISSION_KEYWORD,
                         Application::ASSOC_TYPE_PUBLICATION,
                         $article->getCurrentPublication()->getId(),
-                        [Locale::getLocale(), $article->getData('locale'), Locale::getPrimaryLocale()]
+                        [Locale::getLocale(), $article->getData('locale'), Locale::getPrimaryLocale()],
+                        true
```
touhidurabir: it's by default true so here we can skip passing true .  However don't we only need the `name` here ?

templates/frontend/objects/article_details.tpl
```diff
@@ -173,7 +172,7 @@
 				</h2>
 				<span class="value">
 					{foreach name="keywords" from=$publication->getLocalizedData('keywords') item="keyword"}
-						{$keyword|escape}{if !$smarty.foreach.keywords.last}{translate key="common.commaListSeparator"}{/if}
+						{$keyword.name|escape}{if $keyword.identifier} ({$keyword.identifier|escape}){/if}{if !$smarty.foreach.keywords.last}{translate key="common.commaListSeparator"}{/if}
```
touhidurabir: Just want to confirm, is this the new expected presentation e.g. if the `identifier` is available , we want to show it along with the `name` ?

classes/search/ArticleSearch.php
ajnyga: yes, I had missed that call to vocabularies earlier, this is no replaced with plucking from the publication object.

templates/frontend/objects/article_details.tpl
ajnyga: This is maybe my suggestion of the presentation. My suggestion is that we add this solution and it can be revised if needed. For me the important part is that the data is there and available for a theme to use if wanted.

ajnyga:
hmm, it seems that the whole stable_3_5_0 is broken, there is a recent submodule update which did not pass the tests

touhidurabir:
> hmm, it seems that the whole stable_3_5_0 is broken, there is a recent submodule update which did not pass the tests

Yes, unfortunately . Should be resolved soon .

templates/frontend/objects/article_details.tpl
touhidurabir: Hi @asmecher @bozana just want to confirm if this is the approach we expect where there is vocabs from external source ?

bozana: I think once @jardakotesovec said that we should not change templates for 3.5 so late in the release process, if not really needed -- because we then need to adapt all themes.
So I am not sure :-( I am also not sure how important this change is. Maybe @asmecher and @jardakotesovec can say what they think?

ajnyga: We need to change the template anyway because it will show `array` if we do not. But if you do not want the identifier to show up, we can remove that of course.

bozana: How would that then look like, with the identifier? Do you  Antti-Jussi maybe have an example? -- I have not followed the developments on vocabularies that much...

ajnyga: Näyttökuva 2025-11-19 kello 14 43 04

bozana: Thanks a lot @ajnyga! 
Are the identifiers mostly URLs? -- Could we theoretically link the keyword?

touhidurabir: Hi @ajnyga @bozana , after a discussion with @asmecher , we think that for now it's better not to have the source included in the keywords presentation as this can get big and unreadable when there is a lots of keywords are associated with the submission . Rather than this approach to present it like this way, we can have it how as popover on hover on those keywords and have the keywords mark in different way . Another options I can think of is like `some keyword(<icon>)` where on hover of icon will show the source as popover . Or perhaps keywords a link to external source ?

Or for now , we can just skip in front view until have a better solution/approach decided on .

bozana: Yes, maybe to also ask Devika (but providing her information what could identifiers be)?

ajnyga: maybe then just show the keyword for now - the other data is still there available for themes to use if needed.

Then we can have a longer discussion later about the UI implications for Default theme.

ajnyga: So in many cases the identifiers could be URLs but in Frascati for example they are not.

asmecher: I don't think the front end needs to display information about the vocabulary used -- just the keywords.

ajnyga: so to clarify, the frontend in the code does not show the source/vocabulary used, It shows the identifier of the word together with the label.

ajnyga: Hi, so to not drag this anymore I suggest that I will remove identifier from the frontend and just list the keywords. You can discuss later if something else is wanted.

touhidurabir: > Hi, so to not drag this anymore I suggest that I will remove identifier from the frontend and just list the keywords. You can discuss later if something else is wanted.

@ajnyga I agree. Please remove it . It's pretty much all green at my end and once we have an approval form second reviewer, we can port it to main and complete it .

PR #5088, opened 2025-09-11
pkp/pkp-lib#11327 Public comment headless version

jardakotesovec:


plugins/themes/default/styles/components/comments.less
```diff
@@ -0,0 +1,153 @@
...
+    > svg {
+        color: @offset;
+        height: 1rem;
+        width: 1rem;
+    }
+
+}
+
+
+.BaseCommentsNotificationMessageApproved {
```
taslangraham:
```suggestion
.BaseCommentsNotificationMessageNeedsApproval {
```

templates/frontend/objects/article_details.tpl
```diff
@@ -286,6 +285,16 @@
 				</section>
 			{/if}
 
+			{if $enablePublicComments}
+				<section class="item comments" data-vue-root>
+					<h2 class="label">
+						{translate key="comments.commentsOnArticle"}
```
taslangraham: The locale key `comments.commentsOnArticle` is set to "_Comments on this article"_. Since comments will eventually be available across all apps, "_article_" isn’t always applicable. I’d recommend using a more general key.

The `userComment.commentOnThisPublication` locale was originally created for this purpose, but I’ve just noticed two typos in it. The key should be `userComment.commentsOnThisPublication`, and the `msgstr` should be "_Comments on this publication_". Could you fix those typos and use that locale here?

PR #5117, opened 2025-10-01
pkp/pkp-lib#9295 client side only implementation for issue selection

touhidurabir:
for pkp/pkp-lib#9295 cleanup


classes/template/TemplateManager.php
```diff
@@ -139,24 +146,8 @@ public function setupBackendPage()
...
-
-            if ($index === false || count($menu) <= $index + 1) {
-                $menu['issues'] = $issuesLink;
-            } else {
-                $menu = array_slice($menu, 0, $index + 1, true)
-                    + ['issues' => $issuesLink]
-                    + array_slice($menu, $index + 1, null, true);
-            }
+            
+            $menu['content']['submenu'] = ['issues' => $issuesLink] + $menu['content']['submenu'];
```
jardakotesovec: This change seems unrelated? Blessie was working on moving issue to the content menu. So maybe it somehow slipped here?

pages/index/IndexHandler.php
```diff
@@ -76,8 +76,12 @@ public function index($args, $request)
...
 
             $activeTheme = $templateMgr->getTemplateVars('activeTheme');
+            $journalContentOptions = $activeTheme->getOption('journalContentOrganization');
+            if (!is_array($journalContentOptions)) {
+                $journalContentOptions = JournalContentOption::default($journal);
+            }
 
-            if (in_array(JournalContentOption::CATEGORY_LISTING->value, $activeTheme->getOption('journalContentOrganization'))) {
+            if (in_array(JournalContentOption::CATEGORY_LISTING->value, $journalContentOptions)) {
                 $categories = Repo::category()
```
jardakotesovec, line 85: Curious what this is addressing?

touhidurabir: As part of `9295` we had added some extra option to default theme for content organisation which was causing issue with other themes . so this add a fallback so that even though other these does not get those options added(we will probably add those eventually), they will not break .

classes/template/TemplateManager.php
touhidurabir: I initially worked on to push the `issue` side menu under the `contents` but it seems Blessie also did a similar PR after me which got merged already .

PR #5170, opened 2025-11-06
pkp/pkp-lib#12005 Issue always required for legacy DOI suffix pattern

ewhanson:


bozana:
Hi @ewhanson, all fine, so I merged.
Thanks a lot!

ajnyga:
I think this still has a problem.

It is presumed here https://github.com/pkp/ojs/blob/stable-3_4_0/classes/doi/Repository.php#L67 that there is a issueId available but I think that if there is no issueId yet and the user tries to create a DOI with any custom pattern, it leads to an error because it tries to fetch an issue with NULL.

Maybe

`$issue = $publication->getData('issueId') ? Repo::issue()->get($publication->getData('issueId')) : null;
`

ewhanson:
Thanks for noticing that @ajnyga. Have you tested the fix you suggest? Would you be able to make a PR for it?

PR #5247, opened 2025-12-23
pkp/pkp-lib#12140 Back port decision lost solution for 3.4 introduced for 3.5/main upgrade

touhidurabir:
for pkp/pkp-lib#12140 .

### DIFF
--- a/classes/migration/upgrade/v3_5_0/I11241_MissingDecisionConstantsUpdate.php
+++ b/classes/migration/upgrade/v3_5_0/I11241_MissingDecisionConstantsUpdate.php
@@ -101,7 +101,14 @@ class I11241_MissingDecisionConstantsUpdate extends \PKP\migration\upgrade\v3_4_
             return;
         }
 
-        // Upgrading from a 3.4.0-*
+        // If upgrading from 3.4.0-11 or above, then we have a fixed applied also for it
+        // at https://github.com/pkp/pkp-lib/issues/12140
+        // so nothing to do and return
+        if ($currentVersion->major == 3 && $currentVersion->minor == 4 && $currentVersion->build >= 11) {
+            return;
+        }
+
+        // Upgrading from a 3.4.0-10 or below version
         // Need to figure out the first installed date of 3.4.0-*
         // Then need to update the decisions made before the first version of 3.4.0-* installed
         $firstVersionOf34 = DB::table('versions')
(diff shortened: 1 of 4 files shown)


classes/migration/upgrade/v3_5_0/I11241_MissingDecisionConstantsUpdate.php
```diff
@@ -101,7 +101,14 @@ public function up(): void
             return;
         }
 
-        // Upgrading from a 3.4.0-*
+        // If upgrading from 3.4.0-11 or above, then we have a fixed applied also for it
+        // at https://github.com/pkp/pkp-lib/issues/12140
+        // so nothing to do and return
+        if ($currentVersion->major == 3 && $currentVersion->minor == 4 && $currentVersion->build >= 11) {
+            return;
+        }
```
touhidurabir, line 106: wondering if we should make it more robust by checking if there was a 3.4.0-11 version ever installed . as we have following upgrade path one for 3.4.0 
```xml
<upgrade minversion="3.4.0.0" maxversion="3.4.0.11">
		<!-- Only new 3.4 installations need this fix -->
		<migration class="PKP\migration\upgrade\v3_4_0\I12140_MissingDecisionConstantsUpdate"/>
	</upgrade>
```
so , this patch will be only applied for 3.4.0-11 upgrade from 3.4.0-10 or below . But if a installation upgrade from say 3.4.0-8 to 3.4.0-12, this patch will not be applied for it when upgrading from 3.4.0-12 to 3.5.0-*. so having a check like following 
```php
// If upgrading from 3.4.0-11 or above, then we have a fixed applied also for it
        // at https://github.com/pkp/pkp-lib/issues/12140
        if ($currentVersion->major == 3 && $currentVersion->minor == 4 && $currentVersion->build >= 11) {
            // return;
            $versionOf34_11 = DB::table('versions')
                ->where('product', Application::get()->getName())
                ->where('product_type', 'core')
                ->where('major', 3)
                ->where('minor', 4)
                ->where('build', 11)
                ->first();
            
            // check if the patch applied for 3.4.0-11 exists and if so, 
            // nothing to do and return
            if ($versionOf34_11) {
                return;
            }
        }
```

but again if installation do upgrade from say 3.4.0-5 to directly 3.4.0-12 without this patch and then manually run the 3.4.0-11 migration, then the above check will mess it again . Need a resolution on this .

touhidurabir: After bit more thinking, I think the above assumption is wrong and will make it worse . what we already have at this PR is ok and should work .

jonasraoni (approved):

asmecher:
Cherry-picked

PR #5272, opened 2026-01-14
locales: finalize Uzbek(Cyrillic, Latin) translation for OJS 3.x

shoyim:
### What
- Adds `locale/uz_Cyrl`, a new, more complete Cyrillic-script Uzbek translation.
- Restores `locale/uz` (Latin script) from the current `main` content rather than deleting it, and blanks out only the strings in it that were in Cyrillic script, so they fall back to English and can be re-translated in Latin via Weblate.

### Why
The previous `locale/uz` translation was a mix of scripts. Of 1,209 translated strings across `locale/uz/*.po`: 1,003 (83%) were already untranslated (blank), 161 (13%) were real Latin-script translations, and 45 (4%) were Cyrillic-script strings. Deleting the whole locale (as this PR originally did) would have discarded those 161 existing Latin-script translations for anyone already using `uz`. Splitting the two scripts into `uz` (Latin) and `uz_Cyrl` (Cyrillic) avoids that regression on upgrade.


asmecher:
@shoyimobloqulov, thanks for the PR! Have you seen our [Translating guide](https://docs.pkp.sfu.ca/translating-guide/en/)? We generally accept translate translations through PKP's [Weblate installation](https://translate.pkp.sfu.ca/), which makes collaboration much easier than with just Github.

Was the removal of the old `uz` locale in favour of `uz_Cyrl` intentional? Can you give me some more information about that?

shoyimobloqulov:
The Uzbek language has two types of writing, so I created it in Crl and
Latin, and added existing additions.

Chor, 14-yan, 2026, 20:37 Alec Smecher ***@***.***> yozgan:

shoyimobloqulov:
Yes, it was!

Sesh, 27-yan, 2026, 01:41 Alec Smecher ***@***.***> yozgan:

asmecher:
Thanks, @shoyimobloqulov, and sorry for more questions, but I need to make sure we're not going to cause pain for users who were already using `uz` when they upgrade.

It looks like the previous `uz` translation was a mix of Cyrillic and Latin character sets. Wouldn't it make more sense to delete the Cyrillic parts of that translation so that they can be re-translated in Latin, rather than removing the entire translation including its Latin parts?

shoyim:
Thanks for catching that, @asmecher — you're right. I checked the deleted `locale/uz/*.po` files: out of 1,211 translated strings, only 44 (mostly in `admin.po` and `author.po`) were actually in Cyrillic script; the rest (~97%) were valid Latin-script Uzbek translations.

I've pushed an update that keeps `locale/uz` (Latin) in place and only blanks out those 44 Cyrillic-script strings (so they fall back to English and can be re-translated in Latin via Weblate), instead of deleting the whole locale. The new `locale/uz_Cyrl` addition is unchanged. I've also filled in the PR description with this context.

shoyim:
Correction on my numbers above: I had checked the deleted files' content from this branch's own history, which turned out to be stale — `locale/uz` on `main` had continued to receive independent Weblate updates after this PR was opened (it also had a `locale/uz/locale.po` file my restore was missing entirely, which is what caused the merge conflict flagged above).

Rebuilding from the actual current `main` content, the accurate picture across all 1,209 strings in `locale/uz/*.po` is:
- **1,003 (83%)** are already untranslated (blank, falling back to English)
- **161 (13%)** are real Latin-script translations
- **45 (4%)** are Cyrillic-script strings

So it's a smaller set of Latin translations than I first said, but the principle holds: I've pushed an update (b2fd823109) that rebuilds `locale/uz` from current `main` and blanks only those 45 Cyrillic-script strings, preserving the 161 Latin ones. This also resolves the merge conflict — the PR now shows as mergeable against `main`.

asmecher:
Thanks, @shoyim! Are these human-generated translations (and pull requests), or generated with AI? If you're using an LLM, please check our [AI policy](https://pkp.sfu.ca/pkp-policy-ai/) and check that this complies.

PR #5335, opened 2026-02-12
pkp/pkp-lib#9921 Submodule update ##nibou230/9921## 

nibou230:
For pkp/pkp-lib#9921

I had to recompile the minified JS, not sure this part is perfect even if I used the "buildjs.sh" tool provided.


defstat (changes requested):
Thanks @nibou230! 

Should the `pkp.min.js` file be committed for this (also tagging @jardakotesovec for the good practice for this). I think that it should not.

Also, shouldn't we consider OMP for this change?

jardakotesovec:
@nibou230 @defstat Sorry for delay. pkp.min.js gets rarely changed these days as we build things in vue.js. But we do keep this file in the repo and therefore is important to update it. So I think you can go ahead with it.

PR #5378, opened 2026-02-28
pkp/pkp-lib#12392 Support Funder Data

ajnyga:
https://github.com/pkp/pkp-lib/issues/12392

### DIFF
--- a/api/v1/submissions/index.php
+++ b/api/v1/submissions/index.php
@@ -7,8 +7,8 @@
 /**
  * @file api/v1/submissions/index.php
  *
- * Copyright (c) 2023 Simon Fraser University
- * Copyright (c) 2023 John Willinsky
+ * Copyright (c) 2023-2026 Simon Fraser University
+ * Copyright (c) 2023-2026 John Willinsky
  * Distributed under the GNU GPL v3. For full terms see the file docs/COPYING.
  *
  * @ingroup api_v1_submissions
@@ -29,6 +29,10 @@ if (in_array('jats', $urlParts)) {
 
 if (in_array('dataCitations', $urlParts)) {
     return new \PKP\handler\APIHandler(new \PKP\API\v1\dataCitations\PKPDataCitationController());
+}
+
+if (in_array('funders', $urlParts)) {
+    return new \PKP\handler\APIHandler(new \PKP\API\v1\funders\PKPFunderController());
 }
 
 if (count($urlParts) >= 8 && $urlParts[7] == 'bodyText') {
--- a/dbscripts/xml/upgrade.xml
+++ b/dbscripts/xml/upgrade.xml
@@ -232,6 +232,7 @@
         <migration class="PKP\migration\upgrade\v3_6_0\I12799_MovePublishedSubmissionsToDone"/>
 		<migration class="APP\migration\upgrade\v3_6_0\I12948_RemoveRFC1807Plugin" />
 		<migration class="PKP\migration\upgrade\v3_5_0\I12133_FixCategorySort"/>
+		<migration class="PKP\migration\upgrade\v3_6_0\I12392_Funders" />
 		<code function="rebuildSearchIndex" />
 		<note file="docs/release-notes/README-3.6.0" />
 	</upgrade>
--- a/plugins/generic/datacite/filter/DataciteXmlFilter.php
+++ b/plugins/generic/datacite/filter/DataciteXmlFilter.php
@@ -261,6 +261,11 @@ class DataciteXmlFilter extends \PKP\plugins\importexport\native\filter\NativeEx
         if ($relatedItemsNode) {
             $rootNode->appendChild($relatedItemsNode);
         }
+        // Funding references
+        $fundingReferencesNode = $this->createFundingReferencesNode($doc, $publication);
+        if ($fundingReferencesNode) {
+            $rootNode->appendChild($fundingReferencesNode);
+        }
 
         return $doc;
     }
@@ -638,6 +643,100 @@ class DataciteXmlFilter extends \PKP\plugins\importexport\native\filter\NativeEx
         } else {
             return null;
         }
+    }
+
+    /**
+     * Create funding references node.
+     */
+    public function createFundingReferencesNode(DOMDocument $doc, Publication $publication): ?DOMNode
+    {
+        /** @var DataciteExportDeployment $deployment */
+        $deployment = $this->getDeployment();
+
+        $funders = $publication->getData('funders');
+        if (empty($funders)) {
+            return null;
+        }
+
+        $locale = $publication->getData('locale');
+        $fundingReferencesNode = $doc->createElementNS($deployment->getNamespace(), 'fundingReferences');
+
+        foreach ($funders as $funder) {
+            $funderName = $funder->getLocalizedData('name', $locale);
+
+            // fundingReference requires a funderName
+            if (empty($funderName)) {
+                continue;
+            }
+
+            $grants = $funder->grants ?? [];
+
+            // Even with no grants we still want an entry for the funder
+            $grantEntries = !empty($grants) ? $grants : [null];
+
+            foreach ($grantEntries as $grant) {
+                $fundingReferenceNode = $doc->createElementNS(
+                    $deployment->getNamespace(),
+                    'fundingReference'
+                );
+
+                $fundingReferenceNode->appendChild(
+                    $doc->createElementNS(
+                        $deployment->getNamespace(),
+                        'funderName',
+                        htmlspecialchars($funderName, ENT_COMPAT, 'UTF-8')
+                    )
+                );
+
+                if (!empty($funder->ror)) {
+                    $funderIdentifierNode = $doc->createElementNS(
+                        $deployment->getNamespace(),
+                        'funderIdentifier',
+                        $funder->ror
+                    );
+                    $funderIdentifierNode->setAttribute('funderIdentifierType', 'ROR');
+                    $fundingReferenceNode->appendChild($funderIdentifierNode);
+                }
+
+                if (!empty($grant)) {
+
+                    // Create awardNumber when either a grant number or grant DOI exists.
+                    // This allows an empty awardNumber with just an awardURI
+                    // to support cases where only grantDOI is given.
+                    if (!empty($grant['grantNumber']) || !empty($grant['grantDoi'])) {
+
+                        $awardNumberNode = $doc->createElementNS(
+                            $deployment->getNamespace(),
+                            'awardNumber',
+                            htmlspecialchars($grant['grantNumber'] ?? '', ENT_COMPAT, 'UTF-8')
+                        );
+
+                        if (!empty($grant['grantDoi'])) {
+                            $awardNumberNode->setAttribute(
+                                'awardURI',
+                                htmlspecialchars($grant['grantDoi'], ENT_COMPAT, 'UTF-8')
+                            );
+                        }
+
+                        $fundingReferenceNode->appendChild($awardNumberNode);
+                    }
+
+                    if (!empty($grant['grantName'])) {
+                        $fundingReferenceNode->appendChild(
+                            $doc->createElementNS(
+                                $deployment->getNamespace(),
+                                'awardTitle',
+                                htmlspecialchars($grant['grantName'], ENT_COMPAT, 'UTF-8')
+                            )
+                        );
+                    }
+                }
+
+                $fundingReferencesNode->appendChild($fundingReferenceNode);
+            }
+        }
+
+        return $fundingReferencesNode->hasChildNodes() ? $fundingReferencesNode : null;
     }
 
     /**
--- a/templates/frontend/objects/article_details.tpl
+++ b/templates/frontend/objects/article_details.tpl
@@ -475,6 +475,38 @@
 				</div>
 			{/if}
 
+			{* Funders *}
+			{if $publication->getData('funders')}
+			<section class="item funders" id="funding-data">
+				<h2 class="label">
+					{translate key="submission.funders"}
+				</h2>
+				<div class="value">
+					<ul>
+						{foreach from=$publication->getData('funders') item=funder}
+							<li>
+								<span class="funder">
+									{$funder->getLocalizedData('name')|escape}
+									{if $funder->ror}<a href="{$funder->ror|escape}">{$rorIdIcon}</a>{/if}
+								</span>
+								{if $funder->grants}
+									<ul>
+										{foreach from=$funder->grants item=grant}
+											<li>
+												{if $grant.grantName}<span>{$grant.grantName|escape}</span><br />{/if}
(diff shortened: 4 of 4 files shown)


api/v1/submissions/index.php
```diff
@@ -7,8 +7,8 @@
 /**
  * @file api/v1/submissions/index.php
  *
- * Copyright (c) 2023 Simon Fraser University
- * Copyright (c) 2023 John Willinsky
+ * Copyright (c) 2026 Simon Fraser University
```
kaitlinnewson: This should be `2023-2026` since it's an update to an existing file.

plugins/generic/datacite/filter/DataciteXmlFilter.php
```diff
@@ -794,6 +801,80 @@ public function createRelatedItemsNode(DOMDocument $doc, Issue $issue, Submissio
         return $relatedItemsNode;
     }
 
+    /**
+     * Create fundingReferences node.
+     */
+    public function createFundingReferencesNode(DOMDocument $doc, Issue $issue, Submission $article, ?Publication $publication, ?Galley $galley): ?DOMNode
```
bozana: If we are exporting an issue object Submission can be null, so we need to use `?Submission`

plugins/generic/datacite/filter/DataciteXmlFilter.php
```diff
@@ -794,6 +801,80 @@ public function createRelatedItemsNode(DOMDocument $doc, Issue $issue, Submissio
...
+                    if (!empty($grant['grantNumber'])) {
+                        $fundingReferenceNode->appendChild(
+                            $doc->createElementNS($deployment->getNamespace(), 'awardNumber', htmlspecialchars($grant['grantNumber'], ENT_COMPAT, 'UTF-8'))
+                        );
+                    }
+
+                    $fundingReferencesNode->appendChild($fundingReferenceNode);
+                }
+            }
+        }
```
bozana: Instead of having if... else... and repeating the same code twice, could we maybe do something like this:
```
$i = 0;                                                                                                                                                                
  do {                                                                                                                                                                   
      $grant = $grants[$i] ?? null;                                                                                                                                      
                                                                                                                                                                         
      $fundingReferenceNode = $doc->createElementNS($deployment->getNamespace(), 'fundingReference');                                                                    
      $fundingReferenceNode->appendChild(                                                                                                                                
          $doc->createElementNS($deployment->getNamespace(), 'funderName', htmlspecialchars($funderName, ENT_COMPAT, 'UTF-8'))                                           
      );                                                                                                                                                                 
   
      if (!empty($funder->ror)) {                                                                                                                                        
          $funderIdentifierNode = $doc->createElementNS($deployment->getNamespace(), 'funderIdentifier', $funder->ror);
          $funderIdentifierNode->setAttribute('funderIdentifierType', 'ROR');                                                                                            
          $fundingReferenceNode->appendChild($funderIdentifierNode);                                                                                                     
      }
                                                                                                                                                                         
      if (!empty($grant['grantNumber'])) {                                                                                                                               
          $fundingReferenceNode->appendChild(
              $doc->createElementNS($deployment->getNamespace(), 'awardNumber', htmlspecialchars($grant['grantNumber'], ENT_COMPAT, 'UTF-8'))                            
          );                                                                                                                                                             
      }
                                                                                                                                                                         
      $fundingReferencesNode->appendChild($fundingReferenceNode);
      $i++;
  } while ($i < count($grants));
```

plugins/generic/datacite/filter/DataciteXmlFilter.php
```diff
@@ -794,6 +801,80 @@ public function createRelatedItemsNode(DOMDocument $doc, Issue $issue, Submissio
...
+
+                    if (!empty($funder->ror)) {
+                        $funderIdentifierNode = $doc->createElementNS($deployment->getNamespace(), 'funderIdentifier', $funder->ror);
+                        $funderIdentifierNode->setAttribute('funderIdentifierType', 'ROR');
+                        $fundingReferenceNode->appendChild($funderIdentifierNode);
+                    }
+
+                    if (!empty($grant['grantNumber'])) {
+                        $fundingReferenceNode->appendChild(
+                            $doc->createElementNS($deployment->getNamespace(), 'awardNumber', htmlspecialchars($grant['grantNumber'], ENT_COMPAT, 'UTF-8'))
```
bozana: Here we can see if grant DOI exist and provide it as awardURI attribute, s. note for awardURI here https://datacite-metadata-schema.readthedocs.io/en/4.7/properties/fundingreference/#a-awarduri.

plugins/generic/datacite/filter/DataciteXmlFilter.php
```diff
@@ -794,6 +801,80 @@ public function createRelatedItemsNode(DOMDocument $doc, Issue $issue, Submissio
...
+            return null;
+        }
+
+        $locale = $publication->getData('locale');
+
+        $fundingReferencesNode = $doc->createElementNS($deployment->getNamespace(), 'fundingReferences');
+
+        foreach ($funders as $funder) {
+
+            $funderName = $funder->name[$locale] ?? null;
```
bozana: Maybe just to be on the very safe side to use `$funderName = $funder->getLocalizedData('name', $locale);` here -- this way getLocalizedData will fall back to any found name, I think, in case it does not find one in publication locale. This should actually not happen, but just in case? -- I think there also should be no funder without the name, but...

templates/frontend/objects/article_details.tpl
```diff
@@ -449,6 +449,40 @@
...
+									{if $funder->ror}<a href="{$funder->ror|escape}">{$rorIdIcon}</a>{/if}
+								</span>
+								{if $funder->grants}
+									{foreach from=$funder->grants item=grant}
+										<br />
+										{if $grant.grantNumber}
+											{$grant.grantNumber|escape}
+										{/if}
+										{if $grant.grantName}
+											{if $grant.grantNumber}, {/if}
```
bozana: I think we should use `{translate key="common.commaListSeparator"}` instead of `,`.

templates/frontend/objects/article_details.tpl
```diff
@@ -449,6 +449,40 @@
...
+								<span class="funder">									
+									{$funder->getLocalizedData('name')|escape}
+									{if $funder->ror}<a href="{$funder->ror|escape}">{$rorIdIcon}</a>{/if}
+								</span>
+								{if $funder->grants}
+									{foreach from=$funder->grants item=grant}
+										<br />
+										{if $grant.grantNumber}
+											{$grant.grantNumber|escape}
+										{/if}
```
bozana: What about grant DOI?
Shell we maybe have something like: 
```
{if $grant.grantNumber}
      {if $grant.grantDoi}<a href="{$grant.grantDoi|escape}">{$grant.grantNumber|escape}</a>{else}{$grant.grantNumber|escape}{/if}
{/if} 
```
Not necessarily, but would like to hear what you think...

bozana (commented):
Hi @ajnyga, a few comments/questions here...
Thanks a lot!

plugins/generic/datacite/filter/DataciteXmlFilter.php
ajnyga:
```
                if (!empty($grant['grantNumber'])) {
                    $awardNumberNode = $doc->createElementNS($deployment->getNamespace(), 'awardNumber', htmlspecialchars($grant['grantNumber'], ENT_COMPAT, 'UTF-8'));
                    if (!empty($grant['grantDoi'])) {
                        $awardNumberNode->setAttribute('awardURI', htmlspecialchars($grant['grantDoi'], ENT_COMPAT, 'UTF-8'));
                    }
                    $fundingReferenceNode->appendChild($awardNumberNode);
                }
```

ajnyga:
Thanks both, all comments taken into account.

templates/frontend/objects/article_details.tpl
```diff
@@ -471,6 +471,40 @@
...
+
+							<li>
+								<span class="funder">									
+									{$funder->getLocalizedData('name')|escape}
+									{if $funder->ror}<a href="{$funder->ror|escape}">{$rorIdIcon}</a>{/if}
+								</span>
+								{if $funder->grants}
+									{foreach from=$funder->grants item=grant}
+										<br />
+										{if $grant.grantNumber}
```
kaitlinnewson: I think this won't work in some cases - should you also check `{if $grant.grantDoi}` in case there is a DOI with no grant number?

templates/frontend/objects/article_details.tpl
```diff
@@ -471,6 +471,40 @@
...
+											{if $grant.grantDoi}<a href="{$grant.grantDoi|escape}">{$grant.grantNumber|escape}</a>{else}{$grant.grantNumber|escape}{/if}
```
kaitlinnewson: Since the DOI is stored without the URL, the link won't be resolvable here.

templates/frontend/objects/article_details.tpl
```diff
@@ -471,6 +471,40 @@
...
+					<ul>
+						{foreach from=$publication->getData('funders') item=funder}
+
+							<li>
+								<span class="funder">									
+									{$funder->getLocalizedData('name')|escape}
+									{if $funder->ror}<a href="{$funder->ror|escape}">{$rorIdIcon}</a>{/if}
+								</span>
+								{if $funder->grants}
+									{foreach from=$funder->grants item=grant}
```
kaitlinnewson:
```suggestion
									{foreach from=$funder->grants item=grant}
									{strip}
```

templates/frontend/objects/article_details.tpl
```diff
@@ -471,6 +471,40 @@
...
+										<br />
+										{if $grant.grantNumber}
+											{if $grant.grantDoi}<a href="{$grant.grantDoi|escape}">{$grant.grantNumber|escape}</a>{else}{$grant.grantNumber|escape}{/if}
+										{/if}
+										{if $grant.grantName}
+											{if $grant.grantNumber}{translate key="common.commaListSeparator"}{/if}
+											{$grant.grantName|escape}
+										{/if}
+									{/foreach}
```
kaitlinnewson:
```suggestion
									{/strip}
									{/foreach}
```

templates/frontend/objects/article_details.tpl
ajnyga: grantNumber is required, so if there is no granNumber there can be no grantDoi. I guess the whole check could be removed.

templates/frontend/objects/article_details.tpl
ajnyga: yeah, I mean not sure if it even makes sense to link it here in any case

ajnyga: so left the link there, added the missing prefix

templates/frontend/objects/article_details.tpl
kaitlinnewson: I was able to create a grant with just the grant DOI and no grant number.

ajnyga: ah true, I've dropped that check at some point, maybe trying to debug something. I made grantNumber now required, but I guess it is open to argumentation whether it should be required? @bozana thoughts?

ajnyga: If we want to have a check that requires either a DOI or a number, the I do not think the schema validation hacks it? I now have

```
			"items": {
				"type": "object",
				"properties": {
					"grantDoi": {
						"type": "string",
                        "validation": [
                            "regex:/^\\d+(.\\d+)+\\//"
                        ]
					},
					"grantNumber": {
						"type": "string",
						"validation": [
							"required"
						]
					},
					"grantName": {
						"type": "string"
					}
				}
			}
```

templates/frontend/objects/article_details.tpl
```diff
@@ -471,6 +471,38 @@
...
+									{if $funder->ror}<a href="{$funder->ror|escape}">{$rorIdIcon}</a>{/if}
+								</span>
+								{if $funder->grants}
+									<ul>
+										{foreach from=$funder->grants item=grant}
+											<li>
+												{if $grant.grantName}<span>{$grant.grantName|escape}</span><br />{/if}
+												{if $grant.grantNumber}<span>{translate key="submission.funders.funder.grant.number"} {$grant.grantNumber|escape}</span><br />{/if}
+												{if $grant.grantDoi}<span>{translate key="submission.funders.funder.grant.doi"} <a href="https://doi.org/{$grant.grantDoi|escape}">https://doi.org/{$grant.grantDoi|escape}</a></span>{/if}
+											</li>
```
ajnyga, line 499: @kaitlinnewson After thinking about how to visually show these different parts of grants I opted for listing them and show each element individually if available. It is not stylized, but I think that this is the best approach. Just having for example the Grant Number without any explanation just raises questions.

templates/frontend/objects/article_details.tpl
bozana: I think as until now in the funding plugin the grant numbers nor grant DOIs need to be required. A user might not know them and can only supply the funder (ROR or name). No?

templates/frontend/objects/article_details.tpl
bozana: I think we should ask Devika how to display all information we now have. This is however not a stopper for merging...
p.s. We do not consider the RORs that we now have?

templates/frontend/objects/article_details.tpl
ajnyga: Ah sure, I was not thinking the schema itself and just thought with the logic that if someone starts to add a grant we would require some minimum information. Or they could just leave the grants empty.

templates/frontend/objects/article_details.tpl
ajnyga: ROR is displayed with an icon if one exists. The affiliations had this approach so I used that. I am not sure using an icon like this is the best option though.

But agreed that I do not want to go deeply to making visual decisions here. I was thinking more that this kind of data is probably something that will require a structured display because the grant numbers for example can look very different and a number itself does not tell much to readers.

ajnyga: Ok so did quick mockups on how the data could be presented and let's see what Devika says. I can include both funder and data citations display logic with a separate pr when the designs are done.

dbscripts/xml/upgrade.xml
```diff
@@ -193,6 +193,7 @@
 		<migration class="PKP\migration\upgrade\v3_6_0\I10403_EmailTemplateUserGroupAccess"/>
 		<migration class="APP\migration\install\ReviewerRecommendationsMigration" />
 		<migration class="APP\migration\upgrade\v3_6_0\I1660_ReviewerRecommendations"/>
+		<migration class="PKP\migration\upgrade\v3_6_0\I12392_Funders" />
```
bozana: could we please move it to the end of the list of other migration scripts -- this makes it easier follow what one still needs to run, when working with main...

bozana (commented):
Hi @ajnyga, could we position the migration script here and in OMP and OPS at the end of the list?
Thanks a lot"!

PR #5476, opened 2026-04-07
pkp/pkp-lib#12532 job runner flush buffer and cross request locking

touhidurabir:
for pkp/pkp-lib#12532 .


asmecher:
Merged manually to resolve submodule commits.

PR #5483, opened 2026-04-08
Set lensGalley submodule to track updated stable-3_3_0 branch

wopsononock:
Tested this on a fresh install and it allows previously broken viewing of figures and tables in the XML article viewer.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,7 +22,7 @@
 [submodule "plugins/generic/lensGalley"]
 	path = plugins/generic/lensGalley
 	url = https://github.com/asmecher/lensGalley
-	branch = i1472-fix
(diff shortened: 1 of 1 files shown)


ctgraham:
@asmecher , @wopsononock and I looked at the failing tests, and we don't believe they are related to the proposed change.

asmecher:
Thanks, I've merged this, made similar updates to `stable-3_4_0` and `stable-3_5_0` and edited the `main` branch to remove the `branch` specifier entirely. (We don't generally specify branches on submodules -- the commits speak for themselves. And it's likely to get forgotten going forward when we e.g. split off `stable-3_6_0`.)

PR #5492, opened 2026-04-14
pkp/pkp-lib#12555 fix max version for 3.5.0.x upgrade scripts

bozana:
s. https://github.com/pkp/pkp-lib/issues/12555

### DIFF
--- a/dbscripts/xml/upgrade.xml
+++ b/dbscripts/xml/upgrade.xml
@@ -178,12 +178,10 @@
 		<note file="docs/release-notes/README-3.5.0" />
 	</upgrade>
 
-	<upgrade minversion="3.5.0.0" maxversion="3.5.0.1">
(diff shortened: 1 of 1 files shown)


farrisimin:
We would like an official release for 3.5? we have been wait for quite some time. .😩

asmecher:
@farrisimin, I'm not sure I understand what you're asking for (or that this is the right place for it); can you say more?

farrisimin:
We are currently running 3.4.8 and attempting to upgrade to 3.5.4 but the upgrade is not going through.

asmecher:
@farrisimin, I think you mean 3.4.0-8; there is no 3.4.8.  Could you create a post with details (including the problem/error message and which of our applications you're using -- OJS, OMP, or OPS) on the support forum (https://forum.pkp.sfu.ca)? That's the best way to start debugging.

PR #5525, opened 2026-04-29
pkp/pkp-lib#12584 Add "Summary of Changes/Amendment" field on file revision upload and update publication issues page

blesildaramirez:


schemas/publication.json
```diff
@@ -59,6 +59,15 @@
 		"sectionId": {
 			"type": "integer",
 			"apiSummary": true
+		},
+		"updateType": {
+			"type": "string",
```
jardakotesovec: Ok I see you are adding in here..   Since Bozana confirmed that Amendment make sense for all apps - I think both mibration and schema updates can go to pkp-lib. And we can just separately figure out where exactly to add it for OPS/OMP in the UI.

pages/dashboard/DashboardHandler.php
```diff
@@ -56,6 +57,15 @@ public function setupIndex($request)
...
             ->getCount();
 
+        $updateTypeOptions = [];
+        foreach (UpdateType::cases() as $updateType) {
+            $updateTypeOptions[] = [
+                'label' => $updateType->label(),
+                'value' => $updateType->value,
+            ];
+        }
+        $pageInitConfig['componentForms']['updateTypeOptions'] = $updateTypeOptions;
```
jardakotesovec, line 67: I see we have this pattern already used for VersionStage. Lets think about some alternatives on wednesday. I would like to keep these data snippets being passed on page load to minimum as we are aiming for higher independency between frontend and backend. Will need to think about more about solid approach. But for this PR keep as is.

jardakotesovec (approved):

schemas/publication.json
blesildaramirez: Hi Jarda, I moved this change to pkp-lib as suggested offline. Thanks!

classes/migration/install/OJSMigration.php
```diff
@@ -245,6 +246,7 @@ public function up(): void
             $table->enum('version_stage', array_column(VersionStage::cases(), 'value'))->nullable();
             $table->integer('version_minor')->nullable();
             $table->integer('version_major')->nullable();
+            $table->enum('update_type', array_column(UpdateType::cases(), 'value'))->nullable();
```
bozana: Here as well, no ->nullable, but ->default(UpdateType::NEW_VERSION->value).

jardakotesovec:
merged separately without submodule update so its possible to squash it

PR #5530, opened 2026-05-04
pkp/pkp-lib#12547 Session/cookie management update

touhidurabir:
for pkp/pkp-lib#12547 .

### DIFF
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -45,8 +45,9 @@ session_cookie_name = OJSSID
 ; Session cookie path; if not specified, defaults to the detected base path
 ; session_cookie_path = /
 
-; Number of days to save login cookie for if user selects to remember
-; (set to 0 to force expiration at end of current session)
+; Number of days a session remains valid while idle (minimum 1 day)
+; Values less than 1 are clamped to 1 day
+; To expire sessions when the browser closes, use session_expire_on_close in [security]
 session_lifetime = 30
 
 ; SameSite configuration for the cookie, see possible values and explanations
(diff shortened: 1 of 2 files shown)


config.TEMPLATE.inc.php
```diff
@@ -45,8 +45,9 @@
 ; Session cookie path; if not specified, defaults to the detected base path
 ; session_cookie_path = /
 
-; Number of days to save login cookie for if user selects to remember
-; (set to 0 to force expiration at end of current session)
+; Number of days a session remains valid while idle (minimum 1 day)
+; Values less than 1 are clamped to 1 day
```
taslangraham:
```suggestion
; Values less than 1 will default to 1 day
```

taslangraham (changes requested):
Just the one minor suggestion. Looks good otherwise.

bozana (approved):
Hi @touhidurabir, all fine here

config.TEMPLATE.inc.php
```diff
@@ -45,9 +45,10 @@
 ; Session cookie path; if not specified, defaults to the detected base path
 ; session_cookie_path = /
 
-; Number of days to save login cookie for if user selects to remember
-; (set to 0 to force expiration at end of current session)
-session_lifetime = 30
+; Number of days a session remains valid while idle. Fractional days are allowed
+; (e.g. 0.5 = 12 hours, 0.0833 ~ 2 hours); the value is clamped to a minimum of 1 minute.
```
bozana, line 49: 1 minute or rather 1 day?

touhidurabir: @bozana I think the confusion is coming from `the value is clamped to a minimum of 1 minute.` . As we are allowing fraction value for days (e.g. 0.5 = 12 hours, 0.0833 ~ 2 hours) so in code level we clamp it to 1 min as if someone mistakenly put a value become 0, at login they will be immediately logged out . so thats why that comment . 

if seems confusing, I can remove it line .

taslangraham (approved):

config.TEMPLATE.inc.php
bozana: No, that is fine, I am just wandering why you decided to use min 1 min rather than 1 day as you originally thought

bozana (approved):

PR #5536, opened 2026-05-11
pkp/pkp-lib#11914 Migrate Credit Plugin data

ajnyga:
https://github.com/pkp/pkp-lib/issues/11914


bozana (commented):
Here the upgrade.xml needs to be adapted, to contain the new migration script.

PR #5538, opened 2026-05-12
pkp/pkp-lib#12617 Don't let crossref validation block publishing

taslangraham:
For pkp/pkp-lib#12617


classes/components/forms/publication/PublishForm.php
```diff
@@ -56,17 +56,35 @@ public function __construct($action, $publication, $submissionContext, $requirem
...
         // Set separate messages and buttons if publication requirements have passed
         if (empty($requirementErrors)) {
+            // Get any warnings that should be displayed when publishing.
+            $submission = Repo::submission()->get($publication->getData('submissionId'));
+            $warnings = Repo::publication()->validatePublishWarnings(
+                $publication,
+                $submission,
+                $submissionContext->getSupportedSubmissionLocales(),
+                $submissionContext->getPrimaryLocale()
+            );
```
bozana: Submission is fetched here for every normal publish confirmation dialog -- there is now an extra DB query even when the hook fires no warnings. It is performance regression, that is not needed I think.
See my comment below for a possible solution.

classes/components/forms/publication/PublishForm.php
```diff
@@ -56,17 +56,35 @@ public function __construct($action, $publication, $submissionContext, $requirem
...
+
+            if (!empty($warnings)) {
+                $msg = '<div class="pkpNotification"><b>' . __('publication.publish.warning') . '</b>';
+                $msg .= '<ul>';
+                foreach ($warnings as $error) {
+                    $msg .= '<li>' . $error . '</li>';
+                }
+                $msg .= '</ul></div><br>';
+            }
```
bozana: This rendering task should actually not go into the PHP form class.
See my comment below for a possible solution.

classes/components/forms/publication/PublishForm.php
```diff
@@ -56,17 +56,35 @@ public function __construct($action, $publication, $submissionContext, $requirem
...
+
```
bozana: I think PublishForm does not need to know about the warnings. 
The PublishHadler in pkp-lib could call 
`$warnings = Repo::publication()->validatePublishWarnings($this->publication, $this->submission, $allowedLocales, $primaryLocale);` 
and provide the `$warnings` to the publish template (publish.tpl): 
`$templateMgr->assign('publishWarnings', $warnings);`. 

This way, the template renders the warning block in Smarty above the form — HTML is there where it belongs, using `{foreach}` over the plain array and maybe the `pkpNotification--warning` class that already exists, e.g.:
```
{if $publishWarnings}
  <div class="pkpNotification pkpNotification--warning">
  <strong>{translate key="publication.publish.warning"}</strong>
  <ul>
  {foreach from=$publishWarnings item=warning}
  <li>{$warning}</li>
  {/foreach}
  </ul>
  </div>
{/if}
```

What do you think?

bozana (commented):
Hi @taslangraham, a few comments i.e. a suggestion how to maybe solve it slightly differently. What do you think?

classes/components/forms/publication/PublishForm.php
taslangraham: Thanks for the suggestion, Bozana. I initially thought of doing it this way, but then decided to make the warnings a part of the form, as the normal validation Errors were already being rendered with HTML in the PublishForm. However, your suggestion, along with the performance issue you pointed out in this [comment](https://github.com/pkp/ojs/pull/5538/changes#r3340591466), makes sense. I'll update the code accordingly. Thanks!

taslangraham:
Hey @bozana, thanks for the review. I've made the suggested [changes](https://github.com/pkp/ojs/pull/5538#discussion_r3340600788) to the pkp-lib [PR](https://github.com/pkp/pkp-lib/pull/12717). Please take a look.

PR #5559, opened 2026-05-28
pkp/pkp-lib#12031 update swagger docs for 3.5

kaitlinnewson:
pkp/pkp-lib#12031

### DIFF
--- a/docs/dev/swagger-source.json
+++ b/docs/dev/swagger-source.json
@@ -1,7 +1,7 @@
 {
 	"swagger": "2.0",
 	"info": {
-		"version": "3.4",
+		"version": "3.5",
 		"title": "REST API - OJS",
 		"description": "This document describes how to interact with the REST API in [Open Journal Systems](https://pkp.sfu.ca/software/ojs). It is a technical document intended for software developers who wish to build applications to interact with OJS. This document only describes the REST API for OJS. However, similar REST APIs exist for [Open Monograph Press](https://pkp.sfu.ca/software/omp) and [Open Preprint Systems](https://pkp.sfu.ca/software/ops).\n\nView the [Developer Documentation Hub](https://docs.pkp.sfu.ca/dev) to view all documentation for these applications.",
 		"contact": {
@@ -20,7 +20,7 @@
 		{
 			"name": "Usage",
 			"tags": [
-				"What's New in 3.4",
+				"What's New in 3.5",
 				"Access",
 				"Authentication",
 				"Pagination",
@@ -35,21 +35,27 @@
 				"Submissions",
 				"Submissions / Decisions",
 				"Submissions / Files",
+				"Submissions / JATS",
 				"Submissions / Participants",
 				"Submissions / Publications",
 				"Submissions / Publications / Contributors",
+				"Submissions / Reviewer Suggestions",
 				"Submissions / DashboardViews",
 				"Issues",
 				"Users",
 				"Contexts",
 				"Site",
+				"Sections",
 				"Announcements",
 				"DOIs",
 				"DOIs / Submissions",
 				"DOIs / Issues",
 				"Email Templates",
 				"Institutions",
+				"Invitations",
 				"Mailables",
+				"Reviews",
+				"RORs",
 				"Stats / Contexts",
 				"Stats / Editorial",
 				"Stats / Issues",
@@ -66,9 +72,9 @@
 	],
 	"tags": [
 		{
-			"name": "What's New in 3.4",
+			"name": "What's New in 3.5",
 			"x-traitTag": true,
-			"description": "<blockquote style=\"margin-bottom:2rem;padding:1rem;border:2px solid #0073A8;color: #002c40\">Read the <a href=\"https://docs.pkp.sfu.ca/dev/release-notebooks/en/3.4-release-notebook\">3.4 Release Notebook</a> to learn about all the changes introduced in 3.4.</blockquote>\n\nThe following changes were made to the REST API.\n\n- Use of the `apiToken` as a query parameter is deprecated. Use it in the `Authorization` header instead. See the \"Authentication\" section below.\n- Changes to request body to create a new submission.\n- Changes to submission titles. They may now include HTML in the `fullTitle`, `title`, and `subtitle` properties.\n- Changes to the submission property, `submissionProgress`. It is now a string, not an integer.\n- Changes to the stats endpoints.\n- Changes to the endpoints to manage email templates.\n\nThe following new endpoints have been added.\n\n- New endpoint to save a submission for later.\n- New endpoint to submit a submission.\n- New endpoint to copy a submission file from one file stage to another.\n- New endpoints to manage a publication's contributors.\n- New endpoints to create and register DOIs.\n- New endpoints to record editorial decisions.\n- New endpoints to manage institutions.\n- New endpoints to get mailables."
+			"description": "<blockquote style=\"margin-bottom:2rem;padding:1rem;border:2px solid #0073A8;color: #002c40\">Read the <a href=\"https://docs.pkp.sfu.ca/dev/release-notebooks/en/3.5-release-notebook\">3.5 Release Notebook</a> to learn about all the changes introduced in 3.5.</blockquote>\n\nThe following major changes have been made to the REST API.\n\n- The backend submissions handler (`_submissions`) was extended to support refactored submission listings and review assignments.\n- API handling has been <a href=\"https://github.com/pkp/pkp-lib/issues/7698#issue-1132905079\">migrated from the Slim microframework to Laravel</a>. Third-party additions/modifications to the API that made use of Slim features must be migrated to Laravel's routing toolkit. Handler classes have been renamed from `*Handler` to `*Controller`.\n- Plugins wishing to modify the handling of requests to existing API endpoints should follow <a href=\"https://github.com/pkp/pkp-lib/issues/9434\">a new hook-based pattern</a>.\n\nThe following schema and behavioral changes also affect API consumers.\n\n- Authors can now have <a href=\"https://github.com/pkp/pkp-lib/issues/7135\">multiple institutional affiliations</a>, each optionally linked to a [ROR](https://ror.org/) record. The `affiliation` property on contributors has been replaced with an `affiliations` array of structured affiliation objects.\n- The `Submission` schema now includes a `reviewerSuggestions` array (read-only) listing reviewer suggestions made at submission time, as well as `issueToBePublished` and (for OJS) `scheduledIn` properties.\n- The `Galley` `locale` property now accepts BCP 47-style extended locale codes (e.g. `sr_Cyrl@RS`).\n\nThe following new endpoints have been added.\n\n- New endpoints to list and retrieve [sections](#tag/Sections).\n- New endpoints to retrieve and add records from the [Research Organization Registry (ROR)](#tag/RORs) cache, used to support author affiliations.\n- New endpoints to manage <a href=\"https://github.com/pkp/pkp-lib/issues/9197\">invitations</a>, including creating, populating, inviting, cancelling, and accepting or declining email-based invitations. In 3.5 this toolset is used for email address change confirmations, one-click reviewer access, and the GDPR-compliant user account creation process.\n- New endpoints to manage [reviewer suggestions](#tag/Submissions-Reviewer-Suggestions) made by authors during submission.\n- New endpoints to retrieve [review history](#tag/Reviews), confirm review assignments, export individual reviews as PDF or XML, and submit completed reviews to ORCID.\n- New endpoints to retrieve, upload, and delete [JATS XML](#tag/Submissions-JATS) source files attached to a submission.\n- New endpoints to request and revoke [ORCID](#tag/ORCID) verification for individual authors.\n- New endpoints for authors to retrieve [emails](#tag/Emails) sent to them about their submissions.\n- New private editorial-UI endpoint at `/submissions/{submissionId}/publications/{publicationId}/_components/{component}` that exposes the [publication form components](https://github.com/pkp/pkp-lib/issues/10336) (metadata, references, title/abstract, identifiers, permission/disclosure, change-language helper, and OJS-only issue and submission-payment forms).\n- New endpoint `PUT /submissions/{submissionId}/publications/{publicationId}/changeLocale` to change the primary locale of a publication.\n- The DOI registration agency settings endpoint was relocated from `/contexts/editDoiRegistrationAgencyPlugin` to `PUT /contexts/{contextId}/registrationAgency`."
 		},
 		{
 			"name": "Access",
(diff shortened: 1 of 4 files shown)


docs/dev/swagger-source.json
```diff
@@ -110,6 +116,10 @@
 		{
 			"name": "Submissions / Files"
 		},
+		{
+			"name": "Submissions / JATS",
+			"description": "Endpoints to retrieve, upload, and delete the JATS XML source file associated with a submission's current publication."
```
taslangraham:
```suggestion
			"description": "Endpoints for managing the JATS XML source file associated with a submission's current publication."
```

docs/dev/swagger-source.json
```diff
@@ -119,6 +129,10 @@
 		{
 			"name": "Submissions / Publications / Contributors"
 		},
+		{
+			"name": "Submissions / Reviewer Suggestions",
+			"description": "Endpoints to manage reviewer suggestions made by an author when submitting their work. Suggestions can be reviewed and approved by editors and converted into reviewer assignments.\n\nThe `POST`, `PUT`, and `DELETE` endpoints in this group will only succeed while the submission is still incomplete (`submissionProgress` is a non-empty string). Once a submission is complete, suggestions become read-only."
```
taslangraham:
```suggestion
			"description": "Endpoints for managing reviewer suggestions made by an author when submitting their work. Suggestions can be reviewed and approved by editors and converted into reviewer assignments.\n\nThe `POST`, `PUT`, and `DELETE` endpoints in this group will only succeed while the submission is still incomplete (`submissionProgress` is a non-empty string). Once a submission is complete, suggestions become read-only."
```

docs/dev/swagger-source.json
```diff
@@ -161,10 +175,26 @@
...
+			"name": "Reviews",
+			"description": "Endpoints to retrieve review round history, confirm review assignments on behalf of a reviewer, export individual review responses as PDF or XML, and submit completed reviews to ORCID."
+		},
+		{
+			"name": "RORs",
+			"description": "Endpoints to query the local cache of records from the [Research Organization Registry (ROR)](https://ror.org/). These records are used to identify institutions when adding author affiliations."
+		},
+		{
+			"name": "Sections",
+			"description": "Endpoints to retrieve sections (or series) configured in the current context. Most authenticated users can access these endpoints."
```
taslangraham:
```suggestion
			"description": "Endpoints to retrieve sections (or series) configured in the current context (journal, press, or preprint server). Most authenticated users can access these endpoints."
```

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8043,1516 @@
...
+					"200": {
+						"x-summary": "Success",
+						"description": "The JATS XML and metadata for the publication.",
+						"schema": {
+							"$ref": "#/definitions/JatsFile"
+						}
+					},
+					"403": {
+						"x-summary": "Request Unauthorized",
+						"description": "You do not have permission to view JATS XML for this submission.",
```
taslangraham:
```suggestion
						"description": "You do not have permission to view the JATS XML for this submission.",
```

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8043,1516 @@
...
+						"name": "publicationId",
+						"description": "Publication ID",
+						"in": "path",
+						"required": true,
+						"type": "integer"
+					},
+					{
+						"name": "file",
+						"description": "The JATS XML file to attach to this publication.",
+						"in": "formData",
```
taslangraham, line 8830: Should this be `body` instead of `formData`?

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8043,1516 @@
...
+					"400": {
+						"x-summary": "Request Invalid",
+						"description": "No file was provided in the request, or the upload failed.",
+						"schema": {
+							"$ref": "#/definitions/ErrorResponse"
+						}
+					},
+					"403": {
+						"x-summary": "Request Unauthorized",
+						"description": "You do not have permission to upload JATS XML for this submission.",
```
taslangraham:
```suggestion
						"description": "You do not have permission to upload a JATS XML for this submission.",
```

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8043,1516 @@
...
+					},
+					{
+						"name": "offset",
+						"in": "query",
+						"required": false,
+						"type": "integer"
+					},
+					{
+						"name": "status",
+						"description": "Filter by invitation status (e.g. `pending`, `accepted`, `declined`, `cancelled`, `expired`).",
```
taslangraham: Could list all the possible options in an `enum` prop, similar to how it is done in other places of the docs. Example:

```
"enum": [
    "pending",
	"accepted",
	"declined",
    .....
]
```

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8043,1516 @@
...
+			"post": {
+				"tags": [
+					"Invitations"
+				],
+				"summary": "Create a new invitation.",
+				"description": "Create a new invitation of the given `type`. The request body should contain the data required to populate the invitation (such as invitee email, target context id, and role assignments).",
+				"parameters": [
+					{
+						"name": "type",
+						"description": "The invitation type to create.",
```
taslangraham:
```suggestion
						"description": "The type of invitation to create.",
```

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8043,1516 @@
...
+						"name": "invitationId",
+						"in": "path",
+						"required": true,
+						"type": "integer"
+					}
+				],
+				"responses": {
+					"200": {
+						"x-summary": "Success",
+						"description": "The cancelled invitation."
```
taslangraham: The description here suggests that some data is returned; however, looking at the [code](https://github.com/pkp/pkp-lib/blob/0a52f0f935d9d5f487f325845af902fc77427a31/classes/invitation/invitations/userRoleAssignment/handlers/api/UserRoleAssignmentCreateController.php#L160-L171), the response appears to be an empty array.

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8043,1516 @@
...
+					}
+				}
+			}
+		},
+		"/invitations/{invitationId}/key/{key}": {
+			"get": {
+				"tags": [
+					"Invitations"
+				],
+				"summary": "Receive an invitation using its access key.",
```
taslangraham:
```suggestion
				"summary": "Retrieve an invitation using its access key.",
```

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8043,1516 @@
...
+						"description": "The single-use access key sent to the invitee.",
+						"in": "path",
+						"required": true,
+						"type": "string"
+					}
+				],
+				"responses": {
+					"200": {
+						"x-summary": "Success",
+						"description": "The invitation details."
```
taslangraham: Can the shape of the returned invitation be included here?  I believe it is available in pkp-lib at `classes/invitation/invitations/userRoleAssignment/resources/UserRoleAssignmentInviteResource.php`

taslangraham (changes requested):
@kaitlinnewson, I only have a few questions and suggestions. Everything looks good otherwise.

docs/dev/swagger-source.json
kaitlinnewson: In most cases I see in the API docs we don't specify this, but it seems inconsistent throughout - maybe for the API docs we should only qualify this in details related to the `contexts` endpoint?

docs/dev/swagger-source.json
kaitlinnewson: I think when the type is "file", this is correct under the swagger spec. From https://swagger.io/specification/v2/:

> If type is "file", the [consumes](https://swagger.io/specification/v2/#operation-consumes) MUST be either "multipart/form-data", " application/x-www-form-urlencoded" or both and the parameter MUST be [in](https://swagger.io/specification/v2/#parameter-in) "formData".

If my understanding is correct, there may be a couple of more places in the docs to fix as well.

taslangraham: I've read the attached docs, and you're correct. Do you intend to review those other places on this PR, or delegate to a separate issue?

kaitlinnewson: Actually we don't have other cases of `"type": "file"` to correct - there is another case of `multipart/form-data` but it's a little different from this since it has an object reference, so I'll leave that for now.

docs/dev/swagger-source.json
taslangraham: Yeah, that approach makes sense.

docs/dev/swagger-source.json
taslangraham: Thanks for checking on that.

taslangraham (approved):

docs/dev/swagger-source.json
```diff
@@ -119,6 +129,10 @@
 		{
 			"name": "Submissions / Publications / Contributors"
 		},
+		{
+			"name": "Submissions / Reviewer Suggestions",
+			"description": "Endpoints for managing reviewer suggestions made by an author when submitting their work. Suggestions can be reviewed and approved by editors and converted into reviewer assignments.\n\nThe `POST`, `PUT`, and `DELETE` endpoints in this group will only succeed while the submission is still incomplete (`submissionProgress` is a non-empty string). Once a submission is complete, suggestions become read-only."
```
ewhanson: Is the implication here that `submissionProgress` is a non-empty string while the submission is still incomplete? If so, I would throw an `i.e.` at the start of the parenthetical phrase to make it totally clear.

docs/dev/swagger-source.json
```diff
@@ -161,10 +175,26 @@
 			"name": "Institutions",
 			"description": "Only admins and journal managers can access these endpoints."
 		},
+		{
+			"name": "Invitations",
+			"description": "Endpoints used to create, send, and respond to email-based invitations. In 3.5 the invitation toolset is used to request an email address change, to provide one-click reviewer access (replacing the older access-keys mechanism), and to support the GDPR-compliant user account creation process. Some endpoints in this group are publicly accessible (using an invitation key) so that invitees can accept, refine, or decline an invitation without authenticating."
```
ewhanson: I would err on the side of not including specific version information in the description, and just have it be `The invitation toolset [...]`. It's probably best to keep it up-to-date as it change rather than version the description within the text.

docs/dev/swagger-source.json
ewhanson: Another minor note on the OJS vs OMP/OPS applicability: is `series` applicable to OJS at all? I know we generally try to make the docs generic, but they are specific to OJS, I would remove reference to series if that's an OMP-specific aspect.

docs/dev/swagger-source.json
```diff
@@ -344,6 +375,204 @@
...
+				"name": {"en": "jats-23-20250612-101530.xml"},
+				"fileStage": 30,
+				"mimetype": "application/xml",
+				"jatsContent": "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<article>...</article>",
+				"isDefaultContent": false
+			}
+		},
+		"InvitationUser": {
+			"type": "object",
+			"description": "A summary of a user associated with an invitation. Returned for both an existing matched user and the prospective new user built from the invitation payload.",
```
ewhanson: Another nit-picky detail, so feel free to take-or-leave it: `summary` implies to me that there's a non-summary version of this.

docs/dev/swagger-source.json
```diff
@@ -575,6 +804,78 @@
...
+				},
+				"givenName": {
+					"$ref": "#/definitions/LocaleObject"
+				},
+				"fullName": {
+					"description": "The reviewer's full name (derived from `givenName` and `familyName`).",
+					"readOnly": true,
+					"$ref": "#/definitions/LocaleObject"
+				},
+				"displayInitial": {
```
ewhanson: I don't have a concrete answer, but there's some discrepancy between `displayInitial` and `displayInitials` throughout the codebase. This fits with the actual value used in the code, but it seems like it should be plural throughout. Not sure if it's worth changing as part of this, though...

docs/dev/swagger-source.json
```diff
@@ -575,6 +804,78 @@
...
+				"approvedAt": {
+					"type": "string",
+					"description": "The date the suggestion was approved by an editor (ISO 8601 date-time, or `null` if not approved)."
+				},
+				"existingUserId": {
+					"type": "integer",
+					"description": "If the suggested reviewer matches an existing user in the system, the user ID of that user.",
+					"readOnly": true
+				},
+				"existingReviewerRole": {
```
ewhanson: Again, not sure if this out of scope, but it seems like this should be something like `hasExistingReviewerRole` because to me, `existingReviewerRole` implies the filed will have the role/role ID, not whether there is a role present.

docs/dev/swagger-source.json
```diff
@@ -890,7 +1192,7 @@
...
+				"description": "* Unpublished submissions are only returned for assigned participants, editors and admins.\n* `abstract` includes HTML markup.\n* `pages` is usually a `string`. But in some cases it may return an array of arrays. Each array contains two items: the start and end page. Example: `[[25, 31], [45, 62]]`.\n* `submissionProgress` will be a non-empty string if the submission is incomplete.\n* `stages`, `reviewRounds` and `reviewAssignments` should only be returned for participants assigned to the submission.\n* `reviewAssignments` should only expose user-identifying information for editors and admins.\n* `reviewerSuggestions` lists reviewer suggestions made by the author at submission time. This array is read-only here; manage suggestions via the `/submissions/{submissionId}/reviewers/suggestions` endpoints.\n* `scheduledIn` is the ID of the issue this submission has been scheduled into, when applicable.\n* `issueToBePublished` is a summary of the issue this submission will appear in once it is published.",
```
ewhanson:
```suggestion
				"description": "* Unpublished submissions are only returned for assigned participants, editors and admins.\n* `abstract` includes HTML markup.\n* `pages` is usually a `string`, but in some cases it may return an array of arrays. Each array contains two items: the start and end page. Example: `[[25, 31], [45, 62]]`.\n* `submissionProgress` will be a non-empty string if the submission is incomplete.\n* `stages`, `reviewRounds` and `reviewAssignments` should only be returned for participants assigned to the submission.\n* `reviewAssignments` should only expose user-identifying information for editors and admins.\n* `reviewerSuggestions` lists reviewer suggestions made by the author at submission time. This array is read-only here; manage suggestions via the `/submissions/{submissionId}/reviewers/suggestions` endpoints.\n* `scheduledIn` is the ID of the issue this submission has been scheduled into, when applicable.\n* `issueToBePublished` is a summary of the issue this submission will appear in once it is published.",
```

ewhanson: The diff is a bit difficult to parse, but I suggest combining the sentence that starts with `pages` is usually [...] and the following sentence.

ewhanson: > `stages`, `reviewRounds` and `reviewAssignments` should only be returned for participants assigned to the submission.\n* `reviewAssignments` should only expose user-identifying information for editors and admins.

The `should` implies to me there's some action needed on the user's part for this. Is it always the case or not always? My guess is this should just say something like they _are_ only returned for etc. rather than _should be_, but I'll defer to you for that.

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8224,1535 @@
...
+					{
+						"name": "isInactive",
+						"description": "Filter results by active or inactive sections.",
+						"in": "query",
+						"required": false,
+						"type": "boolean"
+					},
+					{
+						"name": "editorIds",
+						"description": "Filter results by one or more user ids assigned to a section as an editor.",
```
ewhanson:
```suggestion
						"description": "Filter results by one or more user IDs assigned to a section as an editor.",
```

docs/dev/swagger-source.json
```diff
@@ -7773,6 +8224,1535 @@
...
+					{
+						"name": "submissionId",
+						"description": "Submission ID",
+						"in": "path",
+						"required": true,
+						"type": "integer"
+					},
+					{
+						"name": "approved",
+						"description": "When provided, filter suggestions by whether they have been approved by an editor.",
```
ewhanson: Maybe language that says something like it's an optional parameter that... rather than "when provided"?

ewhanson (changes requested):

docs/dev/swagger-source.json
kaitlinnewson: Since we don't publish API docs for OMP/OPS, perhaps we could just say "series in OMP" instead (which I've gone ahead with here)? There are a few mentions of OMP/OPS in the API docs already.

ewhanson: Seems like a good approach. 👍

ewhanson (approved):

PR #5567, opened 2026-06-01
pkp/pkp-lib#12593 discussion-related template migration

Vitaliy-1:


touhidurabir (approved):
The main operation is in the `pkp-lib`'s migration and this OJS pr is only to add new migration and clean up of task email template from default ones , thus approving . But the tests are failing .

touhidurabir (approved):

PR #5568, opened 2026-06-02
pkp/pkp-lib#12841 Infrastructural support of laravel logging

touhidurabir:
for pkp/pkp-lib#12841 .

### DIFF
--- a/config.TEMPLATE.inc.php
+++ b/config.TEMPLATE.inc.php
@@ -583,6 +583,49 @@ deprecation_warnings = Off
 log_web_service_info = Off
 
 
+;;;;;;;;;;;;;;;;
+; Log Settings ;
+;;;;;;;;;;;;;;;;
+
+[logs]
+
+; Default logging channel. Available channels:
+;   - stack: Combines multiple channels (configurable via log_stacks; defaults to daily)
+;   - single: Single log file ({files_dir}/logs/app.log)
+;   - daily: Daily rotating log files
+;   - stderr: Output to stderr
+;   - syslog: System log
+;   - errorlog: PHP error log
+;   - null: Discard all logs
+log_channel = daily
+
+; Minimum log level to record. Available levels (in order of severity):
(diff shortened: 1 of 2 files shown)


config.TEMPLATE.inc.php
```diff
@@ -574,6 +574,53 @@
...
+;   - warning: Exceptional occurrences that are not errors
+;   - error: Runtime errors that do not require immediate action
+;   - critical: Critical conditions
+;   - alert: Action must be taken immediately
+;   - emergency: System is unusable
+log_level = error
+
+; Filename for the application log under {files_dir}/logs/.
+; Defaults to {application name}.log (e.g. ojs2.log, omp2.log, ops2.log).
+; log_file_name = ojs2.log
```
asmecher: Is there any reason we're using the `2` suffix? I'm even thinking we shouldn't name the log specific to each application; we could just call it `error_log` or something similar.

In fact, is there a benefit to making the name configurable at all?

touhidurabir: > Is there any reason we're using the 2 suffix? 

that's just as an example using same application name . 

> we could just call it error_log or something similar.

we can name it whatever we want to but better to named it as `error_log` which will confuse with PHP's default error log . and log does not only contains error but any debug, info, alert (basically any PSR log level) so error_log name does not justify it also . 

> In fact, is there a benefit to making the name configurable at all?

Truthfully not much , it's just a convenience. currently default named to application's name but we can have simple `logs.log`  as default also . and not make it configurable .

asmecher: I think we should just fix it as `system.log` or something. Filenames in the files directory shouldn't be configurable; they should be named consistently.

touhidurabir: I say go for `app.log` as it is app's own log .

touhidurabir: removed configurable log file name ability and set the log file name to app.log .

PR #5569, opened 2026-06-02
Add Japanese localization fixes for OJS

yuk1-kondo:
## Summary

This PR updates Japanese localization coverage for OJS 3.5 and adds missing English source strings for several OJS settings messages that are already referenced by other locales.

Changes include:

- Add missing English source strings in `locale/en/manager.po` for scheduled-task dependent review/subscription reminders.
- Improve Japanese translations in core OJS locale files where strings were empty or fell back to English.
- Add Japanese locale files for OJS invitation, SUSHI, import/export, reports, and DataCite components so Japanese users do not see untranslated labels or raw localization keys.

## Motivation

While operating an OJS 3.5 Japanese installation, several management screens showed untranslated English text or raw localization keys such as `##...##`. The affected strings were visible in journal setup, statistics/reports, import/export, SUSHI/COUNTER, and related management UI.

## Validation

- Parsed all changed `.po` files with `polib` successfully.
- Checked the translations against a live Japanese OJS 3.5 installation before preparing this upstream contribution.


asmecher:
Hi @yuk1-kondo,

Thanks for contributing! I've added some comments [on this PR](https://github.com/pkp/pkp-lib/pull/12849#issuecomment-4603971308) that apply here too.

yuk1-kondo:
Cross-posting the response from pkp/pkp-lib#12849 since this PR has the same shape:

I audited every English source key I had proposed adding here, and all three were already absent from the 3.5 code paths (`*.tpl`, `*.js`, `*.php`). The raw `##key##` rendering I observed came from a live install with legacy plugins/templates still referencing the old keys — not upstream `stable-3_5_0`.

Force-pushed an updated branch that:

- Reverts the English source additions in `locale/en/manager.po`.
- Removes Japanese entries whose `msgid` is no longer present in the 3.5 English source.
- Empties Japanese `msgstr` values that were still equal to the English source so there's no English mixed into the JP files.

This now stands as a translation-only contribution. Happy to move it to Weblate per the PKP Translation Guide if that's the preferred path.

asmecher (approved):

asmecher:
Thanks, @yuk1-kondo! I've merged this. See [my comment on the other PR](https://github.com/pkp/pkp-lib/pull/12849#issuecomment-4675672379).

PR #5576, opened 2026-06-05
pkp/pkp-lib#12800 Tighten review round association with publications

ewhanson:


dbscripts/xml/upgrade.xml
```diff
@@ -226,6 +226,7 @@
         <migration class="PKP\migration\upgrade\v3_6_0\I12251_SupportMediaFiles" />
         <migration class="PKP\migration\upgrade\v3_6_0\I12772_UserCommentForeignKeys" />
 		<migration class="PKP\migration\upgrade\v3_6_0\I12584_AddPublicationUpdateType" />
+        <migration class="PKP\migration\upgrade\v3_6_0\I12800_ReviewPublicationAssociationNullable" />
```
bozana, line 228: It seems PKP\migration\upgrade\v3_6_0\I12049_AddPublishReviewCurateSupport adds publication_id, back-fills it via addPublicationIds(), adds the FK + index — and then this script throws all of that away. Maybe we can then remove that I12049_AddPublishReviewCurateSupport fully?

bozana (commented):
Hi @ewhanson, just one comment regarding the migration scripts.
Thanks!

PR #5587, opened 2026-06-10
pkp/pkp-lib#12883 fix datacite plugin testUsername validation

kaitlinnewson:


plugins/generic/datacite/classes/DataciteSettings.php
```diff
@@ -110,20 +110,15 @@ public function getFields(Context $context): array
...
-        });
-
-        // If username exists, there will be the possibility to register from within OJS,
-        // so the test username must exist too
-        $validator->after(function (Validator $validator) use ($props) {
-            if (!empty($props['username']) && empty($props['testUsername'])) {
-                $validator->errors()->add('testUsername', __('plugins.importexport.datacite.settings.form.testUsernameRequired'));
+                if (empty($props['testUsername'])) {
+                    $validator->errors()->add('testUsername', __('plugins.importexport.datacite.settings.form.testUsernameRequired'));
+                }
```
bozana: I think the test username is not required -- the user can just want to use test prefix and export the XMLs and do something with them manually. 
Thus, I think the logic here was to require it only if the username also exist + test mode. Thus I think the code that was there is OK, it just needs to be moved into this if test mode statement.

plugins/generic/datacite/locale/en/locale.po
```diff
@@ -73,10 +73,10 @@ msgid "plugins.importexport.datacite.settings.form.testDOIPrefix"
 msgstr "Test DOI Prefix"
 
 msgid "plugins.importexport.datacite.settings.form.testDOIPrefixRequired"
-msgstr "A test DOI prefix is required when the \"use test prefix for DOI registration\" is selected."
+msgstr "A test DOI prefix is required when \"use test prefix for DOI registration\" is selected."
 
 msgid "plugins.importexport.datacite.settings.form.testUsernameRequired"
-msgstr "A test username is required when a production username is supplied."
+msgstr "A test username is required when \"use test prefix for DOI registration\" is selected."
```
bozana: "and a production username is supplied"

bozana (commented):
Hi @kaitlinnewson,
I think the problem should be solved slightly differently -- see my comments.
Thanks a lot!

PR #5604, opened 2026-06-16
pkp/pkp-lib#12005 Issue always required for legacy DOI suffix pattern

bozana:
s. https://github.com/pkp/pkp-lib/issues/12005

and 

https://github.com/pkp/pkp-lib/issues/12232

### DIFF
--- a/classes/doi/Repository.php
+++ b/classes/doi/Repository.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * @file classes/doi/Repository.php
  *
@@ -62,23 +63,12 @@ class Repository extends \PKP\doi\Repository
         }
 
         // If not using default suffix, additional checks are required
-        $issueId = $publication->getData('issueId');
-        if ($issueId === null) {
-            throw new DoiException(
-                DoiException::PUBLICATION_MISSING_ISSUE,
-                $submission->getCurrentPublication()->getLocalizedFullTitle(),
-                $publication->getLocalizedFullTitle()
-            );
-        }
+        $suffixPattern = $this->getPubIdSuffixPattern($publication, $context);
+        $patternNeedsIssue = $suffixPattern ? PubIdPlugin::suffixHasIssuePattern($suffixPattern) : false;
+        // generateSuffixPattern only needs the issue when $patternNeedsIssue is true
+        $issue = $patternNeedsIssue && $publication->getData('issueId') ? Repo::issue()->get($publication->getData('issueId')) : null;
 
-        $issue = Repo::issue()->get($publication->getData('issueId'));
-        if ($issue === null) {
-            throw new DoiException(
-                DoiException::PUBLICATION_MISSING_ISSUE,
-                $submission->getCurrentPublication()->getLocalizedFullTitle(),
-                $publication->getLocalizedFullTitle()
-            );
-        } elseif ($issue && $context->getId() != $issue->getJournalId()) {
+        if ($patternNeedsIssue && ($issue === null || $context->getId() != $issue->getJournalId())) {
             throw new DoiException(
                 DoiException::PUBLICATION_MISSING_ISSUE,
                 $submission->getCurrentPublication()->getLocalizedFullTitle(),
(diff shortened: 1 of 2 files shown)


classes/plugins/PubIdPlugin.php
```diff
@@ -340,6 +340,25 @@ public static function generateCustomPattern($context, $pubIdSuffix, $pubObject,
...
 
+    /**
+     * Checks if the given pubId suffix pattern includes any elements that would require
+     * the publication object to be assigned to an issue (volume number, issue number, or year).
+     *
+     * @param string $pubIdSuffix Full legacy suffix pattern to check against
+     */
+    public static function suffixHasIssuePattern(string $pubIdSuffix): bool
+    {
+        $hasVolumeNumber = (bool) PKPString::regexp_match('/%v/', $pubIdSuffix);
```
kaitlinnewson: This won't work for 3.5/main as `PKPString::regexp_match()` was removed - could replace with ` preg_match` or `str_contains`.

classes/plugins/PubIdPlugin.php
```diff
@@ -340,6 +340,25 @@ public static function generateCustomPattern($context, $pubIdSuffix, $pubObject,
...
+        $hasIssueNumber = (bool) PKPString::regexp_match('/%i/', $pubIdSuffix);
+        $hasYear = (bool) PKPString::regexp_match('/%Y/', $pubIdSuffix);
+
+        if ($hasVolumeNumber || $hasIssueNumber || $hasYear) {
```
kaitlinnewson: Could simplify to something like:

`return ($hasVolumeNumber || $hasIssueNumber || $hasYear);`

or like:

```
    return str_contains($pubIdSuffix, '%v')
        || str_contains($pubIdSuffix, '%i')
        || str_contains($pubIdSuffix, '%Y');
```

classes/doi/Repository.php
```diff
@@ -104,20 +100,23 @@ public function mintGalleyDoi(Galley $galley, Publication $publication, Submissi
         }
 
         // If not using default suffix, additional checks are required
+        $patternNeedsIssue = PubIdPlugin::suffixHasIssuePattern($this->getPubIdSuffixPattern($galley, $context));
```
kaitlinnewson: It looks like `getPubIdSuffixPattern()` can return null, so `suffixHasIssuePattern()` should consider that too

classes/doi/Repository.php
```diff
@@ -62,28 +63,23 @@ public function mintPublicationDoi(Publication $publication, Submission $submiss
...
-        } elseif ($issue && $context->getId() != $issue->getJournalId()) {
-            throw new DoiException(
-                DoiException::PUBLICATION_MISSING_ISSUE,
-                $submission->getCurrentPublication()->getLocalizedFullTitle(),
-                $publication->getLocalizedFullTitle()
-            );
+        $patternNeedsIssue = PubIdPlugin::suffixHasIssuePattern($this->getPubIdSuffixPattern($publication, $context));
+        $issue = $publication->getData('issueId') ? Repo::issue()->get($publication->getData('issueId')) : null;
+
+        if ($patternNeedsIssue) {
```
kaitlinnewson: Could optionally collapse these since they are returning the same (and for galley below), e.g.

`if ($patternNeedsIssue && ($issue === null || $context->getId() != $issue->getJournalId())) {`

kaitlinnewson (changes requested):

classes/doi/Repository.php
bozana: I am now checking earlier if getPubIdSuffixPattern() is null, so that suffixHasIssuePattern() can expect the real pattern

classes/plugins/PubIdPlugin.php
bozana: oh, such an elegant solution! :-) 👍

classes/plugins/PubIdPlugin.php
bozana: using your elegant suggestion below

kaitlinnewson (approved):

classes/doi/Repository.php
```diff
@@ -62,23 +63,12 @@ public function mintPublicationDoi(Publication $publication, Submission $submiss
...
-            throw new DoiException(
-                DoiException::PUBLICATION_MISSING_ISSUE,
-                $submission->getCurrentPublication()->getLocalizedFullTitle(),
-                $publication->getLocalizedFullTitle()
-            );
-        }
+        $suffixPattern = $this->getPubIdSuffixPattern($publication, $context);
+        $patternNeedsIssue = $suffixPattern ? PubIdPlugin::suffixHasIssuePattern($suffixPattern) : false;
+        // generateSuffixPattern only needs the issue when $patternNeedsIssue is true
+        $issue = $patternNeedsIssue && $publication->getData('issueId') ? Repo::issue()->get($publication->getData('issueId')) : null;
```
kaitlinnewson, line 69: This should be backported to 3.4 as well (there are still some errors in 3.4).

bozana: I think the major fix of the issue leading to an error will be done here, within this issue: https://github.com/pkp/pkp-lib/issues/12637. Maybe I could port there also this smaller code improvements.

bozana: Or the other way around. Maybe I can port all changes from here back -- so it is clearly visible within this issue...

PR #5635, opened 2026-06-28
pkp/pkp-lib#12152 Support Contributor Roles and Types

ajnyga:
Datacite

https://github.com/pkp/pkp-lib/issues/12152

### DIFF
--- a/plugins/generic/datacite/filter/DataciteXmlFilter.php
+++ b/plugins/generic/datacite/filter/DataciteXmlFilter.php
@@ -27,6 +27,8 @@ use APP\submission\Submission;
 use DOMDocument;
 use DOMNode;
 use Exception;
+use PKP\author\contributorRole\ContributorRoleIdentifier;
+use PKP\author\contributorRole\ContributorType;
 use PKP\context\Context;
 use PKP\core\PKPString;
 use PKP\db\DAORegistry;
@@ -215,6 +217,11 @@ class DataciteXmlFilter extends \PKP\plugins\importexport\native\filter\NativeEx
             }
             $rootNode->appendChild($subjectsNode);
         }
+        // Contributors
+        $contributorsNode = $this->createContributorsNode($doc, $publication);
+        if ($contributorsNode) {
+            $rootNode->appendChild($contributorsNode);
+        }
         // Dates
         $rootNode->appendChild($this->createDatesNode($doc, $issue, $article, $publication, $galleyFile, $publicationDate));
         // Language
@@ -305,8 +312,6 @@ class DataciteXmlFilter extends \PKP\plugins\importexport\native\filter\NativeEx
                 if (!empty($creator)) {
                     $creators[] = [
                         'name' => $creator,
-                        'orcid' => null,
-                        'affiliations' => null
                     ];
                     break;
                 }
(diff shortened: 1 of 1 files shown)


plugins/generic/datacite/filter/DataciteXmlFilter.php
```diff
@@ -315,11 +322,36 @@ public function createCreatorsNode(DOMDocument $doc, Issue $issue, ?Publication
                 $authors = $publication->getData('authors');
                 assert(!empty($authors));
                 foreach ($authors as $author) { /** @var Author $author */
-                    $creators[] = [
-                        'name' => $author->getFullName(false, true, $publication->getData('locale')),
-                        'orcid' => $author->getData('orcidIsVerified') ? $author->getData('orcid') : null,
-                        'affiliations' => $author->getAffiliations()
-                    ];
+
```
bozana, line 324: Creators are required, but with the code below they could be empty -- if no author role or all anonymous. So maybe in such a case to fall back to what old code did -- to consider all authors, i.e. something like
```
            $creatorsFallback[] = [
                  'name' => $author->getFullName(false, true, $publication->getData('locale')),
                  'orcid' => $author->getData('orcidIsVerified') ? $author->getData('orcid') : null,
                  'affiliations' => $author->getAffiliations()
              ];

```
and then later in this switch-case:
```
// DataCite requires at least one creator. If nobody was tagged AUTHOR,
      // fall back to listing every author, as before this change.
      if (empty($creators)) {
          $creators = $creatorsFallback;
      }
      break;

```

plugins/generic/datacite/filter/DataciteXmlFilter.php
```diff
@@ -365,6 +411,112 @@ public function createCreatorsNode(DOMDocument $doc, Issue $issue, ?Publication
...
+            if ($contributor['ror']) {
+                $node = $doc->createElementNS($deployment->getNamespace(), 'nameIdentifier');
+                $node->appendChild($doc->createTextNode($contributor['ror']));
+                $node->setAttribute('schemeURI', 'https://ror.org/');
+                $node->setAttribute('nameIdentifierScheme', 'ROR');
+                $contributorNode->appendChild($node);
+            }
+            if ($contributor['affiliations']) {
+                foreach ($contributor['affiliations'] as $affiliation) {
+                    $node = $doc->createElementNS($deployment->getNamespace(), 'affiliation');
```
bozana, line 515: maybe to assure the affiliation name is not empty, like elsewhere:
```
$institutionName = $affiliation->getLocalizedName($publication->getData('locale'));
          if (trim($institutionName ?? '') === '') {
              continue;
          }
```
and use `$institutionName` below.

bozana (commented):
Hi @ajnyga, a few comments.

plugins/generic/datacite/filter/DataciteXmlFilter.php
ajnyga: Do you mean present a contributor that is not an author as an author in the metadata?

bozana: yes :-( present them all, as it was until now, in case there is no creator. I do not have a better solution, because creators cannot be empty

ajnyga: I am not sure I like this. For me it would make more sense just not allow DOI registering if there are no creators - if that is a datacite requirement. Using just any contributor causes bad metadata.

ajnyga: Did you check this
https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/overview/#table-1-datacite-mandatory-properties

Guidance for handling missing mandatory property values

ajnyga: https://datacite-metadata-schema.readthedocs.io/en/4.6/appendices/appendix-3/

Examples of usage

```
<creators>
  <creator>
    <creatorName>:unav</creatorName>
  </creator>
</creators>
```

bozana: oh, great! then we can use `:unkn` for Anonymous ?

bozana: hmmm... and if no author role, what would be best to use?

ajnyga: if no author is given, then that is also "known to be unknown"?

ajnyga: or ":unav"?

bozana: I have just asked, lets see what they say :-)

ajnyga: I added the changes already since the final version is probably just about editing the correct standard value

bozana: They said to use ":unav" for this case.

ajnyga: great, that is what I added already earlier :)

PR #5652, opened 2026-07-06
Update VPAT yaml file to the latest audit

israelcefrin:
VPAT now covers OJS/OMP/OPS Editorial View (Dashboard) and Default Theme is an addendum to the Accessibility Conformance Report.

### DIFF
--- a/docs/vpat.yaml
+++ b/docs/vpat.yaml
@@ -1,109 +1,127 @@
-title: OJS Default Theme Accessibility Conformance Report
+title: OMP/OJS/OPS Editorial View Accessibility Conformance Report
 product:
-  name: OJS Default Theme
-  version: OJS 3.3.0-17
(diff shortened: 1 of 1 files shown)


asmecher (approved):

asmecher:
Thanks, @israelcefrin! We also have a copy of this file in `main`, and at least some aspects of it should be updated -- e.g. the contact information. I'm not sure what the best way to manage this is, given that there's risk of these documents forking. What do you think?

israelcefrin:
Hi @asmecher , can we port it to the main branch? This is the latest version and covers all OJS dashboard. Do you recommend me to file a new PR?

asmecher:
@israelcefrin, that works for me. I've cherry-picked it forward to `main` so the two are synced.

PR #5655, opened 2026-07-07
pkp/pkp-lib#12343 Consider data citations for metadata distribution

ajnyga:
https://github.com/pkp/pkp-lib/issues/12343

### DIFF
--- a/plugins/generic/datacite/filter/DataciteXmlFilter.php
+++ b/plugins/generic/datacite/filter/DataciteXmlFilter.php
@@ -31,6 +31,7 @@ use PKP\author\contributorRole\ContributorRoleIdentifier;
 use PKP\author\contributorRole\ContributorType;
 use PKP\context\Context;
 use PKP\core\PKPString;
+use PKP\dataCitation\DataCitation;
 use PKP\db\DAORegistry;
 use PKP\facades\Locale;
 use PKP\galley\Galley;
@@ -65,6 +66,8 @@ define('DATACITE_RELTYPE_ISPARTOF', 'IsPartOf');
 define('DATACITE_RELTYPE_ISPREVIOUSVERSIONOF', 'IsPreviousVersionOf');
 define('DATACITE_RELTYPE_ISNEWVERSIONOF', 'IsNewVersionOf');
 define('DATACITE_RELTYPE_ISPUBLISHEDIN', 'IsPublishedIn');
+define('DATACITE_RELTYPE_ISSUPPLEMENTEDBY', 'IsSupplementedBy');
+define('DATACITE_RELTYPE_REFERENCES', 'References');
 
 // Description types
 define('DATACITE_DESCTYPE_ABSTRACT', 'Abstract');
@@ -784,6 +787,16 @@ class DataciteXmlFilter extends \PKP\plugins\importexport\native\filter\NativeEx
                     }
                     unset($relatedGalley, $doi);
                 }
+                // Data citations.
+                foreach ($publication->getData('dataCitations') ?? [] as $dataCitation) {
+                    $relatedIdentifier = $this->getDataCitationRelatedIdentifier($dataCitation);
+                    if ($relatedIdentifier) {
+                        $relatedIdentifiersNode->appendChild($node = $doc->createElementNS($deployment->getNamespace(), 'relatedIdentifier', htmlspecialchars($relatedIdentifier['identifier'], ENT_COMPAT, 'UTF-8')));
+                        $node->setAttribute('relatedIdentifierType', $relatedIdentifier['relatedIdentifierType']);
+                        $node->setAttribute('relationType', $relatedIdentifier['relationType']);
+                    }
+                    unset($dataCitation, $relatedIdentifier);
+                }
                 break;
             case isset($issue):
                 // Parts: articles in this issue.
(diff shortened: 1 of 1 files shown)


plugins/generic/datacite/DataciteExportPlugin.php
```diff
@@ -383,7 +374,7 @@ public function _checkForTar()
      */
     public function _tarFiles($targetPath, $targetFile, $sourceFiles)
     {
-        assert((bool) $this->_checkForTar());
+        assert($this->tarFunctional());
```
bozana: I have just merged a cleanup that removed the assert() calls in this plugin — it was redundant with the check already done by the caller in exportAsDownload() before calling _tarFiles(). When you rebase, you'll likely get a conflict on that line; the fix is to just drop it rather than update it to reference tarFunctional().

plugins/generic/datacite/DataciteExportPlugin.php
```diff
@@ -356,22 +355,14 @@ public function updateDepositStatus(DataObject $object, string $status, ?string
...
         $tarBinary = Config::getVar('cli', 'tar');
-        if (empty($tarBinary) || !is_executable($tarBinary)) {
-            $result = [
-                ['manager.plugins.tarCommandNotFound']
-            ];
-        } else {
-            $result = true;
-        }
-        return $result;
+        return !empty($tarBinary) && file_exists($tarBinary);
```
bozana: `return !empty($tarBinary) && is_executable($tarBinary);`

If tar exists on disk but isn't executable (permissions, wrong file, etc.), this now reports "available" when it isn't...

plugins/generic/datacite/DatacitePlugin.php
```diff
@@ -279,6 +279,10 @@ public function isPluginConfigured(Context $context): bool
             }
         }
 
+        if (!$this->isTarAvailable()) {
+            return false;
+        }
```
bozana: tar is only needed when bundling more than one exported file into a single archive — _tarFiles() is only ever called from exportAsDownload() when `count($exportedFiles) > 1`. Deposits (exportAndDeposit()) never call it at all; each object gets deposited individually. So this isn't really a plugin configuration problem — the plugin can register/deposit/export single items just fine without tar.

Tying it into isPluginConfigured() has a bigger effect than intended, though: that flag gates all per-item register/deposit actions in the UI, so a journal that only ever handles one item at a time would be blocked from using the plugin at all, even though it would never hit the tar code path.

Two suggestions:
1. Check tarFunctional() at the point where it actually matters — when more than one item is being exported/downloaded as a bundle — and surface a specific error there, rather than gating overall plugin configuration.
I.e. to have something like:
In DataciteExportPlugin.php's exportAsDownload():
```
if (count($objects) > 1 && !$this->tarFunctional()) {                                                                                          
    $outputErrors[] = ['manager.plugins.tarCommandNotFound'];                                                                                  
    return null;                                                                                                                               
}    
```  
2. Keep the settings-page notice, but reword it to make clear it only affects downloading/exporting multiple items as one archive — not DOI registration/deposit in general — so it doesn't read as "this plugin is broken."
E.g. something like:
In DatacitePlugin.php:
```
public function isTarAvailable(): bool                                                                                                             
{                                                                                                                                                  
    return $this->_getExportPlugin()->tarFunctional();                                                                                             
}  
```
In DataciteSettings.php:
```
        $text = '';
        if (!$this->agencyPlugin->isTarAvailable()) {
            $text .= '<div class="pkpNotification pkpNotification--warning">';
            $text .= '<p>' . __('plugins.importexport.datacite.error.tarNotConfigured') . '</p>';
            $text .= '</div>';
        }

```
and the locale key, something like:
`The tar binary could not be found on the server. This only affects exporting more than one item at a time, which requires bundling the individual files into a single archive; registering or depositing DOIs, and exporting a single item at a time, are unaffected. To enable bundling, install tar and configure its path in the [cli] tar setting of config.inc.php.`

bozana (commented):
Hi @ajnyga, I think we should handle missing tar differently -- see my comments. Let me know what do you think.
The data citation export is fine.
I have merged the crossref and jatsTemplate PR, so you would need to reset hard those to the newest pkp main, and re-add the submodules again.
Thanks a lot!

plugins/generic/datacite/DatacitePlugin.php
ajnyga: > the plugin can register/deposit/export single items just fine without tar.
 
 Are you sure about this? I get the tar related error when I try to export a single DOI.

ajnyga: So testing with the current main branch, if I export one DOI with the datacite plugin the file that is being dowloaded is "null.json".

This problem is fixed when the tar path is corrected.

ajnyga: So this https://github.com/pkp/ojs/blob/main/plugins/generic/datacite/DataciteExportPlugin.php#L206 is being called before the code checks how many files are being downloaded.

If the configured tar path is invalid, _checkForTar() fails and execution falls through to return null; on line 251. That null becomes the temporaryFileId with no errors visible and the browser saves that JSON response as null.json.

bozana: Yes, this is what I suggested to change here to be fixed: `if (count($objects) > 1 && !$this->tarFunctional()) { `

bozana: but now i see those error messages are not displayed in the UI. So, if you want, I will try to fix that all with TAR and you can here only consider data citations....

ajnyga: So I think I have everything you suggested but like you mentioned the plugin is not showing the error messages with $outputErrors

ajnyga: leaving just the data citations part meaning that incorrect tar config would still break this.

PR #5656, opened 2026-07-07
pkp/pkp-lib#12950 add version and relations to OAI metadata formats

kaitlinnewson:


plugins/oaiMetadataFormats/marc/OAIMetadataFormat_MARC.php
```diff
@@ -71,4 +79,41 @@ public function toXml($record, $format = null)
...
+                $relation->marcField = match ($relation->relationType) {
+                    VersionRelationType::IS_NEW_VERSION_OF => '780',
+                    VersionRelationType::IS_PREVIOUS_VERSION_OF => '785',
+                    default => '775',
+                };
+                // 780/785 second indicator 2 = Supersedes / Superseded by; 775 leaves it undefined.
+                $relation->marcRelationIndicator = $relation->marcField === '775' ? ' ' : '2';
+                // Relationship information ($i): name the update type
+                // when known, otherwise fall back to the version label.
+                $relation->relationLabel = $relation->updateType?->label() ?: $relation->versionString;
```
bozana: because of that problem with updateType mentioned in pkp-lib PR, maybe to drop it fully here and only use versionString?
here and in MARC21

plugins/oaiMetadataFormats/marc/OAIMetadataFormat_MARC.php
```diff
@@ -71,4 +79,41 @@ public function toXml($record, $format = null)
...
+        return Repo::publication()->getVersionRelations($publication, $article, $journal)
+            ->map(function (object $relation) use ($request, $dispatcher, $journal, $article): object {
+                $relation->identifier = $relation->doi ?:
+                    $dispatcher->url(
+                        $request,
+                        PKPApplication::ROUTE_PAGE,
+                        $journal->getPath(),
+                        'article',
+                        'view',
+                        [$article->getId(), 'version', $relation->publicationId],
```
bozana: should we use `$article->getBestId()` ?
here and in MARC21

bozana (commented):
Hi @kaitlinnewson, a few comments

plugins/metadata/dc11/filter/Dc11SchemaArticleAdapter.php
```diff
@@ -217,6 +216,16 @@ public function &extractMetadataFromDataObject(&$article)
             }
         }
 
+        // Relation: the immediately preceding published version (backward-only, chain-only)
+        $versionRelation = Repo::publication()->getVersionRelation($publication, $article, $journal);
+        if ($versionRelation) {
+            if ($versionRelation->doiUrl) {
+                $dc11Description->addStatement('dc:relation', $versionRelation->doiUrl);
+            } elseif ($includeUrls) {
+                $dc11Description->addStatement('dc:relation', $request->getDispatcher()->url($request, PKPApplication::ROUTE_PAGE, null, 'article', 'view', [$article->getBestId(), 'version', $versionRelation->publicationId], urlLocaleForPage: ''));
```
bozana: hmmm... `null` instead of journal path was used above too, but this seems to be not correct -- the OAI can be used on the site level and in that case the URL would look something like `http://ojs-master.bb/index.php/index/article/view/mwandenga`, which is not correct. Could you please fix to use journal path everywhere for DC?

bozana (commented):
hi @kaitlinnewson, just one comment -- a bug from earlier that we could fix now. Thanks a lot!

PR #5661, opened 2026-07-09
pkp/pkp-lib#13011 Route citation API requests through the submissions controller for proper access control

bozana:
see https://github.com/pkp/pkp-lib/issues/13011


api/v1/submissions/index.php
```diff
@@ -39,6 +39,17 @@
...
 }
 
+// submissions/{submissionId}/publications/{publicationId}/citations[/{citationId}[/reprocessCitation]]
+// Excludes the publication-level bulk citation actions handled by SubmissionController below
+// (.../citations/importAdditionalCitations, .../citations/deleteCitationsByPublicationId, .../citations/reprocessCitationsByPublicationId).
+if (count($urlParts) >= 8 && $urlParts[7] == 'citations' && !in_array($urlParts[8] ?? null, [
+    'importAdditionalCitations',
+    'deleteCitationsByPublicationId',
+    'reprocessCitationsByPublicationId',
+], true)) {
```
touhidurabir: @bozana  This is not a blocker or anything wrong but what are the chances of moving all the citation related api to the citation controller . IMO, that make it cleaner and more centralise . However I understand that is out the scope of issue so up to you if you want to refactor it .

touhidurabir (approved):
@bozana this look good . Just one comment and up to you of you want to refactor it .

api/v1/submissions/index.php
bozana: You mean to also move these 3 functions to PKPCitationController?

touhidurabir: Right, if that is not a lot of work . not sure why those are there (probably there is some reason and I am guessing wrong) but seems like making change to citation based on submission and publication which we have here anyway as we have updated the API path that includes the submission and publication id .

bozana: yes, I will do that... Thanks a lot! -- I do not why it did not cross my mind... :sweat_smile: ;-)

bozana:
Hi @touhidurabir, I have refactored it. If you would like to take a look, please do. Else, once the tests pass I will merge...
Thanks a lot!

PR #5674, opened 2026-07-14
 pkp/pkp-lib#12922 add OAI record versions for DOI versions and cleanup OAI classes

kaitlinnewson:


classes/oai/ojs/OAIDAO.php
```diff
@@ -274,13 +291,19 @@ public function _getRecordsRecordSetQuery($setIds, $from, $until, $set, $submiss
...
+                'p.version_major AS version_major',
             ])
             ->join('publications AS p', 'a.current_publication_id', '=', 'p.publication_id')
             ->leftJoin('issues AS i', 'i.issue_id', '=', 'p.issue_id')
             ->join('sections AS s', 's.section_id', '=', 'p.section_id')
             ->join('journals AS j', 'j.journal_id', '=', 'a.context_id')
             ->where('j.enabled', '=', 1)
             ->where('p.status', '=', PKPPublication::STATUS_PUBLISHED)
+            ->when($versioningJournalIds, function ($query, $versioningJournalIds) {
+                return $query->whereNotIn('j.journal_id', $versioningJournalIds);
```
bozana, line 314: It seems articles disappear from OAI-PMH entirely if their current publication isn't yet a Version of Record -- see $versionQuery below.

classes/oai/ojs/OAIDAO.php
```diff
@@ -291,56 +314,145 @@ public function _getRecordsRecordSetQuery($setIds, $from, $until, $set, $submiss
...
+                    $join->on('tsos.tombstone_id', '=', 'dot.tombstone_id');
+                    $join->where('tsos.assoc_type', '=', Application::ASSOC_TYPE_SECTION);
+                    $join->where('tsos.assoc_id', '=', (int) $sectionId);
+                })->addSelect(['tsos.assoc_id']);
+            }, function ($query) {
+                return $query->addSelect([DB::raw('NULL AS assoc_id')]);
+            })
+            // Extra columns to match the live-record branches for the UNION.
+            ->addSelect([
+                DB::raw('NULL AS publication_id'),
```
bozana, line 446: tombstones have no publication_id/version_major, so deleting an article only tombstones the bare identifier via AticleTombstoneManager::insertArticleTombstone(), meaning per-version identifiers like article/5/version/1 never get a deletion record

classes/oai/ojs/OAIDAO.php
```diff
@@ -244,11 +240,16 @@ public function setOAIData($record, $row, $isRecord = true)
 
     /**
      * @copydoc PKPOAIDAO::_getRecordsRecordSet
```
bozana: @copydoc PKPOAIDAO::getRecordsRecordSetQuery()

bozana (commented):
Hi @kaitlinnewson, a few comments, but most important:
1) we should consider AO and PMUR as well
2) tombstones need a rewrite:
- On Publication\Repository::unpublish($publication): after unpublishing, check whether this was the last live representative of its identifier using the same "latest published minor of major" logic $versionQuery already implements — and if so, tombstone that specific identifier there and then. It should not be gated on the submission ever having reached STATUS_PUBLISHED.
- On Publication\Repository::publish($publication): if this publish makes a previously-tombstoned identifier live again (bare or a specific /version/N), remove that tombstone — again regardless of version stage, so PMUR/AO republishing correctly clears stale tombstones too.

Thanks a lot!

classes/oai/ojs/OAIDAO.php
```diff
@@ -264,23 +270,53 @@ public function _getRecordsRecordSetQuery($setIds, $from, $until, $set, $submiss
...
+            })
+            ->pluck('journal_id')
+            ->all();
+
+        // Version stages exposed as their own OAI record when DOI versioning is enabled.
+        $versionedStages = [
+            VersionStage::AUTHOR_ORIGINAL->value,
+            VersionStage::PUBLISHED_MANUSCRIPT_UNDER_REVIEW->value,
+            VersionStage::VERSION_OF_RECORD->value,
+        ];
```
bozana: Maybe to use `array_column(VersionStage::cases(), 'value')`. We have those listed here and in ArtileTombstoneManager, so we can eventually forget about maintaining them...

classes/article/ArticleTombstoneManager.php
```diff
@@ -34,26 +40,114 @@ public function __construct()
...
+     *
+     */
+    protected function versionedStages(): array
+    {
+        return [
+            VersionStage::AUTHOR_ORIGINAL->value,
+            VersionStage::PUBLISHED_MANUSCRIPT_UNDER_REVIEW->value,
+            VersionStage::VERSION_OF_RECORD->value,
         ];
+    }
```
bozana, line 67: Maybe to use `array_column(VersionStage::cases(), 'value')` instead? -- This was maybe needed one we only considered VoR, but now we consider every stage...

bozana (commented):
Hi @kaitlinnewson, 2 comments...

PR #5715, opened 2026-07-28
pkp/pkp-lib#13062 Add support for peer review DOI export & deposit in Datacite plugin

taslangraham:
For pkp/pkp-lib#13062

### DIFF
--- a/plugins/generic/datacite/DataciteExportPlugin.php
+++ b/plugins/generic/datacite/DataciteExportPlugin.php
@@ -25,12 +25,15 @@ use PKP\config\Config;
 use PKP\context\Context;
 use PKP\core\DataObject;
 use PKP\core\PKPApplication;
+use PKP\db\DAORegistry;
 use PKP\doi\Doi;
 use PKP\file\FileManager;
 use PKP\file\TemporaryFileManager;
 use PKP\galley\Galley;
 use PKP\plugins\Plugin;
 use PKP\submission\Representation;
+use PKP\submission\reviewAssignment\ReviewAssignment;
+use PKP\submission\reviewRound\ReviewRoundDAO;
 
 // DataCite API
 define('DATACITE_API_RESPONSE_OK', 201);
@@ -95,6 +98,11 @@ class DataciteExportPlugin extends DOIPubIdExportPlugin
     public function getSubmissionFilter()
     {
         return 'article=>datacite-xml';
+    }
+
+    public function getPeerReviewFilter()
+    {
+        return 'peerReview=>datacite-xml';
     }
 
     /**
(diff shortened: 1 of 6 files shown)


plugins/generic/datacite/filter/DataciteXmlFilter.php
```diff
@@ -382,6 +496,59 @@ public function createTitlesNode(
...
+                    $reviewNumber = array_search($reviewAssignment->getId(), $reviewIds) + 1;
+
+                    // Translate locale using Locale::getBundle so that a `null` value is returned for missing locale
+                    // instead of the ##locale.key format
+                    $title = Locale::getBundle($locale)
+                        ->translateSingular('plugins.generic.dataCite.reviewTitle', [
+                            'publicationTitle' => $publication->getLocalizedTitle($locale),
+                            'revisionNumber' => $revisionNumber,
+                            'reviewNumber' => $reviewNumber,
+                        ]);
```
taslangraham, line 543: I didn't come across any suggested peer review title format from DataCite to help differentiate one review from the next, so I adapted the title format used in Crossref.

plugins/generic/datacite/locale/en/locale.po
```diff
@@ -93,3 +93,6 @@ msgstr "DataCite"
 
 msgid "plugins.generic.datacite.deposit.unsuccessful"
 msgstr "Registration was not successful! Please check your configuration and try again."
+
+msgid "plugins.generic.dataCite.reviewTitle"
+msgstr "Review: {$publicationTitle} (Round{$revisionNumber}/Review{$reviewNumber})"
```
kaitlinnewson, line 98: Since we are creating a review title for Datacite, Crossref, and JATS, I wonder if it makes sense to move this to pkp-lib to reduce duplication? And perhaps even a helper to generate the title when needed?

Then we can also standardize to one format, as I went with something slightly different in JATS (https://github.com/pkp/pkp-lib/issues/12198) (but am open to changes). I do think however that the publication title would benefit from quotation marks around it, as well as a space after "Round" and "Review".

What do you think @taslangraham and @bozana?

plugins/generic/datacite/DataciteExportPlugin.php
```diff
@@ -440,6 +448,28 @@ public function _getObjectUrl($request, $context, $object)
...
+                $publication = Repo::publication()->get($reviewRound->getData('publicationId'));
+
+                $url = $dispatcher->url(
+                    $request,
+                    PKPApplication::ROUTE_PAGE,
+                    $context->getPath(),
+                    'article',
+                    'view',
+                    [
+                        $publication->getData('urlPath') ?? $publication->getData('submissionId'),
```
bozana, line 465: if the publication is not the current publication the url could be wrong, right? maybe to use [$submission->getBestId(), 'version', $publication->getId()] instead, here and for Crossref?

PR #5719, opened 2026-07-30
pkp/pkp-lib#13115 DataCite: handle publications with no assigned issue and replace assert()

bozana:
see https://github.com/pkp/pkp-lib/issues/13115

### DIFF
--- a/plugins/generic/datacite/filter/DataciteXmlFilter.php
+++ b/plugins/generic/datacite/filter/DataciteXmlFilter.php
@@ -26,6 +26,7 @@ use APP\publication\Publication;
 use APP\submission\Submission;
 use DOMDocument;
 use DOMNode;
+use Exception;
 use PKP\context\Context;
 use PKP\core\PKPString;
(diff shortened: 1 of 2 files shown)


plugins/generic/datacite/filter/DataciteXmlFilter.php
kaitlinnewson:
```suggestion
                break;
```

kaitlinnewson:
```suggestion
            $publicationDate = $issue?->getDatePublished();
```

kaitlinnewson: Not necessarily in scope, but if time allows it would be good to clean up some of these `asserts` and replace with exceptions and/or better handling as appropriate.

kaitlinnewson: And consider how to handle null pubdate below?

kaitlinnewson (approved):

PR #5748, opened 2026-08-19
pkp/pkp-lib#13195 Consider version stages other than VoR in export pl…

bozana:
…ugins, mark deprecated export code

see https://github.com/pkp/pkp-lib/issues/13195

### DIFF
--- a/classes/galley/DAO.php
+++ b/classes/galley/DAO.php
@@ -19,6 +19,8 @@ class DAO extends PKPGalleyDAO
      * @param null|mixed $pubIdSettingName
      * @param null|mixed $pubIdSettingValue
      * @param null|mixed $rangeInfo
+     *
+     * @deprecated 3.4
      */
     public function getExportable(int $contextId, $pubIdType = null, $title = null, $author = null, $pubIdSettingName = null, $pubIdSettingValue = null, $rangeInfo = null, ?int $issueId = null)
     {
--- a/classes/galley/DAO.php
+++ b/classes/galley/DAO.php
@@ -19,6 +19,8 @@ class DAO extends PKPGalleyDAO
      * @param null|mixed $pubIdSettingName
      * @param null|mixed $pubIdSettingValue
      * @param null|mixed $rangeInfo
+     *
+     * @deprecated 3.4.0
      */
     public function getExportable(int $contextId, $pubIdType = null, $title = null, $author = null, $pubIdSettingName = null, $pubIdSettingValue = null, $rangeInfo = null, ?int $issueId = null)
     {
--- a/classes/plugins/PubObjectsExportPlugin.php
+++ b/classes/plugins/PubObjectsExportPlugin.php
@@ -23,6 +23,7 @@ use APP\issue\Issue;
 use APP\journal\Journal;
 use APP\journal\JournalDAO;
 use APP\notification\NotificationManager;
+use APP\publication\enums\VersionStage;
 use APP\publication\Publication;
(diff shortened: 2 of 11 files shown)


classes/galley/DAO.php
```diff
@@ -19,6 +19,8 @@ class DAO extends PKPGalleyDAO
      * @param null|mixed $pubIdSettingName
      * @param null|mixed $pubIdSettingValue
      * @param null|mixed $rangeInfo
+     *
+     * @deprecated 3.4.0
```
kaitlinnewson: Should this (and/or others in this PR) be 3.6.0, since it's the first release where it's deprecated? Not sure what our typical practice is for deprecation.

classes/publication/DAO.php
```diff
@@ -67,11 +67,15 @@ public function fromRow(object $primaryRow): Publication
     }
 
     /**
-     * Get all published VoR versions (eventually with a pubId assigned and) matching the specified settings.
-     * Only the latest minor version of a major version will be considered.
+     * Get all published versions (eventually with a pubId assigned and) matching the specified settings.
```
kaitlinnewson:
```suggestion
     * Get all published versions (with a pubId assigned) matching the specified settings.
```

classes/plugins/PubObjectsExportPlugin.php
```diff
@@ -884,40 +896,50 @@ public function executeCLICommand($scriptName, $command, $context, $outputFile,
...
      *
+     * Uses getExportable() rather than submission.status, which only reflects
+     * a published Version of Record, not earlier exportable stages.
+     *
      * @param array $submissionIds
      * @param Journal $context
      *
      * @return array
      */
     public function getPublishedSubmissions($submissionIds, $context)
```
kaitlinnewson, line 907: Should `DOIPubIdExportPlugin::getPublishedSubmissions()` be updated as well? I guess only medra is using it (maybe another thing to change later too 🙂).

controllers/grid/publications/ExportPublishedPublicationsListGridCellProvider.php
```diff
@@ -104,7 +104,11 @@ public function getTemplateVarsFromRowColumn($row, $column)
             case 'submissionId':
                 return ['label' => $publication->getData('submissionId')];
             case 'version':
-                return ['label' => $publication->getData('versionMajor') . '.' . $publication->getData('versionMinor')];
+                return ['label' => __('publication.versionStage.display', [
```
kaitlinnewson, line 107: Could this be simplified to `return ['label' => $publication->getData('versionString')];`?

kaitlinnewson (approved):

classes/galley/DAO.php
bozana: it seems to be OK, I checked with Alec. Maybe we will even remove them all now in 3.6 (because not LTS)

controllers/grid/publications/ExportPublishedPublicationsListGridCellProvider.php
bozana: This would create a longer strings i.e. "Version of Record", "Published Manuscript Under Review", "Author Original", which is too much for the grid.

classes/plugins/PubObjectsExportPlugin.php
bozana: This function is also not used any more. (It is also different that the one here.) I marked a few more functions there deprecated as well...

PR #5757, opened 2026-08-21
pkp/pkp-lib#13208 submodule update

bozana:
…ugins, mark deprecated export code

see https://github.com/pkp/pkp-lib/issues/13208

only submodule update


bozana:
i am merging this because the tests failures are unrelated...

PR #5768, opened 2026-08-26
pkp/pkp-lib#13225 Issue TOC mail content article and galley link fixed

touhidurabir:
for pkp/pkp-lib#13225 .


templates/frontend/objects/galley_link.tpl
```diff
@@ -62,7 +78,7 @@
 {/if}
 
 {* Don't be frightened. This is just a link *}
-<a class="{if $isSupplementary}obj_galley_link_supplementary{else}obj_galley_link{/if} {$type|escape}{if $restricted} restricted{/if}" href="{url page=$page op="view" path=$path}"{if $id} id="{$id}"{/if}{if $labelledBy} aria-labelledby="{$labelledBy}"{/if}>
+<a class="{if $isSupplementary}obj_galley_link_supplementary{else}obj_galley_link{/if} {$type|escape}{if $restricted} restricted{/if}" href="{url router=PKP\core\PKPApplication::ROUTE_PAGE journal=$galleyJournalPath page=$page op="view" path=$path}"{if $id} id="{$id}"{/if}{if $labelledBy} aria-labelledby="{$labelledBy}"{/if}>
```
kaitlinnewson: Should this check `{if $galleyJournalPath}` as well, before `journal=...`?

templates/frontend/objects/galley_link.tpl
```diff
@@ -18,13 +18,29 @@
  * @uses $currentJournal Journal The current journal context
  * @uses $journalOverride Journal An optional argument to override the current
  *       journal with a specific context
+ * @uses $journal Journal An optional argument naming the journal these galleys
```
kaitlinnewson: Could `$currentJournal` be used instead of a new argument?

templates/frontend/objects/galley_link.tpl
```diff
@@ -18,13 +18,29 @@
...
 	{assign var="currentJournal" value=$journalOverride}
 {/if}
 
+{* Resolve the journal this link belongs to, so that the URL stays correct when the
+ * current request has no journal context -- for example when this template is rendered
+ * into a notification email from a queued job. See IssueEmailVariable::getIssueToc().
+ * The link below pins ROUTE_PAGE for the same reason: that job may be flushed at the end
+ * of a component request, whose router cannot build a page URL. *}
+{assign var="galleyJournalPath" value=null}
+{if $journalOverride}
```
kaitlinnewson: Is this branch needed since, if this is set, then it's captured in `$currentJournal`?

templates/frontend/objects/galley_link.tpl
touhidurabir: good point and after a bit digging, it seems this is not needed at all. 

The only place where a per-object `$journal` genuinely differs from `$currentJournal` is site-wide search and that same include passes `hideGalleys=true`, which `article_summary.tpl` uses to skip the galley loop entirely. So `galley_link.tpl` is never reached from search. `$journal` had nothing to add here. so basically the use of `$journal` is over checking and kind of dead. 

This allows folds all the checks just to `$currentJournal` and the change moved to just link building with a `if...else` check, one single line . 

This also address the other 2 points as well .

kaitlinnewson (approved):

classes/mail/variables/IssueEmailVariable.php
```diff
@@ -81,6 +81,7 @@ protected function getIssueToc(): string
         IssueHandler::_setupIssueTemplate($request, $this->issue, $this->getContext(), false);
 
         $templateMgr->assign([
+            'journal' => $this->getContext(),
```
bozana, line 84: $currentJournal is provided in the line above, with IssueHandler::_setupIssueTemplate -- see my comment below.

templates/frontend/objects/article_summary.tpl
```diff
@@ -32,7 +32,7 @@
 <div class="obj_article_summary">
 	{if $publication->getLocalizedData('coverImage')}
 		<div class="cover">
-			<a {if $journal}href="{url journal=$journal->getPath() page="article" op="view" path=$articlePath}"{else}href="{url page="article" op="view" path=$articlePath}"{/if} class="file">
+			<a {if $journal}href="{url router=PKP\core\PKPApplication::ROUTE_PAGE journal=$journal->getPath() page="article" op="view" path=$articlePath}"{else}href="{url router=PKP\core\PKPApplication::ROUTE_PAGE page="article" op="view" path=$articlePath}"{/if} class="file">
```
bozana, line 35: Eventually we could use something like:
```
{assign var="linkJournal" value=$journal|default:$currentJournal}
{if $linkJournal}{assign var="journalPath" value=$linkJournal->getPath()}{/if}
```
and then have here only:
`<a href="{url router=PKP\core\PKPApplication::ROUTE_PAGE journal=$journalPath page="article" op="view" path=$articlePath}" class="file">`
This removes the need to extra assign `$journal` in `IssueEmailVariable`, because `$currentJournal` is already assigned via `IssueHandler::_setupIssueTemplate.`

bozana: eventually only 
`{assign var="linkJournal" value=$journal|default:$currentJournal}`
and no change in `IssueEmailVariable` here for 3.5
and  collapsing `{if $journal}…{else}…{/if}` into single `{url}` calls via `$journalPath` only in main?

bozana (approved):
Hi @touhidurabir, the changes look good, just a few comments, if they would even improve it all. You can then maybe decide what you think is the best way forward.

templates/frontend/objects/article_summary.tpl
touhidurabir: > Eventually we could use something like:
> 
> ```
> {assign var="linkJournal" value=$journal|default:$currentJournal}
> {if $linkJournal}{assign var="journalPath" value=$linkJournal->getPath()}{/if}
> ```
> 
> and then have here only: `<a href="{url router=PKP\core\PKPApplication::ROUTE_PAGE journal=$journalPath page="article" op="view" path=$articlePath}" class="file">` This removes the need to extra assign `$journal` in `IssueEmailVariable`, because `$currentJournal` is already assigned via `IssueHandler::_setupIssueTemplate.`

@bozana this would have been possible if we support the context aware jobs (as the mail is delivered via queue job) but we are not . So a job related to one context can run in another context's request life cycle via Job runner (not cause problem in CLI worker). 

so the one that can be wrong is $journal as, for example `AnnouncementFeedGatewayPlugin` and `GatewayHandler` assign journal into the request-wide TemplateManager, so when journal B's feed request flushes the queue, `$journal|default:$currentJournal` picks journal B for the article links. The explicit assign in IssueEmailVariable overwrites it. It also keeps theme copies of article_summary.tpl working.

Also this fix it for theme without modifying for them .

bozana: I see...

PR #5769, opened 2026-08-26
pkp/pkp-lib#12311 submodules updates

bozana:
see https://github.com/pkp/pkp-lib/issues/12311

### DIFF
--- a/jobs/statistics/ProcessUsageStatsLogFile.php
+++ b/jobs/statistics/ProcessUsageStatsLogFile.php
@@ -17,6 +17,7 @@
 namespace APP\jobs\statistics;
 
 use APP\core\Application;
+use APP\statistics\StatisticsHelper;
 use APP\statistics\TemporaryItemInvestigationsDAO;
 use APP\statistics\TemporaryItemRequestsDAO;
 use APP\statistics\TemporaryTotalsDAO;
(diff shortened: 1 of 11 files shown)


jobs/statistics/ProcessUsageStatsLogFile.php
```diff
@@ -75,7 +76,7 @@ protected function insertTemporaryUsageStatsData(object $entry, int $lineNumber)
             $temporaryInstitutionsDao->insert($entry->institutionIds, $lineNumber, $this->loadId);
             if (!empty($entry->submissionId)) {
                 $temporaryItemInvestigationsDao->insert($entry, $lineNumber, $this->loadId);
-                if ($entry->assocType == Application::ASSOC_TYPE_SUBMISSION_FILE) {
+                if (in_array($entry->assocType, [Application::ASSOC_TYPE_SUBMISSION_FILE, Application::ASSOC_TYPE_JATS])) {
```
ewhanson: Just curious if this should be hard  coded here or if it should be drawn from `TemporaryTotalsDAO::getItemRequestAssocTypes()` as look like the same set of items from an uninformed, outsider perspective.

ewhanson (approved):

jobs/statistics/ProcessUsageStatsLogFile.php
bozana: Hi @ewhanson, good catch! I now put it into StatisticsHelper::getItemRequestAssocTypes(), so that all use the same set.

PR #5810, opened 2026-09-09
pkp/pkp-lib#13221 Implement crossref's citedBy feature

taslangraham:
For pkp/pkp-lib#13221


bozana:
Hi @taslangraham, also here, the hook ArticleHandler::view could be used for this, I think -- no need to have something plugin specific in the core here...

bozana:
The default there does not have the cited-by 'button' which is ok for now -- ORE is the priority for now. We need to see with Devika and Jarda, but maybe it could use the hook Templates::Article::Footer::PageFooter (similarly to Crossmark button).

plugins/themes/default/styles/components/citedBy.less
```diff
@@ -0,0 +1,87 @@
...
+/* --- Visual --- */
+// Styles for the modal content
+.CrossrefCitedByBody__citationsList {
+	list-style: none;
+	padding: 1.43rem;
+}
+
+.CrossrefCitedByBody__actionsCopyBtn {
+	background-color: @bg-base;
+	color: #ffffff;
```
bozana, line 24: `@bg-base` is the journal's configurable base colour. When it's light, the theme switches the text colour to dark via `@text-bg-base` (DefaultThemePlugin.php:172-174), but the hardcoded `#ffffff` here gives white text on a light button. Could this use color: `@text-bg-base;`? Maybe to double check with Jarda :-)

plugins/themes/default/styles/components/citedBy.less
```diff
@@ -0,0 +1,87 @@
+@import (reference) "../components.less";
```
bozana, line 1: The default theme now styles the Crossref plugin's internal class names (CrossrefCitedByBody__…, CrossrefCitedBy__…). This couples the core theme to the plugin's markup: if a class is renamed in the plugin, the styling here silently stops applying. Is this the approach we want, or should the plugin's own CSS carry these styles, with the theme only adjusting colours or fonts where needed?

bozana (commented):
Hi @taslangraham, just a few comments, that maybe should be double checked with Jarda :-)

